PR #5, opened 2015-02-21
[FIX] Add gulp-util to dev dependencies

stevenzeiler:
Otherwise build does not work out of the box


jbenet:
is this still needed? cc @travisperson

travisperson:
I don't believe so, I haven't had any issues. @stevenzeiler what issues were you running into? I don't see where gulp-utils was being used.

travisperson:
I think this can be closed.

@jbenet.

jbenet:
Ok

Issue #6, opened 2015-02-24
Missing vendored files


travisperson:
`/static/bootstrap/3.3.2/css/bootstrap.min.css` [This link should resolve in this repo](https://github.com/protocol/ipfs-webui/tree/master/static/bootstrap/3.3.2/css/bootstrap.min.css)

https://github.com/protocol/ipfs-webui/blob/master/html/index.html#L17

`/static/bootstrap/3.3.2/js/bootstrap.min.js` [This link should resolve in this repo](https://github.com/protocol/ipfs-webui/tree/master/static/bootstrap/3.3.2/js/bootstrap.min.js)

https://github.com/protocol/ipfs-webui/blob/master/html/index.html#L28

These either need to be vendored or produced during the static step in gulp.

STRML:
This can probably be closed?

jbenet:
yep!

PR #12, opened 2015-02-27
Fixes issue #6

travisperson:
Added the missing bootstrap minified js file that was expected to
be in ./static/bootstrap/3.3.2/js

Moved the bootstrap.min.css file to the ./static/bootstrap/3.3.2/css
directory.


jbenet:
thanks!

PR #13, opened 2015-02-28
When the webui is running on port 80, it will not have an undefined port

### LINKED ISSUE #4: reverse proxy for webui not connecting
we are running an ipfs gateway on a server.

i'd like to be able to run the API behind a reverse proxy so that we can add basic authentication for the node but expose the gateway to the public. also, i'd like the url to resolve without having to type the ports when i want to check the gateway.

setup: ipfs with API on `0.0.0.0:5001` nginx in front reverse proxying to `ipfsc.erisbootstrap.sexy` ... nginx is working properly for the gateway (see #3 ) and urls to ipfs.erisbootstrap.sexy (which is routed to the gateway port) all resolve fine.

problem: when i connect to ipfsc.erisbootstrap.sexy which is proxied to 5001, the webui loads fine but does not connect to the API properly. it is just a plain UI without any connection details. however when I connect via the IP:Port combo it resolves fine. links below:

* url i'd like to use which won't connect to the api on the node: https://ipfsc.erisbootstrap.sexy/webui
* url which i'm not keen to use which does connect to the api on the node: http://54.93.113.247:5001/webui

travisperson:
`window.location.port` will return `undefined` when no part is specified
on the url. This happens for port 80 as it's a default and not required.

This ends up with undefined being passed to the IPFS API, to which it
responds by using the default 5001.

This fixes #4

### DIFF
--- a/js/views/page.jsx
+++ b/js/views/page.jsx
@@ -4,7 +4,7 @@ var RouteHandler = require('react-router').RouteHandler
 var Link = require('react-router').Link
 
 var host = window.location.hostname
-var port = window.location.port
+var port = window.location.port || 80
 if(localStorage.daemon) {
   host = localStorage.daemon.host
   port = localStorage.daemon.port


jbenet:
> window.location.port will return undefined when no part is specified
on the url. This happens for port 80 as it's a default and not required.

Wonder why they didn't make `window.location.port` default to whatever the browser's default http port is.

jbenet:
thanks! LGTM

Issue #11, opened 2015-02-27
Build step needs to verify that the build directory is empty


travisperson:
If a file is removed from the static folder (or really anywhere) it could be left in the build directory from a previous build. We should remove everything from the build directory prior to doing a new build to make sure we are working with a clean slate.

jbenet:
Yep, +1. should clean, then make when publishing.

PR #14, opened 2015-02-28
Cleans the directory prior to a new building being published

travisperson:
This fixes issue #11 to ensure that prior to doing a publish we have a clean build directory and we don't pull in any extra files.


jbenet:
Great!

PR #23, opened 2015-04-04
Fix globe. Fixes #22

### LINKED ISSUE #22: increase contrast of globe

STRML:
Added a nice world texture and fixed the texture path logic. `window.location.pathname` doesn't include the hash so it is safe to simply append to it.


jbenet:
@STRML oooh! nice! i think we should be able to toggle-- some people like the darker one.

STRML:
I don't know if it's just my screen, but the darker one looks 100% black
to me.

jbenet:
@STRML oh interesting. it looks like this to me:

STRML:
Yeah, does that not look completely black to you? I mean, I see the points designating peers, but the globe itself is just a black ball.

STRML:
With the new texture:

imgur

jbenet:
> Yeah, does that not look completely black to you?

hmm, no, may be your screen?

> With the new texture:

that looks really nice!! how about a simple button toggle on a floating button on the top-right of the globe's div? (like the button's in github's zen view)

STRML:
Interesting. I don't know why it shows up completely black, thought this was a good screen (MBPr). I'll add a button.

jbenet:
> Interesting. I don't know why it shows up completely black, thought this was a good screen (MBPr).

Hmm. am on a MBAir :/  lmk if you figure it out. if it shows up black for lots of users it's bad :(

> I'll add a button.

thanks!

js/pages/connections.jsx
```diff
@@ -17,109 +30,156 @@ module.exports = React.createClass({
...
                   nonce: t.state.nonce++
-                })
-              })
+                });
+              });
             });
           }
-        })
-      })
-    }
```
jbenet: i prefer no semicolons, its ok if some sneak in, but i'm not for changing the style to have semicolons because i dont want to make this codebase a ground for semilcolon wars.

jbenet:
@travisperson would you CR this?

whyrusleeping:
its almost completely black on my screen too, (MBPr)

js/pages/connections.jsx
STRML: I thought AST was pretty universally recognized as "a bad idea". The worst thing IMO is an inconsistent style. I am working on a `lint` branch that will add a precommit hook and a `.jshintrc` (I'm a big fan, cough, of [JSXHint](https://github.com/STRML/JSXHint)). I personally run a very strict linting configuration that has prevented countless errors in my own projects. I think it is very useful to add and if enforced by a precommit hook it is easy to maintain.

jbenet: on semicolons: https://github.com/feross/standard  <-- see the semicolons part.

> The worst thing IMO is an inconsistent style. 

very much agreed!

I'd really like to use https://github.com/feross/standard if it were to work with this repo. It's like `go fmt` and resolves all sorts of silly discussions on style. it's a good idea to standardize it across all js. are you ok with this one?

STRML: :+1: I see there is a sublime text plugin and it works with JSX, so that's fine by me. Especially if it has an auto style fixer.

jbenet: great! also, +1 on pre-commit-hook. want to check it into something like `hooks/` or `dev/` ? I wish github let you set hooks to execute on github, like checking style and failing it. i guess we can use travis...

STRML:
Rebased to remove the merge commit and did a little cleanup. I removed some of the changes that were semicolon-only; we can fix the rest in another changeset that adds `standard`.

js/pages/connections.jsx
STRML: I'm a big fan of [pre-commit](https://www.npmjs.com/package/pre-commit). I use it in all my projects. In combination with a travis.yml we should be good.

I already have a branch in progress. I'll move it from jsxhint to standard.

jbenet: sounds great

jbenet:
@STRML this LGTM. i cant test it locally atm-- can trust that it works. (shipping it requires shipping a go-ipfs update so can test it before then).

whyrusleeping:
the 'peer pillars' fail to show up for me now.
selection_001

whyrusleeping:
hrm, might be a different cause, all the locations on my peers show up as unknown

jbenet:
> hrm, might be a different cause, all the locations on my peers show up as unknown

i think thats adblocking? may  not be able to hit the geocoder

PR #26, opened 2015-04-04
Add `npm run dev`, read PORT env, and automatically open browser from gulp.

STRML:


jbenet:
thanks! LGTM

jbenet:
thanks for the PR @STRML

PR #27, opened 2015-04-05
Add precommit hook and fix feross/standard lint errors.

STRML:
Alright - as per our discussion, we're now complying with `standard`.

Took a while to work out all the lint errors (the fixstyle util doesn't work with jsx) but it should be smooth sailing from now on.

Just a note on the module.exports syntax, React can automatically figure out via JSX what the displayName is if you use `var Foo = React.createClass({})`, but it appears it can't do it if you do `var Foo = module.exports = React.createClass({})`, thus the changes. It's really nice to have the displayName for debugging.

### DIFF
--- a/.sublimelinterrc
+++ b/.sublimelinterrc
@@ -1,0 +1,13 @@
+{
+    "linters": {
+        "jshint": {
+            "@disable": true
+        },
+        "jsxhint": {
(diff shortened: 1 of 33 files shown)


jbenet:
awesome!

PR #29, opened 2015-04-06
Move to Webpack from Gulp.

STRML:
Webpack can take care of all the loading we need. In combination with a simple Makefile we have no need for grunt/gulp.

This commit deletes a lot of files in `/static/js` in favor of npm packages instead so that they can be properly versioned and more easily loaded.

With webpack we get faster hot builds, and even hot reloading of styles and React components (!) using react-hot-loader.

Note that the two commits integrating `standard` are part of this PR - the merge conflicts were too crazy to rebase it out. Once you merge that PR, this one will be properly scoped.


STRML:
BTW the build is working: http://gateway.ipfs.io/ipfs/QmYoxEvtV4B4wtpbre6Sb3Dm2dXGmR3CbT7SpEd6broiHM/#/

API of course is not accessible from there. We could make it so if we added CORS headers to the api and set a configuration option in the webui for the API location, defaulting to `localhost:5001`.

whyrusleeping:
14,195 deletions.... whatever you did. I'm okay with this.

STRML:
Most of that is assets that can be loaded from npm rather than put in the repository directly.

jbenet:
The removal of static assets is not desired. 

The static directory (along with every single published version) is checked in so that we can re-provide any _exact_ published version again. builds are non-reproducible with npm (i.e. you do not get the _exact_ same bits, as lots of versions get updated. it's nearly impossible to make sure the entire tree of versions is locked. Until we have an _extremely_ reliable ipfs seeding service, we must keep things checked in ourselves.

Once the static assets are back in + the diff becomes manageable, i'll take another look

---

On webpack, let's add precise and clear reasoning why to the readme. I personally strongly prefer browserify, and substack and other node people will be helping us out with things along the way (they all prefer browserify too). Since I'm not the maintainer of this repo, i'm fine doing whatever people who want to develop and push for, but want to make sure we have clear reasoning for with advantages, as others will ask to change it back. (one note on this, we strongly prefer modularity over monolithic tools)

STRML:
It's clearly superior for development, and offers a lot better options for
shimming traditional scripts. We can also better optimize the bundle, split
code, etc if we prefer. It has powerful capabilities for replacing modules
depending on environment, etc, etc. It's a wonderful tool.

As for strongly preferring browserify - have you used both, or are you just
more familiar with browserify?

As for modules, I disagree strongly - the previous approach was
inconsistent (some modules in npm, some not). And of course, npm shrinkwrap
was designed for this exact purpose. We should just shrinkwrap tagged
releases.

jbenet:
> It's clearly superior for development

This is a subjective opinion-- let's keep those out of decision making, as everyone will clash somewhere.

> offers a lot better options for
shimming traditional scripts. We can also better optimize the bundle, split
code, etc if we prefer. It has powerful capabilities for replacing modules
depending on environment,

yeah that's o/ what i'm looking for. to switch, i want a list of concrete reasons why this repo uses webpack and not browserify. I think both tools are good enough to work with in most cases, but subtle differences can make teams more productive. outlining those avoids arguments.

> As for strongly preferring browserify - have you used both, or are you just
more familiar with browserify?

I've used both. My opinion is not relevant, but if desired: I strongly disagree with the tooling philosophy with which webpack is architected. It encourages monolithic development, and this in practice ends up with way less reusable code. It also encourages this. To get a sense of how I think, see https://github.com/jbenet/random-ideas/issues/27

> As for modules, I disagree strongly - the previous approach was
inconsistent (some modules in npm, some not). And of course, npm shrinkwrap
was designed for this exact purpose. We should just shrinkwrap tagged
releases.

Not sure what you mean here. when i say modularity, I mean a way of clearly separating codebases into small, manageable pieces. For example, i think this webui could be split into multiple modules. Not about the assets.

On the static assets, i do not want to have shrinkwrapped bundles, it defeats the whole point of storing things with ipfs and leveraging the deduplication it provides. A new release may only change a few files. Currently, browseirfy bundles everything into one file, and I wanted (and still want) to break that apart so that each module is loaded in its own file. While this is _traditionally_ not done in normal websites being served, this webui is intended to be served locally 90% of the time. dedup is much more important as latency gains are irrelevant here.

STRML:
Okay, I'm on my laptop I can get into it with a bit more details:

Regarding webpack vs browserify:

Webpack allows us to:

1. [Hot-reload React components](https://github.com/gaearon/react-hot-loader). Major productivity boost here as we can modify modules quickly while maintaining the current layout and application state
2. Have greater [insight into our builds](https://webpack.github.io/analyse/) and dependency graphs
3. Bundle smaller image assets as [data uris](https://github.com/webpack/url-loader)
4. Easily [split builds](https://webpack.github.io/docs/code-splitting.html) into chunks based on size or on logical separation (e.g. commons module + [one module per page](https://webpack.github.io/docs/optimization.html#multi-page-app)). This means, as you mentioned, that we can easily split our build to the point where new releases may only change a few files, and the majority of loads can reused resources from previous builds.
5. Can easily build bundles targetted toward Node or NW
6. Move chunks into [appCache](https://github.com/lettertwo/appcache-webpack-plugin)
7. [Deduplication support](https://webpack.github.io/docs/list-of-plugins.html#dedupeplugin)
8. [Replace / stub modules](https://webpack.github.io/docs/list-of-plugins.html#contextreplacementplugin) to prevent duplication. For example, one module may depend on lodash, another on underscore, you can intercept the load and just provide lodash or even a smaller module with just the required functionality. Useful also for e.g. [removing locales from moment.js](http://stackoverflow.com/questions/25384360/how-to-prevent-moment-js-from-loading-locales-with-webpack)
9. Keeps libraries from polluting the global object
10. Opens up the ability to render server-side without having to deal with browserify-shim (imo it is such a painful module to get working correctly)
11. Keeps dependencies out of index.html, where they may be overlooked, or dependency order may be brittle, and they cannot be effectively chunked or logically grouped. Additionally if we stop using one of these deps, it will automatically be removed from the build if it is not `require()d` - not so for `<script>` includes in the html, which may be forgotten.


As for js files in the `static` directory, what I meant is that the way webui was doing it previously was ineffective for your goals because it was inconsistent. Some deps were in npm and some were not, and the npm deps were not shrinkwrapped. So we still had the problem of unrepeatable builds.

If we use `npm shrinkwrap`, we can lock down all versions and more or less guarantee repeatable builds. We should be doing that anyway, or we could check in `node_modules`. If we do, then it makes more sense to use the npm modules instead of bundles in `static` - they're easier to `require()` and easier to update, and their dependencies can be deduped.

jbenet:
> As for js files in the static directory, what I meant is that the way webui was doing it previously was ineffective for your goals because it was inconsistent. Some deps were in npm and some were not, and the npm deps were not shrinkwrapped. So we still had the problem of unrepeatable builds.

those deps were being placed in a `bundle.js` inside `static/` no? afair, everything was copied into `static/`. the goal is to use that as a staging area, `ipfs add -r`, and then freeze everything in a directory as an old version.

jbenet:
> If we use npm shrinkwrap, we can lock down all versions and more or less guarantee repeatable builds. We should be doing that anyway, or we could check in node_modules. If we do, then it makes more sense to use the npm modules instead of bundles in static - they're easier to require() and easier to update, and their dependencies can be deduped.

whatever we use we should copy everything into `build/` ipfs add it, and then freeze it at the version dir. we can do better things as the ipfs tooling improves.

STRML:
Ah. That is what already happens.

jbenet:
Hey @STRML -- the PR doesnt merge cleanly. want to rebase?

also given above discussion-- should anything else change, or good to go?

jbenet:
comments @travisperson ?

index.js
```diff
@@ -0,0 +1,27 @@
...
+  res.sendFile('index.html', {root: path.join(__dirname, 'static', 'html')})
+})
+
+app.get('/bundle.js', function (req, res) {
+  res.sendFile('bundle.js', {root: path.join(__dirname, 'build')})
+})
+
+app.use(ipfs_static)
+app.use(express.static(require('path').join(__dirname, 'build')))
+app.use('/static', express.static(require('path').join(__dirname, 'static')))
```
jbenet: Pls only do `requires` at the top of the file

dev.js
```diff
@@ -0,0 +1,45 @@
...
+app.use(ipfs_static)
+app.use(express.static(require('path').join(__dirname, 'build')))
+app.use('/static', express.static(require('path').join(__dirname, 'static')))
+
+var server = app.listen(port, function () {
+  console.log('Static server listening on port', port)
+})
+
+// Create webpack-dev-server on port + 1 (usually 8011)
+var compiler = webpack(require('./webpack-dev-server.config.js'))
```
jbenet: move 4 require statements to top of file.

webpack.config.js
```diff
@@ -0,0 +1,56 @@
...
+      {test: /\/node_modules\/ipfs-api\/index\.js/, loader: 'transform-loader/cacheable?brfs'},
+      // fonts
+      // the url-loader uses DataUrls.
+      // the file-loader emits files.
+      { test: /\.woff(2)?(\?v=[0-9]\.[0-9]\.[0-9])?$/, loader: 'url-loader?limit=10000&minetype=application/font-woff' },
+      { test: /\.(ttf|eot|svg)(\?v=[0-9]\.[0-9]\.[0-9])?$/, loader: 'file-loader' },
+
+      // Shims for browser modules
+      { test: require.resolve('jquery'), loader: 'expose?jQuery' }, // expose to window for libs like bootstrap
+      { test: require.resolve('./static/js/globe'), loader: 'exports?DAT.Globe!imports?THREE=three' }
```
jbenet: this whole file is magic configuration. i already don't like this ...

STRML: It is no different than [browserify-shim](https://github.com/thlorenz/browserify-shim#a-config-inside-packagejson-without-aliases) which would be needed anyway if we wanted to shim the modules in `/static/js` and `require()` them properly.

PR #35, opened 2015-04-27
Fixes #31, allowing unpinning via the webui

### LINKED ISSUE #31: Files added via the webui can not be unpinned via the webui.
Current when adding anything file to the webui they get pinned recursively. Any file you add via `ipfs add` gets pin recursively to protect the file, and all it's blocks, from a `gc`. The `node-ipfs-api` currently does not send a `-r`/`--recursive` when running the `pin rm` command (`api.pin.remove`), so on any file you try to "delete" (the X), will return an error saying it is added recursively.

travisperson:
There are few fixes in this, mostly though this is to allow users to
unpin from the webui. This commit includes a version bump of the
`node-ipfs-api`, as well as passing in the recursive flag.

An unexpected addition to this commit is due to a change in `go-ipfs`
passing back an object of keys instead of an array.

### DIFF
--- a/js/pages/files.jsx
+++ b/js/pages/files.jsx
@@ -12,12 +12,12 @@ var Files = React.createClass({
     function getFiles () {
       t.props.ipfs.pin.list(function (err, pinned) {
         if (err || !pinned) return t.error(err)
-        t.setState({ pinned: pinned.Keys.sort() })
+        t.setState({ pinned: Object.keys(pinned.Keys).sort() })
       })
 
       t.props.ipfs.pin.list('recursive', function (err, pinned) {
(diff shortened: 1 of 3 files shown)


travisperson:
See https://github.com/ipfs/go-ipfs/commit/0a6b880bee0117f657aedbbd24bfc07a2361e3da for change to `js/pages/file.jsx`

jbenet:
@travisperson would love to be able to have some testing on these commits. they look good to me, but we dont have automated testing :(

travisperson:
Definitely, that is something we should focus on in the next sprint, would love to get @STRML in on the discussion as well.

PR #39, opened 2015-04-29
Show files in recursively pinned directories, wrap additions

krl:
This might be a bit controversial, i removed the different tabs partly because the rearrangement of the way the files are queried for. 

At this point, the interface asks for all recursively pinned objects, and pipes them through ipfs.ls, to get the files that actually have displayable names, and lists just those.

This also depends on my small fix to the api, we should probably also do a new release first and fix the version in package.json

### DIFF
--- a/js/pages/files.jsx
+++ b/js/pages/files.jsx
@@ -1,35 +1,36 @@
 var React = require('react')
 var FileList = require('../views/filelist.jsx')
 var LocalStorage = require('../utils/localStorage')
+var _ = require('lodash')
 var $ = require('jquery')
 
 var Files = React.createClass({
   getInitialState: function () {
-    var t = this
 
-    var files = LocalStorage.get('files') || []
-
-    function getFiles () {
-      t.props.ipfs.pin.list(function (err, pinned) {
-        if (err || !pinned) return t.error(err)
-        t.setState({ pinned: Object.keys(pinned.Keys).sort() })
-      })
-
-      t.props.ipfs.pin.list('recursive', function (err, pinned) {
-        if (err || !pinned) return t.error(err)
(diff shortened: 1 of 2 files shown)


krl:
And yes, this also automatically recursive-pins any file added this way

jbenet:
yay for progress!

what about files that are pinned directly? And files not pinned? Idea originally was to have a "cached" tab too that showed all the objects stored locally.

One way to get around the lack of names In some objects is to have a checkbox that filters out the ones lacking dir names? (What's worse, not showing objects that are taking up space, or not showing clean names?)

jbenet:
(May be good to post screenshots of webui changes, so people can follow along without having to download compile and run. (Which is impossible in iOS :( ))

jbenet:
@krl still unsure whether this PR is the right way fwd atm -- a couple screenshots + discuss re comments above? o/   (also relevant to mention what we were discussing today-- namely still showing tabs, with:

- files -- all the roots that are named -- so not proper roots, but the first children of roots i guess. (files inside directories can be accessed via the directory) (this covers the average user's use case: wanting to see files as they added them, with names).
- objects -- lists all the object roots (hashes, not names) -- this covers the power user case of wanting to see what's here.

And then "cached" and "pinned" could be filters on the page. I.e. toggle between "all | pinned | cached". 
i can sketch this out if needed for clarity.

@krl am i being clear saying "roots"? basically mean objects which are not pointed at by any other object.

whyrusleeping:
I think that the files tab should be replaced with the local file browser i propose in #41

jbenet:
:+1:

krl:
deprecated waiting for proper pinning system

PR #44, opened 2015-05-03
Make geoip requests over ipfs

krl:


jbenet:
@krl this is sweet!! i'll merge this in.

PR #45, opened 2015-05-06
make full path default in object view (fixes #42)

### LINKED ISSUE #42: "raw" button breaks when using ipns paths
repro: on the webui, put an ipns path in the search bar, i.e. `/ipns/ipfs.git.sexy`. Then click the "Raw" button on the resulting page, it will redirect you to a gateway link that will be broken as it tries to prepend the path with `/ipfs/` resulting in `/ipfs/ipns/ipfs.git.sexy`

krl:


jbenet:
rebased on top of linted master

jbenet:
:+1:

PR #47, opened 2015-05-07
linting for jsx files

krl:
Did the manual regression testing, everything checks out.


travisperson:
:heart: This is great.

jbenet:
@krl @travisperson what sort of testing should we put in place for the webui?

Issue #49, opened 2015-05-10
Progress feedback, loading animation


krl:
We should have some kind of loading animation, when objects are being resolved etc.

Also, a way to set context timeout to something much lower would be great for ux.

olizilla:
We have some loading spinners now, but we need to do much more to show the user that their request is "going out to a network of peers that are trying to help answer your question, so, like just wait a little longer... I mean, look at all this amazing p2p work going on!" rather "loading plz wait. you are in a queue. your call is important to us"

I'm closing this out in favour of creating some more specific issues on the subject as we get time.

Issue #43, opened 2015-05-03
add providers of option to dag page


whyrusleeping:
It would be really cool if when i load an object in the dag viewer, i could click a button to retrieve a list of providers of the object i'm looking at.

whyrusleeping:
from the CLI, this is done with `ipfs dht findprovs <hash>`

krl:
I have the basics in place to make this happen now, will postpone until after next release.

whyrusleeping:
@krl if you were having trouble with the output of findprovs, try this patch: http://mars.i.ipfs.io:8080/ipfs/QmbTY1cgy33noZU9skA5PHAJeL4fsbLTPwBRfVU3Twz539

and make sure to add stream-channels=true to your query, like: 
`curl localhost:5001/api/v0/dht/findprovs/QmbbowxNyjgNcY9qwJPr4JAGP7A9vdkipoihF5jAkgahiY\?stream-channels=true`

krl:
@whyrusleeping will try it as i set up multi-node api tests, is this part of a pull request somewhere?

PR #50, opened 2015-05-10
Object view error checking and path parsing

krl:
Add a ipfs permalink for ipns data:

ipns-path-permalink

Data does not display a big empty box when empty:
permalink-data

Error handling: (see #49)

error-handling

These things are moving towards #43


krl:
btw, the font issues only appear when i run in dev-mode, looking at this too

jbenet:
right now, the links + data boxes are a bit odd -- doesn't feel like a single object. wonder if pushing them together may be a good visual, something like: 



but with better indicators / headers or whatever

krl:
new-object-links

new-object-data

Some alignment improvements, proposed grouping-of-objects.

krl:
Changed 'go back' into 'Parent object', which is more to the point.

Not using 'window.location.hash = something' anymore, but context.router.transitionTo()

Moved path parsing into its own utils/path.js for use in other parts of the app

jbenet:
@krl these changes looks great to me -- thanks for the screenshots. 


- i really like the combined links + data box. + love that you clearly tell the user that it's empty. :+1:  
- the permalink probably should be outside still, sorry for not being clear. it could also have a "copy" button like other places in the ui.
- i'd prefer -- when there is data -- for the data to occupy the whole box, instead of having the word "Data" and the actual data nested one more box deeper. looks too complex.

For example:



totally right that we need to clarify the difference between the two segments to people, so maybe we could do something like:

image

where the `xml protobuf json` links change the encoding in place, and the "external link" is a link to `/object/data?arg=<key>&enc=<encoding>` (in the api route for now, later on we can see about exposing read only things like `/object/data` to the normal gateway, or something).

### we can probably address these things in a separate PR-- whatever you prefer. I'm good to merge as is or fix.

---

(( disclaimer for my pedantic stylistic nits: sorry to go back and forth on it, since i designed lots of the UX here (along with a designer) i'm going to claim responsibility for it. most of it doesn't matter but some things (even small) like the placement of boxes have strong implications in terms of people's understanding of things. i think right now i have the most cohesive perspective into all the ipfs touch points and can probably make the most informed decisions. i hope to defer all this to a full time designer once we get one. ))

krl:
3objectview

3objectview2

Still not 100% sure about the headings, IANRAD either, but i think this addresses your stylistic comments.

Pedantic is good!

krl:
Btw, json/protobuf/xml viewer is also something i think we should have here.

jbenet:
LGTM

jbenet:
thanks for making the design changes! :+1:

PR #51, opened 2015-05-16
Remove polling interval from files view

### LINKED ISSUE #20: /files with thousands of files added makes my browser hang
I don't know how many files/directories I have added, but as an experiment I have added thousands of them.

Browsing to `http://localhost:5001/ipfs/QmXX7YRpU7nNBKfw75VG7Y1c3GwpSAGHRev67XVPgZFv9R/#/files/all` makes Firefox and Chromium on Linux hang a lot.

krl:
This made the files page go from eating 100% cpu to idle.

I'd argue the relevant time you want to update these listings, is while adding files through the webui, and as this case is already covered without polling, i think we should just drop this line.

Cannot think of any usecase where you would want to have polling updates of pinned files at this point.

Maybe it does not fix #20 in its entirety, but it makes a huge difference in performance.

### DIFF
--- a/js/pages/files.jsx
+++ b/js/pages/files.jsx
@@ -27,7 +27,6 @@ var Files = React.createClass({
     }
 
     getFiles()
-    t.pollInterval = setInterval(getFiles, 1000)
 
     return {
       files: files,


whyrusleeping:
Im totally on board with 100% CPU -> idle. :+1:

jbenet:
Yeah let's do it



—

PR #52, opened 2015-05-18
Set gateway based on window hostname

### LINKED ISSUE #3: raw file links point to bound gateway instead of effective gateway
context: we have an IPFS gateway on a server. gateway is bound to `0.0.0.0:8080`. in front of gateway, nginx is reverse proxying `8080` to `443`. that all works fine. in addition, the API port is bound to `0.0.0.0:5001`. 

when i go to the webui [here](http://54.93.113.247:5001/ipfs/QmaaqrHyAQm7gALkRW8DcfGX3u8q9rWKnxEMmf7m9z515w/#/files) and hover on links to the raw files they resolve to `http://0.0.0.0:8080/ipfs/....` rather than our gateway. 

if there was a way to pass an effective base URL to the webui for links that would be tremendous, or some other way to have the ui understand that our public facing gateway is running on a different URL than what the backend is bound to.

Great work, BTW, the UI looks fantastic.

krl:
fixes #3

### DIFF
--- a/js/views/page.jsx
+++ b/js/views/page.jsx
@@ -24,8 +24,9 @@ var Page = React.createClass({
       if (err || !res) return console.error(err)
 
       var split = res.Value.split('/')
-      var host = split[2], port = split[4]
-      t.setState({ gateway: 'http://' + host + ':' + port })
+      var port = split[4]
+
+      t.setState({ gateway: 'http://' + window.location.hostname + ':' + port })
     })
 
     ipfs.version(function (err, res) {


js/views/page.jsx
```diff
@@ -41,7 +42,7 @@ var Page = React.createClass({
       version: '',
       updateAvailable: false,
       updating: false,
-      gateway: 'http://127.0.0.1:8080'
+      gateway: 'http://127.0.0.1o:8080'
```
jbenet: did this sneak in?

krl:
@jbenet it did sneak in

PR #59, opened 2015-06-02
hot-reload without webpack

krl:
This throws out all the webpack related things, and replaces it with a gulp-like system for hot-reloading of code. @oleavr's bygg!


krl:
NB It does not change how the actual release build is happening.

travisperson:
So much cleaner :heart:

jbenet:
LGTM

PR #65, opened 2015-06-07
add woff2 font file

krl:
My browser is asking for this file as well.


travisperson:
I saw this and started to add the file, but then stopped. I thought I commented on this but I guess I didn't. 

We might look at making this part of the build step as we will probably want to make sure we pull the font files from the version of font-awesome we have installed. This can either be a post-install script or part of the build process to copy over the font files from the npm module.

PR #72, opened 2015-07-16
Pulls correct font-awesome fonts into the static directory

travisperson:
Uses a postinstall script to pull the correct font-awesome fonts
into the static directory.

Updates `three` to `0.67.0` as `0.66.0` has an error which causes
browserify (and probably anything) to error due to a missing newline.

Changes the lint command to work on osx


jbenet:
LGTM. want me to test it?

travisperson:
Ya, would be good to just make sure it all runs and gets setup. Tomorrow I'll be checking to see if there are any side effects to not having the xargs flag. It didn't appear to cause any issues removing it on OS X, but I'm. It certain for Linux.

PR #73, opened 2015-07-25
Modularized webui

krl:
Major rehaul of the webui architecture.

Logs and Config are not in yet.

Config is blocked on making ipfs.add solid in the browser, and logs on changing back the log entry point into the api.


krl:
Ok, i got a mega-conflict here since master changed by one commit while i was reorganizing everything. I could manually incorperate the patch, but i think a rebase will simply not be possible here.

Any git advice here?

jbenet:
Manual is fine, just flag the previous commit so we can make sure it's the same in CR :)



—

jbenet:
(Same/similar)



—

README.md
```diff
@@ -1,14 +1,11 @@
...
 
-
 ## Dev
 
 ```bash
 git clone https://github.com/ipfs/webui
 cd ipfs-webui
-npm install
-# Runs server on port 8000. Set env PORT to change.
+./alldeps.sh
```
jbenet: why remove the makefile? makefile can do this easier. + can have other targets, like publish.

jbenet:
should probably use webcomponents, as we discussed with @diasdavid, but will CR some anyway.

@diasdavid, would be great to get your input here too

apps/connections/index.jsx
```diff
@@ -91,4 +93,9 @@ var Connections = React.createClass({
   }
 })
 
-module.exports = Connections
+ipfsapp.define({
+  init: function (ipfs) {
+    React.render(<Connections ipfs={ipfs} />,
+                 document.getElementById('mount'))
+  }
+})
```
jbenet: if we could match the typical 

```js
module.exports = function() {
}
// or 
module.exports = {} 
```
type of thing, that may be nice. maybe not easy? this may be ok. impressions @diasdavid?

apps/connections/js/connection.jsx
```diff
@@ -1,5 +1,5 @@
 var React = require('react')
-var Peer = require('./peer.jsx')
+var Peer = require('../../../js/views/peer.jsx')
```
jbenet: this is importing across apps. :-1:  shared module

apps/connections/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "standard": "3.3.2"
+  },
+  "ipfs-web-app": {
+    "displayname": "Connections",
+    "sort": 1,
+    "icon": "fa-globe",
+    "build": {
+      "dist/script.js": "browserify -t reactify index.jsx",
+      "dist/style.css": "lessc style/style.less"
+    }
```
jbenet: as we discussed, this may be the job of the parent. 

then again, it is nice to have a definition, for a listing of ipfs apps one can use. 

(i can see benefits of both sides, i guess. @diasdavid ?)

apps/connections/style/style.less
```diff
@@ -0,0 +1,47 @@
+@import "../../../style/common.less";
```
jbenet: cross app import

jbenet:
hard to tell everything that changed becuase the github diff is off and the PR is huge. maybe splitting off the apps into own repos will help scope the CR :+1:

jbenet:
@krl in general like where this is at. if it all works well, that's very nice already! :+1:

apps/connections/index.jsx
diasdavid: I typically see (and do) code like

```
// If the module has a constructor 
exports = module.exports [= <name of func that might be a constructor or main call of this module>]

// and then, for other calls:
exports.anotherThing = function () { }
```

If the module is a facade, I do `exports.<something>` for all of them instead of doing `module.exports = {}` 

This way, if you can declare exported values along the file and you don't have to update the exported interface in two places.

apps/connections/package.json
diasdavid: Sorry, not sure if I understand completely why this obj exists in the package.json, is it kind of the manifest.json one would do for a browser extension of chrome app? If apps become web components, shouldn't we use their way to format?

diasdavid:
@krl any luck with the webcomponetized webui? It would be cool to get that also with https://github.com/ipfs/webui/pull/75 to go-ipfs, so that https://github.com/ipfs/electron-app/issues/23 gets fixed

//cc @whyrusleeping

jbenet:
yeah status here?

diasdavid:
Whats the status here @krl ? ETA to be finished? There are some people looking to help on the webui and it would be awesome if they can contribute on the new Web Componentized version

krl:
@diasdavid hey, sorry for being out of the loop for a bit, i'm going to have a look at this again next week

diasdavid:
@krl nice :) If you have the times to also describe the goal of 'Web Componetizing' all of the things, could be good for people doing work on the other IPFS apps to follow the model. @giodamelio wants to contribute to the webui :)

jbenet:
we should come up with a simple plan to webcomponentize all the things, and
merge this, etc. other people want to help here and maybe we can divide up
the work.

@krl maybe the first next step is to outline the rest of the necessary
steps.

Dignifiedquire:
Also it would be good to have an outline about the tooling especially in relation to react, as from
what I have gathered so far the tooling seems quite problematic at some points especially the integration parts.

diasdavid:
@krl updates on this PR? I eager and excited for the moment we can write tests for the webui

Dignifiedquire:
After the latest discussions I will close this (see notes here: https://github.com/ipfs/pm/issues/55#issuecomment-157145331) Will create issues with detailed todos in the next days, but the gist is that we will use react components for the webui with a simple build process rather than trying to get webcomponents and react married. 

Thanks @krl for all the work and discussions!

PR #78, opened 2015-08-22
Update README.md

drathir:
Correction of port on which currently server listen. Also i guess have problem with env port=<port_number> , env port="<port_number>"  none of them working, but need to be confirmed by someone more skilled...


jbenet:
thanks @drathir

PR #82, opened 2015-09-22
Bypass CORS in development by proxying `Origin` headers correctly

masylum:
#### :tophat: What? Why?

`go-ipfs` does some kind of poor man's CORS by checking the `Origin` header.
Current web-ui dev build adds correctly `referer` header but not the origin one.

#### How to test

Adding files in development shouldn't return `403` anymore.

### DIFF
--- a/byggfile.js
+++ b/byggfile.js
@@ -11,12 +11,17 @@ var rename = require('bygg-plugins-light/rename')
 var proxy = require('proxy-middleware')
 var url = require('url')
 
+var HOST = 'http://localhost:5001'
+
 bygg.task('serve', function () {
   return build()
     .pipe(bygg.write('build/'))
     .pipe(serve(3000, function (app) {
-      var opts = url.parse('http://localhost:5001/api')
-      opts.headers = {'referer': 'http://localhost:5001/'}
+      var opts = url.parse(HOST + '/api')
+      opts.headers = {
+        referer: HOST + '/',
+        origin: HOST
+      }
       return app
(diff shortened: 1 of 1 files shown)


jbenet:
> Current dev build adds correctly `referer` header but not the origin one.

Is this the current go-ipfs build?  (@rht could you fix this?)

jbenet:
@diasdavid @dignifiedquire mind CRing this?

masylum:
This is against: `ipfs version 0.3.8-dev`

Dignifiedquire:
lgtm.

@jbenet any idea how I can pull this into electron-app to test if this solves the issue I was having there?

diasdavid:
If I understand correctly, this is a fix for only when we run a webui in test mode (from a local webserver) and not from IPFS, right? We can't fake origin header when running from a browser and with no proxy available.

masylum:
This is only for the development environment. `go-ipfs` is trying to validate the port `5001` by checking the `Origin` header. Since the web app in development mode runs in `3000` the calls to the API return a `403`. Same issue probably affects `electron-app` indeed.

jbenet:
btw, the daemon can be run with CORS settings or disabled (see `ipfs daemon --help). though this probably makes sense for ease of user.

regardless, should we be fixing the origin header in go-ipfs? are missing it over there?

masylum:
LGTM or close the PR. This is only for the development environment and makes it more enjoyable to have a working environment by default. It's not obvious nor friendly to have to disable CORS for development.

PR #86, opened 2015-10-02
Updates the 'back to console home' link in the 404

### LINKED ISSUE #79: 404 error doesn't link back to webui
I'm not quite sure if this is intentional or not, but when I enter a hash in the search bar of the webui that doesn't exist, the link to that let's you "Go to console home" sends you to `http://localhost:5001` rather than `http://localhost:5001/webui` resulting in another 404 error.

CaioAlonso:
Fixes  #79.


jbenet:
We can likely ship a new version of webui with `0.3.8` cc @whyrusleeping

PR #93, opened 2015-11-09
Use lookupPretty from ipfs-geoip

Dignifiedquire:


jbenet:
hm we should have tests in this repo. can someone else take a look at this?

diasdavid:
This one is straight forward as it factors out geoip. For future, more significant changes, we should have tests first.

Dignifiedquire:
@diasdavid we all dream of tests, don't we..

PR #118, opened 2015-12-07
German localization

digital-dreamer:
This is a complete German localization of the UI by @Dignifiedquire. Two other localizations are currently under constriction on Transifex, so stay tuned for more.


Dignifiedquire:
LGTM

---

@digital-dreamer is there a better way to keep things synchronized between this repo and Transifex?

jbenet:
thanks!

Issue #140, opened 2015-12-24
Convert all bootstrap components to use react-bootstrap


Dignifiedquire:
There is a lot of use of things like `<div class="some-bootstrap-class"></div>`. All these should be removed in favor of using [react-bootstrap](https://github.com/react-bootstrap/react-bootstrap) components.

RichardLitt:
Should this be off of master or webpack?

RichardLitt:
@Dignifiedquire What do you think of https://github.com/RichardLitt/webui/commit/476b427549cde7d0a3ac17f91e5eab3e454014fd. I think that is what you were going for. Let me know, and I'll push it on top of the `webpack` branch.

Dignifiedquire:
@RichardLitt yes that looks good, I did some more changes last night while traveling, so please rebase onto the latest `webpack` branch, and make it a PR against that branch so we have a nice way of tracking these individual changes :)

RichardLitt:
Should be closed now.

PR #142, opened 2015-12-27
Converted majority of BS elements

RichardLitt:
Converted most of the BS elements to ReactBootstrap, as per #140. I did not do all of them - I am not sure how Button would react, for instance, and I want to make sure I do it right before I mess something up.


RichardLitt:
@Dignifiedquire Fixed linting error.

Dignifiedquire:
:+1:

RichardLitt:
Passed. This should be merged before more work is done on the `webpack` branch.

Dignifiedquire:
Done :)

PR #143, opened 2015-12-29
Update README with Access-Control headers for development

travisperson:
Adds required configurations to run the WebUI on port 3000 so that it can properly make request to the running daemon.


Dignifiedquire:
Thanks

Issue #133, opened 2015-12-21
Remove jquery


Dignifiedquire:
We should drop jquery as a dependency and replace all instances with alternatives. For all things bootstrap we should use react-bootstrap and drop bootstrap with jquery

- [x] `scripts/pages/files.js:5:16`
- [x] `scripts/pages/objects.js:3:16`
- [x] `scripts/views/config.js:3:16`
- [x] `scripts/views/editable.js:2:16`
- [x] `scripts/views/filelist.js:2:16`
- [x] `scripts/views/page.js:5:16`

diasdavid:
@travisperson made this on https://github.com/ipfs/webui/pull/159 🎉

Dignifiedquire:
Please do not close until it's merged :)

luigiplr:
Doing in: #195

Dignifiedquire:
Fixed in #159

RichardLitt:
Thanks!

PR #150, opened 2015-12-31
Removed jQuery from linkHandler

RichardLitt:
Don't have time for much more today, but this should work. I'm not sure how to test it, though. Related to #133

### DIFF
--- a/app/scripts/linkhandler.js
+++ b/app/scripts/linkhandler.js
@@ -1,12 +1,11 @@
 // import Router from 'react-router'
-import $ from 'jquery'
 
 export default function linkHandler () {
   var external = /^(https?:)?\/\//i
   var staticLink = /^\/static\//i
 
-  $(document.body).on('click', 'a', function (event) {
-    var href = $(this).attr('href')
+  document.body.getElementByTagName('a').onclick = function (event) {
+    var href = this.getAttribute('href')
     console.log(href)
 
     // pass through if external or static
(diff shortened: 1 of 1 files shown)


Dignifiedquire:
I think we can actually drop this linkhandler module completely if we move to the latest react-router

Dignifiedquire:
Actually it's already commented out and not required, just pushed a removal of it. Sorry :blush:

RichardLitt:
@Dignifiedquire I've seen this sort of linkhandler a few places. Can you explain why it was necessary?

RichardLitt:
I keep running across this thing, anyway, so I made it into the [react-linkhandler](https://github.com/RichardLitt/react-linkhandler) module.

Dignifiedquire:
I actually don't understand why it was necessary. For all links that should be handled by `react-router` one uses `import {Link} from 'react-router'` and for all regular links you use `<a href='..'></a>` as in regular html

RichardLitt:
@jbenet You originally created this. Can you explain why?

RichardLitt:
@jbenet put it in Beagle too. I just modularized it there, don't know why I didn't do that earlier. I don't think I was as modular-friendly as I am now. Cf: https://github.com/BeagleLab/beagle/blob/master/js/main.js#L11

Dignifiedquire:
Strange, makes no sense to me as explained above

Issue #151, opened 2015-12-31
Introduce good logging


dignifiedquire:
To make debugging and tracing easier we should ~~replace `debug` with [bows](https://github.com/latentflip/bows) and~~ add good logs everywhere using debug were it makes sense.

dignifiedquire:
Debug is integrated and at the moment we don't have major debugging/logging issues so closing for now.

PR #152, opened 2016-01-01
refactor: Improve style setup

Dignifiedquire:


app/styles/bootstrap.less
```diff
@@ -0,0 +1,49 @@
...
+@import "../../node_modules/bootstrap/less/button-groups";
+@import "../../node_modules/bootstrap/less/input-groups";
+@import "../../node_modules/bootstrap/less/navs";
+@import "../../node_modules/bootstrap/less/navbar";
+@import "../../node_modules/bootstrap/less/breadcrumbs";
+@import "../../node_modules/bootstrap/less/pagination";
+@import "../../node_modules/bootstrap/less/pager";
+@import "../../node_modules/bootstrap/less/labels";
+@import "../../node_modules/bootstrap/less/badges";
+@import "../../node_modules/bootstrap/less/jumbotron";
```
diasdavid: we need `jumbotron` and all the others here?

app/scripts/app.js
```diff
@@ -3,6 +3,8 @@ import ReactDOM from 'react-dom'
 import Router from 'react-router'
 import routes from './routes'
 
+require('../styles/app.less')
+
 if (process.env.NODE_ENV !== 'production') {
   window.uiDebug = require('debug')
```
diasdavid: this could be just `window.debug` so that it becomes clear that it is just our debugging thing.

diasdavid:
overall LGTM :)

app/styles/bootstrap.less
Dignifiedquire: This is a 1:1 copy of the requires from bootstrap itself, I have not removed any unused code yet.

app/scripts/app.js
Dignifiedquire: That depends if we want to bundle the whole `debug` module, as the current build system will actually not bundle with this setup

app/styles/bootstrap.less
Dignifiedquire: Moved to https://github.com/ipfs/webui/issues/153

app/scripts/app.js
Dignifiedquire: Also, that will be improved when we do #151

PR #154, opened 2016-01-01
tests: Add example test for react component

Dignifiedquire:
cc @RichardLitt @diasdavid


app/scripts/views/config.js
```diff
@@ -11,7 +11,7 @@ export default React.createClass({
 
   getInitialState: function () {
     return {
-      body: JSON.stringify(this.props.config, null, '  '),
+      body: JSON.stringify(this.props.config, null, 2),
```
travisperson: This just blew my mind a little bit, never knew you could specify a number instead of a string.

Dignifiedquire: :)

package.json
```diff
@@ -49,6 +49,7 @@
     "karma-mocha": "^0.2.1",
     "karma-sourcemap-loader": "^0.3.6",
     "karma-webpack": "^1.7.0",
+    "legit-tests": "^0.6.3",
```
diasdavid: so legit!

diasdavid:
This looks nice! Can we also have a test example for a piece of functionality?

Dignifiedquire:
@diasdavid what kind of functionality do you mean?

diasdavid:
Right now the test is to see if it the view is rendered correctly, right? I guess that every test can be performed by an amount of assertions on the elements that were rendered, but for things like "Check if the files uploaded list is correct" could be tested without the UI part?

Dignifiedquire:
Closing in favor of #156

PR #155, opened 2016-01-01
feat: Replace console.log with debug

Dignifiedquire:
This is the setup for #151. The actual good logging messages still need to be written at a later point

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,7 @@
 {
   "parser": "babel-eslint",
-  "extends": ["standard", "standard-react"]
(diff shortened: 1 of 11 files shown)


Dignifiedquire:
After some discussion on IRC we might go with `debug` needs more research

Dignifiedquire:
Changed to `debug` as that seems to be the choice these days and it supports `stderr` in node.

Dignifiedquire:
@RichardLitt can you review this one as well and merge if you are happy?

RichardLitt:
Everybody wins.

PR #156, opened 2016-01-02
refactor: Upgrade to react-router@2

### LINKED ISSUE #131: Upgrade react-router to 2.0
See https://github.com/rackt/react-router/blob/master/CHANGES.md for details on what to do

Dignifiedquire:
This also upgrades the test story to be more
useful as I found some issues with the one before.
So this superseeds #154.

Closes #131


Dignifiedquire:
Tests pass locally in Chrome, fail only in firefox because of https://github.com/bevry/sortobject/pull/2

app/scripts/views/icon.js
```diff
@@ -0,0 +1,17 @@
+import React, {PropTypes} from 'react'
```
RichardLitt: Why not `{React, PropTypes}`?

RichardLitt: Or just React?

Dignifiedquire: Because that's something different. To use jsx you need to have the top level `React` in scope, and for the `PropTypes` I don't want to write `React.PropTypes`. For more details see [this article](http://www.2ality.com/2014/09/es6-modules-final.html)

app/scripts/views/nav.js
```diff
@@ -1,42 +1,35 @@
...
   displayName: 'Nav',
+
+  contextTypes: {
+    router: React.PropTypes.object.isRequired
+  },
+
   render: function () {
     return (
       <div className='row'>
         <ul id='side' className='nav nav-sidebar'>
```
RichardLitt: Should we be using react-bootstrap for nav and nav-sidebar?

Dignifiedquire: I haven't found `nav-sidebar` on react-bootstrap yet

app/scripts/views/icon.js
RichardLitt: Hmm, interesting. Interesting shorthand I hadn't seen before. Thanks.

app/scripts/views/nav.js
RichardLitt: Noted.

RichardLitt:
This all looks good to me, from what I understood of it.

RichardLitt:
You! You're on :fire:

PR #159, opened 2016-01-03
Remove jquery

### LINKED ISSUE #133: Remove jquery

travisperson:
This strips out all the jquery used throughout the webui.

Closes #133


diasdavid:
👏👏👏👏👏👏👏👏 :D

app/scripts/views/config.js
```diff
@@ -113,7 +111,7 @@ export default React.createClass({
         {error}
         {buttons}
         <div className='textarea-panel panel panel-default padded'>
-          <textarea className='panel-inner' spellCheck='false' onChange={this.handleChange} value={this.state.body} />
+          <textarea ref='textareaConfig' className='panel-inner' style={{height: this.state.height}} spellCheck='false' onChange={this.handleChange} value={this.state.body} />
```
Dignifiedquire: Can you add line breaks here please, this line is getting a bit long

app/scripts/views/filelist.js
```diff
@@ -41,43 +35,61 @@ export default React.createClass({
...
+            file = { id: file }
           }
 
-          var gatewayPath = t.props.gateway + '/ipfs/' + file.id
-          var dagPath = '#/objects/' + file.id
-
-          var tooltip = (
-            <Tooltip id={file.id}>{i18n.t('Remove')}</Tooltip>
-          )
+          var dagPath = `#/objects/${file.id}`
```
Dignifiedquire: let's use `const` and `let` while we are at it

app/scripts/views/filelist.js
```diff
@@ -41,43 +35,61 @@ export default React.createClass({
...
+          var gatewayPath = `${this.props.gateway}/ipfs/${file.id}`
+          var unpin = (e) => {
+            e.preventDefault()
+            this.props.ipfs.pin.remove(file.id, {r: true}, function (err, res) {
+              console.log(err, res)
```
Dignifiedquire: Please use `debug` for logging

app/scripts/views/filelist.js
```diff
@@ -41,43 +35,61 @@ export default React.createClass({
...
   }
-})
+}
+
+FileList.propTypes = {
+  ipfs: React.PropTypes.object,
+  files: React.PropTypes.array,
+  namesHidden: React.PropTypes.bool,
+  gateway: React.PropTypes.string
+}
```
Dignifiedquire: You can write propTypes inline:

```js
class MyClass extends Component {
  static propTypes = {
    // .. my prop types
  }
}
```

app/scripts/views/filelist.js
```diff
@@ -41,43 +35,61 @@ export default React.createClass({
...
+
+const FileItem = ({gatewayPath, dagPath, file, unpin}) => {
```
Dignifiedquire: Let's put each component into a separate file please

Dignifiedquire:
Thanks @travisperson for doing this, I left some nit picks inline

Dignifiedquire:
Another note, when do not require jquery anymore we can drop it from the deps in `package.json`

app/scripts/views/filelist.js
```diff
@@ -41,43 +35,61 @@ export default React.createClass({
...
+      <td className='id-cell'><code>{file.id}</code></td>
+      <td className='action-cell'>
+        <a target='_blank' href={gatewayPath}>{i18n.t('RAW')}</a>
+        <span className='separator'>|</span>
+        <a href={dagPath}>{i18n.t('DAG')}</a>
+        <span className='separator'>|</span>
+        <a href='#' onClick={unpin}>
+          <OverlayTrigger placement='right' overlay={tooltip}>
+            {/* The block element is required otherwise the overlay is *over* the icon */}
+            <div style={{display: 'inline-block'}}><i className='fa fa-remove'></i></div>
```
Dignifiedquire: You can use `<Icon glyph='remove' />`

app/scripts/pages/files.js
```diff
@@ -63,14 +59,12 @@ export default React.createClass({
...
     console.log('dragleave')
     this.setState({ dragging: false })
-    $(e.target).removeClass('hover')
     e.stopPropagation()
     e.preventDefault()
   },
 
   onDrop: function (e) {
     this.setState({ dragging: false })
-    $(e.target).removeClass('hover')
```
RichardLitt: Won't removing `hover` affect the CSS?

RichardLitt:
LGTM, given the things above. :+1:

app/scripts/pages/files.js
travisperson: The only css the defiend a `hover` class was for an element matching the following selector `.file-add-container .file-add-container-inner` which none of these effected. The `hover` class for that is handled on line `154` (old code) or `148` (new code).

app/scripts/views/filelist.js
```diff
@@ -1,34 +1,26 @@
-import React from 'react'
-import $ from 'jquery'
+import React, { Component } from 'react'
 import i18n from '../utils/i18n.js'
-import {Table, Tooltip, OverlayTrigger} from 'react-bootstrap'
+import {Table} from 'react-bootstrap'
+import FileItem from './fileitem'
+import Debug from 'debug'
```
travisperson: I'm not a fan of this, though I saw it used another place (Caps should be reserved for constructors)

app/scripts/views/filelist.js
```diff
@@ -41,43 +33,28 @@ export default React.createClass({
...
-          var tooltip = (
-            <Tooltip id={file.id}>{i18n.t('Remove')}</Tooltip>
-          )
+          const dagPath = `#/objects/${file.id}`
+          const gatewayPath = `${this.props.gateway}/ipfs/${file.id}`
+          const unpin = (e) => {
+            e.preventDefault()
+            this.props.ipfs.pin.remove(file.id, {r: true}, (err, res) => {
+              if (err) {
+                debug(err.message)
```
travisperson: Do we want to debug log errors we are not properly handling (displaying to the user), or simply just `console.error` them so they are more transparent and can be seen when they occure? I saw this as well in the PR just merged for logging.

app/scripts/views/filelist.js
Dignifiedquire: Agreed that's why I switched to using lower case and then const log = debug('mycomponent')

package.json
```diff
@@ -13,7 +13,6 @@
     "i18next-sprintf-postprocessor": "^0.0.6",
     "i18next-xhr-backend": "^0.0.5",
     "ipfs-api": "^2.10.0",
-    "jquery": "^2.1.4",
```
Dignifiedquire: whoooo :party:

whyrusleeping:
whats the plan for doing basic arithmetic without jQuery? will we need to require some other framework for that?

Dignifiedquire:
if you write JavaScript you only need the force and no arithmetic

app/scripts/pages/files.js
```diff
@@ -48,14 +46,12 @@ export default React.createClass({
 
   addFile: function (e) {
     e.preventDefault()
-    $(ReactDOM.findDOMNode(this)).find('.file-select').click()
-    return
+    this.refs.fileSelect.click()
```
Dignifiedquire: I don't understand why need to trigger a `click` event on the input field Oo

app/scripts/views/fileitem.js
```diff
@@ -0,0 +1,44 @@
...
+      <td className='filelist-name'><a target='_blank' href={gatewayPath}>{file.name}</a></td>
+      <td className='id-cell'><code>{file.id}</code></td>
+      <td className='action-cell'>
+        <a target='_blank' href={gatewayPath}>{i18n.t('RAW')}</a>
+        <span className='separator'>|</span>
+        <a href={dagPath}>{i18n.t('DAG')}</a>
+        <span className='separator'>|</span>
+        <a href='#' onClick={unpin}>
+          <OverlayTrigger placement='right' overlay={tooltip}>
+            <Glyphicon glyph='remove' />
```
Dignifiedquire: I meant `Icon` in `app/scripts/views/icon.js` as that uses font awesome and I think Glyphicon uses the bootstrap icon font which we don't use.

app/scripts/views/filelist.js
```diff
@@ -41,43 +33,28 @@ export default React.createClass({
...
         </thead>
         <tbody>
-        {files ? files.map(function (file) {
-          if (typeof file === 'string') file = { id: file }
-
-          var type = '?'
-          if (file.name) {
-            var lastDot = file.name.lastIndexOf('.')
-            if (lastDot !== -1) type = file.name.substr(lastDot + 1, 4).toUpperCase()
+        {files ? files.map(file => {
```
Dignifiedquire: Can we please this long code in `map` in a function instead of inline in jsx?

app/scripts/pages/files.js
travisperson: Honestly not sure if it's required... I didn't check lol. I just changed it to not use jquery. What it's doing though is the button exposed to click is just a regular button, and the file input is hidden. Without this I'm taking a guess that you will only be able to drag and drop.

There might be a way around, but nothing that I know of off the top of my head.

app/scripts/views/fileitem.js
travisperson: So I tried this and the overlay no longer worked for some reason.

app/scripts/views/filelist.js
travisperson: Alright sounds good. Is this a function you'd want to define on the class or just inside the render method? I don't see much of a reason to define it on the class.

app/scripts/views/filelist.js
travisperson: @Dignifiedquire

app/scripts/views/fileitem.js
Dignifiedquire: Interesting, will check that out

app/scripts/pages/files.js
Dignifiedquire: I see, lets leave it for now

app/scripts/utils/fileExtention.js
```diff
@@ -0,0 +1,14 @@
+export default function fileExtention (name, defaultExt = '?') {
```
Dignifiedquire: should probably be fileExtension, also lets use file-extension.js for the filename, some of us have cas insensitive filesystems, right @whyrusleeping ;)

whyrusleeping: i think a better solution would be to disable support for case insensetive filesystems

Dignifiedquire:
@tperson do you think you can finish this soonish?

travisperson:
I've been waiting on comments from you for a few things. This is in the `webui` repo so you can take it over if you just want to update it with the changes you think are required.

Dignifiedquire:
Thanks @travisperson I finished it up for now so we can merge it :)

PR #160, opened 2016-01-03
Added a log

RichardLitt:
As noted by @diasdavid in https://github.com/ipfs/webui/pull/130#discussion_r48386312

### DIFF
--- a/app/scripts/utils/localStorage.js
+++ b/app/scripts/utils/localStorage.js
@@ -1,4 +1,8 @@
 import _ from 'lodash'
+import debug from 'debug'
+
+const log = debug('utils:localStorage')
+const warn = debug('utils:localStorage:warn')
 
 // Utility to make interacting with localstorage less painful.
 var ls = process.browser && window.localStorage
@@ -30,11 +34,13 @@ function doOperation (op) {
     if (op === 'getItem') {
       try {
         result = JSON.parse(result)
-      } catch (e) {}
+      } catch (e) {
+        log('Unable to parse result as json. Message: %s', e.message)
+      }
     }
 
     return result
   } catch (e) {
-    console.warn('LocalStorage error: ' + e.message)
+    warn(e.message)
   }
 }


app/scripts/utils/localStorage.js
```diff
@@ -30,7 +30,9 @@ function doOperation (op) {
     if (op === 'getItem') {
       try {
         result = JSON.parse(result)
-      } catch (e) {}
+      } catch (e) {
+        console.log('LocalStorage error: Unable to parse result as json. Message: ' + e.message)
```
Dignifiedquire: Please use `debug` for logging

RichardLitt:
How does this look?

app/scripts/utils/localStorage.js
```diff
@@ -1,4 +1,7 @@
 import _ from 'lodash'
+import Debug from 'debug'
+
+const debug = Debug('ipfs:utils:LocalStorage')
```
Dignifiedquire: Can we use the convention I have in #155, i.e.

```js
import debug from 'debug'

const log = debug('utils:localStorage')
```

RichardLitt:
Sure. I got that from /Users/richard/src/ipfs-webui/app/scripts/views/globe.js, which doesn't seem to have the same convention. Maybe we should make a styleguide? Also, there's a console.warn on the same page...

Dignifiedquire:
@RichardLitt yeah that all should be fixed in #155

app/scripts/utils/localStorage.js
```diff
@@ -30,11 +33,13 @@ function doOperation (op) {
...
-      } catch (e) {}
+      } catch (e) {
+        log('Unable to parse result as json. Message: %s', e.message)
+      }
     }
 
     return result
   } catch (e) {
-    console.warn('LocalStorage error: ' + e.message)
+    console.warn(e.message)
```
Dignifiedquire: Let's use `debug` here as well

Dignifiedquire: But with a warn level: `const warn = debug('utils:localStorage:warn')` or sth like that

RichardLitt: Done.

Dignifiedquire:
LGTM :)

PR #166, opened 2016-01-05
fix(object) Ensure keys for all rendered array elements

### LINKED ISSUE #161: Fix react warning on DAG page
Going to http://localhost:3000/#/objects/QmVtU7ths96fMgZ8YSZAbKghyieq7AjxNdcqyVzxTt3qVe?_k=yv8iqf results in

```
Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `ObjectView`. See https://fb.me/react-warning-keys for more information.
```

Dignifiedquire:
Closes #161

Missing:

- [x] Move small components into individual files
- [x] Write tests


app/scripts/views/object/object-links.js
```diff
@@ -0,0 +1,39 @@
...
+        </tbody>
+      </Table>
+    </div>
+  )
+}
+
+ObjectLinks.propTypes = {
+  path: PropTypes.instanceOf(Path).isRequired,
+  links: PropTypes.arrayOf(PropTypes.shape({
+    Hash: PropTypes.string.isRequired,
```
diasdavid: upper `H`?

app/scripts/views/object/parent-link.js
```diff
@@ -0,0 +1,22 @@
+import React, {PropTypes} from 'react'
+import {Link} from 'react-router'
+
+import Path from '../../utils/path'
+import i18n from '../../utils/i18n'
+import Icon from '../icon'
+
+const ParentLink = ({parent}) => {
+  if (!parent) return <span></span>
```
diasdavid: isn't there a `()` missing here? such as return (<span></span>)

test/utils/path.spec.js
```diff
@@ -1,8 +1,48 @@
 import {expect} from 'chai'
-import {parse} from '../../app/scripts/utils/path'
+import Path, {parse} from '../../app/scripts/utils/path'
```
diasdavid: what does `import {}` mean?

app/scripts/views/object/object-links.js
Dignifiedquire: that's what we get from the js-ipfs-api

app/scripts/views/object/parent-link.js
Dignifiedquire: no need for parens on one one line jsx expressions

test/utils/path.spec.js
Dignifiedquire: All the gory details: http://www.2ality.com/2014/09/es6-modules-final.html

short version:

```js
import Path, {parse} from 'some'
// translates roughly into
var Path = require('some')
var parse = Path.parse
```

travisperson: It's variable destructing from the module, allows you to pull named exports out.

```javascript
// in path.js
export function parse() {}
```

Which is like doing `exports.parse = parse`

.babelrc
```diff
@@ -1,5 +1,8 @@
 {
   "presets": ["react", "es2015", "stage-0"],
+  "plugins": [
+    'transform-runtime'
```
RichardLitt: Is this a node module? If not, where is it loaded from?

Dignifiedquire: This is a node module called `babel-plugin-transform-runtime` which is loaded by babel on its own, same with the presets.

app/scripts/utils/path.js
```diff
@@ -1,26 +1,28 @@
...
     if (fragment[0] !== '/') {
       fragment = '/' + fragment
     }
-    return new Path(protocol, name, this.path + fragment)
+    return new Path(this.protocol, this.name, this.path + fragment)
   }
 
-  this.toString = function () {
+  toString () {
     return '/' + this.protocol + '/' + this.name + this.path
```
RichardLitt: Can we have an example of how this looks? Shouldn't there be a divider between name and path?

app/scripts/utils/path.js
```diff
@@ -1,26 +1,28 @@
-function Path (protocol, name, path) {
-  this.protocol = protocol
-  this.name = name
-  this.path = path
+export default class Path {
```
RichardLitt: I'm guessing this is all how new classes are defined in ES6. Otherwise, all of it seems odd.

app/scripts/utils/path.js
Dignifiedquire: Look at the tests for details but I have not changed the functionality from the original here. The divider is added in the parse step to `this.path`

app/scripts/utils/path.js
Dignifiedquire: yes this is ES2015 class syntax: http://www.2ality.com/2015/02/es6-classes-final.html

app/scripts/views/icon.js
```diff
@@ -1,17 +1,20 @@
 import React, {PropTypes} from 'react'
+import classNames from 'classnames'
+
+function Icon ({glyph, large = false}) {
```
RichardLitt: Also guessing that `large = false` is how you enable default args?

Dignifiedquire: yes

app/scripts/views/object.js
```diff
@@ -1,120 +1,42 @@
...
-            {this.props.permalink.toString()}
-          </Link>
-        </li>
-      : null
+export default class ObjectView extends Component {
+  static propTypes = {
+    path: PropTypes.instanceOf(Path).isRequired,
+    permalink: PropTypes.instanceOf(Path),
+    gateway: PropTypes.string.isRequired,
+    object: PropTypes.shape({
```
RichardLitt: `.shape` = ducktyping?

app/scripts/views/object/display-data.js
```diff
@@ -0,0 +1,40 @@
+import React, {PropTypes} from 'react'
+
+import i18n from '../../utils/i18n'
+import RawData from './raw-data'
+
+const DisplayData = ({data}) => {
+  const size = data && data.length - 2
+
+  if (!size) {
```
RichardLitt: Size is not the best const name. I would take the const out entirely, and just use `(data && data...)`

app/scripts/views/object.js
Dignifiedquire: pretty much yes, it validates that the listed keys + types exist on the object

app/scripts/views/object/display-data.js
Dignifiedquire: We need the length down later. Also why do you think it's a bad variable name?

app/scripts/views/object/link-buttons.js
```diff
@@ -0,0 +1,33 @@
...
+import ParentLink from './parent-link'
+
+const LinkButtons = ({gateway, path}) => {
+  return (
+    <div className='link-buttons'>
+      <ParentLink parent={path.parent()} />
+      <a
+          href={gateway + path}
+          target='_blank'
+          className='btn btn-info btn-second'
```
RichardLitt: Can we use react-bootstrap here?

app/scripts/views/object/perma-link.js
```diff
@@ -0,0 +1,24 @@
+import React, {PropTypes} from 'react'
+import {Link} from 'react-router'
+
+import i18n from '../../utils/i18n'
+import Path from '../../utils/path'
+
+const PermaLink = ({url}) => {
+  if (!url) return <span></span>
```
RichardLitt: Why return `<span></span>` at all?

test/views/object.spec.js
```diff
@@ -0,0 +1,40 @@
...
+import DisplayData from '../../app/scripts/views/object/display-data'
+import PermaLink from '../../app/scripts/views/object/perma-link'
+
+describe('Object', () => {
+  it('renders', () => {
+    const path = parse('/ipfs/hello/world')
+    const permalink = parse('/perma/hello/world')
+    const el = shallowRender(
+      <ObjectView
+          path={path}
```
RichardLitt: Generally, I try to alphabetize attributes in newline-delimited lists like these. It makes lookup much easier. Just a thought.

app/scripts/utils/path.js
RichardLitt: Cool.

.babelrc
RichardLitt: Cool.

app/scripts/utils/path.js
RichardLitt: Seems weird to add dividers in two separate places.

app/scripts/views/icon.js
RichardLitt: Thanks!

app/scripts/views/object/display-data.js
RichardLitt: Because you're not just asking about size, you're also checking for existence. `isData` would be better. Also, curious what the `data.length - 2` does -- what if `data == 2`? Won't that return 0, and therefore false?

RichardLitt:
Overall, looks good to me, from an overview of the code and not from an I-tried-to-run-this perspective. Happy to merge it.

app/scripts/views/object/perma-link.js
Dignifiedquire: Because otherwise react is unhappy, as this is a functional component, returning `null` or `''` does not work and react throws an error.

app/scripts/views/object/link-buttons.js
Dignifiedquire: done :)

app/scripts/views/object/perma-link.js
RichardLitt: Cool. Noted.

app/scripts/views/object/display-data.js
Dignifiedquire: Yes and that's intended as far as I understand, because these two bytes are prefixes to the ipfs object

PR #217, opened 2016-01-20
fix(files): Remove use of localstorage.

### LINKED ISSUE #32: Remove local storage
today, when files are added via webui they get put on local storage. 

this should no longer be the case. we should not be using browser local storage for this, as it gets inconsistent with the ipfs repo itself. all[1] state should go through the api. 

[1] i think all. post if you think any should remain in local storage. possible that some user preferences -- now i think about it i think @STRML made it a preference to pick a particular globe. (sorry i haven't published that, was waiting on some bugfixes)

Dignifiedquire:
As discussed this removes the local tab in favor of the
all and pinned tabs. Also makes the drop area for files
visible at all times.

Closes #32

@jbenet @whyrusleeping please take a look if this is okay for listing files for now or if there are better methods.


whyrusleeping:
I think thats good for now. It would be cool to list things inside `ipfs files` in the future

Dignifiedquire:
@whyrusleeping why don't you open an issue for that :P

jbenet:
How does it list all? `ipfs refs local` ?

jbenet:
:+1: for removing local storage

Dignifiedquire:
@jbenet this is how it does it currently: 
* pinned: https://github.com/ipfs/webui/blob/fix/remove-local-storage/app/scripts/pages/files.js#L35
* all: https://github.com/ipfs/webui/blob/fix/remove-local-storage/app/scripts/pages/files.js#L42

(that was like this before) would `refs local` be better? What's the exact difference?

jbenet:
this is fine i think

jbenet:
- a `files` tab may want to list the `ipfs files` root
- i havent looked extensively at the code-- @diasdavid maybe?

jbenet:
LGTM from a UX perspective

app/scripts/pages/files.js
```diff
@@ -33,36 +31,53 @@ default class Files extends React.Component {
     clearInterval(this.pollInterval)
   }
 
-  getFiles = () => {
+  getFiles () {
```
diasdavid: isn't there a missing `=` and `=>` here, even if just for consistency

diasdavid:
Other than the one comment, LGTM :) nice!

app/scripts/pages/files.js
Dignifiedquire: No, the important part is that these are two different sorts of methods, if you declare them like this:

```js
getFiles () {
}
```

it's a regular method. If you declare it like this:

```js
```

PR #218, opened 2016-01-20
feat(connections): Move the globe outside

### LINKED ISSUE #213: Move the globe to it's own page and repo
As the globe is nice, but impacts battery life on laptops very badly on laptops I suggest extracting it into it's own repo and publishing it on it's own. This way it's still usable but not everybody is forced to use it when they open the webui.

@jbenet we talked about this on the apps hangout so please let us know if you are okay with this.

dignifiedquire:
As discussed in #213 this removes the globe from the
connections window and replaces it with a link.
In addition the peers list is transformed into a table.

Closes #213

* Currently missing is the globe which is being developed here: https://github.com/ipfs/connections-globe
* There are no locations listed as ipfs-geoip is simply to slow and unstable to use for this many peers.

### Preview
screen shot 2016-01-20 at 22 56 23


jbenet:
@Dignifiedquire 
- oh i thought we were going to keep a visual too -- what about the 2d map you had in another branch? i think we should have that. the list on its own is not very useful. (and yes link to the globe outside)
- i like the peers table, much nicer. the ID should be the full thing, even if it is hidden by the other column. should be selectable and copy-pastable (the whole thing) (also use the markdown ticks / html code? `like this`)
- why is geoip so slow? it can fetch in the background, no? it can start without locations but over time add them? (would be nice to have flags like bittorrent clients)

Dignifiedquire:
@jbenet tldr one thing at a time

> oh i thought we were going to keep a visual too -- what about the 2d map you had in another branch? i think we should have that. the list on its own is not very useful. (and yes link to the globe outside)

This is not implemented yet, and yes it will come in the future but for now we just want peoples battery life to be a bit happier.


> i like the peers table, much nicer. the ID should be the full thing, even if it is hidden by the other column. should be selectable and copy-pastable (the whole thing) (also use the markdown ticks / html code?

Okay will do

> why is geoip so slow? it can fetch in the background, no? it can start without locations but over time add them?

It's not only slow but also unreliable and not available on the 0.4 network. I agree we want flags and nice locations but I would like to drop it from this list for now and figure out how to make location detection better and more reliable and when that's done adding it back.

jbenet:
> It's not only slow but also unreliable and not available on the 0.4 network. I agree we want flags and nice locations 

We should add it to the 0.4 network, it _has_ to be added there.

> but I would like to drop it from this list for now and figure out how to make location detection better and more reliable and when that's done adding it back. 

historically, our community tends to do this and then take _months_ adding things back. using geoip is very useful in demos. Showing where things are is a big deal. And in my experience always _reliable_ -- it is indeed slow the first times, but once you have the core of the tree cached it's fine. Do not remove this feature. (i think i use the webui locations tab for demos more than any other user in the network right now)

If you want to not load it, then make it a toggle checkbox that people can turn off. it will be redone shortly to add the 2d map anyway.

jbenet:
Also, the "people's batteries" thing is really a fault of the content not being removed + garbage collected properly. it should be just fine for most mobile phones to load this tab and move out in 10s and not have any noticeable effect on battery

dignifiedquire:
Will be resolved in #229

PR #221, opened 2016-01-24
Files browser for 0.4

dignifiedquire:
> **WARNING** Hacks and bad code in here!

### Currently working

* Create directories
* Navigate directories
* Preview images, markdown and raw data

### Goals

- [ ] Clean Code
- [ ] Video previewer
- [ ] Code highlighting on text files
- [ ] Upload files
- [ ] Move folders and files via drag and drop
- [ ] Delete files and folders
- [ ] Folder creation inline
- [ ] Better sorting (first folders, then files, set sort order by clicking on the header)
- [ ] Make it look good
- [ ] Select on click, double click to open
- [ ] Inline rename 
- [ ] Copy?

### Preview

ipfs-files-2
ipfs-files-first


jbenet:
this looks amazing and i love it.

fazo96:
Wow this looks great!!!

Some other things I think it should have:

- publish button to publish object (or the root) via IPNS (maybe with an option to folder wrap or not)
- permalink button (it copies to clipboard or shows an IPFS link to an object)

As a way to copy files:

- duplicate button (creates a copy in the same folder)
- move (to move a file or folder) and paste (to paste it somewhere else)

Of course real file manager functionality (select multiple files, move/copy them) would be better but it would get a lot harder to make.

RichardLitt:
Agreed that this looks awesome!

kpcyrd:
@dignifiedquire could you test a markdown document that contains `<script>alert(document.domain)</script>` as inline html?

Kubuxu:
Are you planning to use highlight.js to highlight source files and maybe detect source files?

dignifiedquire:
@Kubuxu yes sth like that

dignifiedquire:
Closing in favor of #229

PR #229, opened 2016-01-28
[WIP] Implement Redux

dignifiedquire:
## The big todo - Or what needs to happen and where you can help

In general the webui (after this PR) will have six pages under which I will structure the content
### Home
- [x] Containers
  - [x] Implement
  - [ ] Test
- [x] Components
  - [x] Implement
  - [x] Test
- [x] API Calls
  - [x] Implement
  - [x] Test
- [x] Reducers
  - [x] Implement
  - [x] Test
- [x] Sagas
  - [x] Implement
  - [x] Test
#### WorldMap
- [x] Containers
  - [x] Implement
  - [ ] Test
- [x] Components
  - [x] Implement
  - [ ] Test
- [x] API Calls
  - [x] Implement
  - [x] Test
- [x] Reducers
  - [x] Implement
  - [x] Test
- [x] Sagas
  - [x] Implement
  - [x] Test
#### PeersTable
- [x] Containers
  - [x] Implement
  - [ ] Test
- [x] Components
  - [x] Implement
  - [ ] Test
- [x] API Calls
  - [x] Implement
  - [x] Test
- [x] Reducers
  - [x] Implement
  - [x] Test
- [x] Sagas
  - [x] Implement
  - [x] Test
### Files

This should be based on the 0.4 files api, dropbox style like in #221 started
- [ ] Containers
  - [ ] Implement
  - [ ] Test
- [ ] Components
  - [ ] Implement
  - [ ] Test
- [ ] API Calls
  - [ ] Implement
  - [ ] Test
- [ ] Reducers
  - [ ] Implement
  - [ ] Test
- [ ] Sagas
  - [ ] Implement
  - [ ] Test
### DAG Browser
- [ ] Containers
  - [ ] Implement
  - [ ] Test
- [ ] Components
  - [ ] Implement
  - [ ] Test
- [ ] API Calls
  - [ ] Implement
  - [ ] Test
- [ ] Reducers
  - [ ] Implement
  - [ ] Test
- [ ] Sagas
  - [ ] Implement
  - [ ] Test
### Config

This should have an option to edit the raw config, as well as generate a form using [redux-form](https://github.com/erikras/redux-form) 
- [x] Containers
  - [x] Implement
  - [x] Test
- [x] Components
  - [x] Implement
  - [x] Test
- [x] API Calls
  - [x] Implement
  - [x] Test
- [x] Reducers
  - [x] Implement
  - [x] Test
- [x] Sagas
  - [x] Implement
  - [x] Test
### Logs
[...]

PR #247, opened 2016-02-17
add table spec

suisha:
https://github.com/ipfs/webui/issues/137

Adds a test for the table view.

### DIFF
--- a/app/scripts/views/table.js
+++ b/app/scripts/views/table.js
@@ -14,8 +14,8 @@ default class TableView extends Component {
     return (
       <Table responsive>
         <tbody>
-          {this.props.table.map(val => {
-            return <tr><td>{addr(val)}</td></tr>
+          {this.props.table.map((val, i) => {
+            return <tr key={i}><td key={i}>{addr(val)}</td></tr>
           })}
           {this.props.children}
         </tbody>
--- a/app/scripts/views/table.js
+++ b/app/scripts/views/table.js
@@ -15,7 +15,7 @@ default class TableView extends Component {
       <Table responsive>
         <tbody>
           {this.props.table.map((val, i) => {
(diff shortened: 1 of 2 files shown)


app/scripts/views/table.js
```diff
@@ -14,8 +14,8 @@ default class TableView extends Component {
     return (
       <Table responsive>
         <tbody>
-          {this.props.table.map(val => {
-            return <tr><td>{addr(val)}</td></tr>
+          {this.props.table.map((val, i) => {
+            return <tr key={i}><td key={i}>{addr(val)}</td></tr>
```
suisha: I added key attributes here because react needs them in order to find the appropriate DOM nodes inside an array without re-rendering the entire view.

dignifiedquire: Won't react complain about duplicate keys on the `td` key?

suisha: After looking more in-depth into how react handles the :key:  property, I've determined that the `key` on the `td` element actually does nothing in this case, so I've removed it. If `td` was in another (nested) array, it would need a key, but it's not.

Regarding duplicate keys, it seems that keys only [need to be unique amongst it's siblings](https://facebook.github.io/react/docs/reconciliation.html#keys). I believe this is because react doesn't use the key we provide as-is, but rather uses it to generate a unique `data-reactid`. In a case where a parent and a child have the same key, they would still have different `reactid`'s. If we have a key value of `1`, a parent would have something like `.0.1:$1` and child would have `.0.1:$1.1`, so there wouldn't actually be any conflict.

dignifiedquire: Cool thanks for the research and the info!

dignifiedquire:
Thanks :octocat: 

Just as a note, if you plan doing further work, please take a look at the `redux` branch (#229) As that changes a lot of things around here and is where I'm actively working on.

PR #248, opened 2016-02-17
specify node/npm versions

suisha:
https://github.com/ipfs/webui/issues/245#issuecomment-184569472

Specify which versions of node and npm to use for contributors.


dignifiedquire:
Thanks :octocat:

PR #253, opened 2016-02-22
world spec

suisha:
https://github.com/ipfs/webui/issues/137

I'm not sure if [`_renderMap`](https://github.com/ipfs/webui/blob/redux/app/scripts/components/world.js#L21-L91) is testable here without either exposing it or doing a non-shallow render(not advisable for unit tests).

We could also make this a separate react component with width/height as state, if it's reusable.

### DIFF
--- a/test/components/map.spec.js
+++ b/test/components/map.spec.js
@@ -1,0 +1,15 @@
+import {expect} from 'chai'
+import {render} from 'enzyme'
+import sinon from 'sinon'
+import React from 'react'
+
+import WorldMap from '../../app/scripts/components/world-map'
+
+describe('WorldMap', () => {
+  it('should render an svg', () => {
+    const c = [[12, 14]]
+
+    const el = render(<WorldMap coordinates={c}/>)
+    expect(el.find('svg').length).to.equal(1)
+  })
+})
--- a/test/components/world-map.spec.js
+++ b/test/components/world-map.spec.js
@@ -1,14 +1,24 @@
 import {expect} from 'chai'
-import {render} from 'enzyme'
+import {shallow, render} from 'enzyme'
 import React from 'react'
 
 import WorldMap from '../../app/scripts/components/world-map'
 
 describe('WorldMap', () => {
-  it('should render an svg', () => {
-    const c = [[12, 14]]
+  const c = [[12, 14]]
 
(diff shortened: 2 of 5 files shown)


dignifiedquire:
Yeah `_renderMap` is a bit tricky, it does sound like a good idea to pull it into it's own component for ease of testing. Wanna do this in here?

test/components/map.spec.js
```diff
@@ -0,0 +1,15 @@
...
+import sinon from 'sinon'
+import React from 'react'
+
+import WorldMap from '../../app/scripts/components/world-map'
+
+describe('WorldMap', () => {
+  it('should render an svg', () => {
+    const c = [[12, 14]]
+
+    const el = render(<WorldMap coordinates={c}/>)
```
suisha: I decided to do a static render here since this component only has one child.

test/components/world-map.spec.js
```diff
@@ -0,0 +1,14 @@
...
+import React from 'react'
+
+import WorldMap from '../../app/scripts/components/world-map'
+
+describe('WorldMap', () => {
+  it('should render an svg', () => {
+    const c = [[12, 14]]
+
+    const el = render(<WorldMap coordinates={c}/>)
+    expect(el.find('svg').length).to.equal(1)
```
suisha: I tried to think of more meaningful assertions to add here but drew a blank. The resulting output of `_renderMap` is basically all generated by `d3` and I couldn't find anything else on the rendered output that could be used to assert correctness.

dignifiedquire: You could assert that the point is drawn that you passed in

suisha:
:100:

test/components/world-map.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+describe('WorldMap', () => {
+  const c = [[12, 14]]
+
+  it('should render an svg', () => {
+    const el = render(<WorldMap coordinates={c}/>)
+    expect(el.find('svg').length).to.equal(1)
+  })
+
+  it('should have the correct coordiates', () => {
```
dignifiedquire: typo coordi**n**ates

dignifiedquire:
Thanks, one small note other than it looks good. Could you squash your commits together please?

PR #272, opened 2016-03-05
config page redux

suisha:
- [x] Implement config page with redux
- [x] Address feedback
- [x] Tests
- [x] Squash/rebase


suisha:
If this looks ok from a design/architectural perspective I will add tests/squash and wrap this up.

app/scripts/containers/config.js
```diff
@@ -0,0 +1,102 @@
...
+    height: React.PropTypes.bool.isRequired
+  };
+
+  componentWillMount () {
+    this.props.loadConfigPage()
+  }
+
+  componentWillUnmount () {
+    this.props.resetErrorMessage()
+    this.props.savedConfig(false)
```
dignifiedquire: These could be dispatched via a saga on a `leaveConfigPage` action that would make it more consistent with the rest

app/scripts/containers/config.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function mapStateToProps (state) {
+  return {
+    ...state.config,
+    errorMessage: state.errorMessage
+  }
+}
+
+const mapDispatchToProps = {
```
dignifiedquire: Please put this inline in the `connect` call to be consistent with the other containers

app/scripts/sagas/index.js
```diff
@@ -128,6 +132,38 @@ export function * fetchPeerIds () {
   }
 }
 
+export function * loadConfig (action) {
+  try {
+    const ipfsConfig = yield call(api.getConfig)
+    const configJSON = JSON.stringify(ipfsConfig, undefined, 2)
```
dignifiedquire: Stringify is really part of the display logic, that should happen only in the component and not in the state I'd say

app/scripts/services/api.js
```diff
@@ -168,3 +168,27 @@ export const files = {
     return api.files.mkdir(name)
   }
 }
+
+export const getConfig = (api = localApi) => {
+  return new Promise((resolve, reject) => {
+    api.config.show((err, config) => {
```
dignifiedquire: The api already supports promises, you can write this as

```js
return api.config.show()
.then((res) => JSON.parse(res.toString()))
```

app/scripts/services/api.js
```diff
@@ -168,3 +168,27 @@ export const files = {
...
+      } else {
+        resolve(JSON.parse(config.toString()))
+      }
+    })
+  })
+}
+
+export const saveConfig = (config, api = localApi) => {
+  return new Promise((resolve, reject) => {
+    api.config.replace(new Buffer(config), (err) => {
```
dignifiedquire: Same as above

app/scripts/actions/config.js
```diff
@@ -0,0 +1,28 @@
+import {action} from './utils'
+
+export const CONFIG = {
+  LOAD_CONFIG_PAGE: 'LOAD_CONFIG_PAGE',
```
dignifiedquire: for the load part please use [`createPage`](https://github.com/ipfs/webui/blob/redux/app/scripts/actions/utils.js#L33)

dignifiedquire:
Thanks, nice work so far, I left some minor comments :)

app/scripts/sagas/index.js
suisha: I currently have two properties that store config state: `config` and `draft`. `config` is the original and draft contains user changes. If the user clicks reset I set `draft` = `config`.

Would it be acceptable here to store `draft` as a string (since it's unvalidated user input), and `config` as object? I'm not sure how else I could handle this, so I'm open to suggestions.

dignifiedquire: Yes I think it's best to store the draft as unmodified string (as the user entered it) and the the actual config as object.

suisha:
This is ready to rebase/squash and ship. Any other feedback before I force push this branch?

dignifiedquire:
Great work! go ahead and squash everything :)

suisha:
:shipit:

dignifiedquire:
@suisha seems there are merge conflicts after your squash, could you check please?

suisha:
squashed again

dignifiedquire:
Let's do this!

PR #301, opened 2016-03-17
Added support for dutch

brainframe-me:
At least like this we dutch people see something when we connect with our browsers, otherwise it is very confusing because the browser does not do anything! Should maybe be build into the code to resort to default EN locale if the current one is not supported


dignifiedquire:
Thanks, actually it should be configured to show the English version as fallback when nothing else is configured. Need to investigate this

RichardLitt:
Dank je, @brainframe-me!

PR #460, opened 2016-08-14
Fix TypeError on call to input value

louisbourque:
Fixes a `TypeError: e.refs.dagPath.val is not a function` when clicking the GO button with a DagPath in the input box at the top of the webui.


dignifiedquire:
Thank you

PR #575, opened 2016-11-16
Fixes #512, needs UI changes to feel better

### LINKED ISSUE #512: The "delete" button in the web-interface for local data has no effect
In the web-interface under "files" there you can see a list of the local files. If I click under "action" onto the "delete" button (x), it has no effect, say the file stais.

Thanks for fixing this!

Version of ipfs: 0.4.2 and 0.4.3

FrankPetrilli:
Fixes #512, makes delete button work, but UI feel is off; there's no responsiveness that the file was actually unpinned. I've added an alert() with a "please reload" message, but it's insufficient to call a solution.

### DIFF
--- a/app/scripts/views/filelist.js
+++ b/app/scripts/views/filelist.js
@@ -12,13 +12,16 @@ default class FileList extends Component {
     gateway: React.PropTypes.string
   };
 
-  _unpin (event, hash) {
+  _unpin (hash, event) {
     event.preventDefault()
     event.stopPropagation()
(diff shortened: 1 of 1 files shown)


VictorBjelkholm:
I think a better solution would be to pass down a callback from `pages/files.js` (where FileList is being used) that can update the UI by fetching the files again, instead of reloading the entire page.

diasdavid:
This has been a broken feature since forever. Going to merge this PR to get the functionality working. Will add @VictorBjelkholm as an issue.

PR #576, opened 2017-01-19
Translation fix

Infinisil:
German plural of "Peer" is "Peers"


lgierth:
And isn't it "verbunden mit"?

Infinisil:
Yes, there were also some other weird translations, I fixed all of them (but one) I think. I'm not sure what to use for 

    "Go to console home": "Gehe zur Konsole Home",

because this sounds really weird. Would love some other native german speaker to review the file.

dignifiedquire:
Thank you

PR #581, opened 2017-03-16
File manager

alanshaw:
This is the file manager from #229 along with the redux additions and toastr component for feedback.

#229 is tricky to review because it is so broad in scope. The idea is to split it into multiple PRs so that they're easier to review and merge.

N.B. I appreciate this PR is also very large but it is significantly smaller than #229!

The plan after this is merged is to refactor the code slightly to allow the various pages of the app to be more standalone/modular, so that, for example, the file manager can be used in a different project.

Once the groundwork is in place for pages of the app to be standalone, we can PR for each page, extracting the work done in #229 and combining it with the new structure.

This is 99.9% the good work of @dignifiedquire but please feel free to ask me questions about the code as I'm fairly familiar with it now.

### Screenshots

screen shot 2017-03-15 at 15 49 00

#### Selecting files

screen shot 2017-03-15 at 15 49 21

#### Create folder

screen shot 2017-03-15 at 15 50 20

#### Delete item

screen shot 2017-03-15 at 15 51 00

#### Drag & drop upload

screen shot 2017-03-15 at 15 52 32

#### Context menu

screen shot 2017-03-15 at 15 51 38

screen shot 2017-03-16 at 13 37 18

#### File preview

screen shot 2017-03-15 at 15 48 53

screen shot 2017-03-15 at 15 48 36

### Next steps

Firstly, there's not enough tests for this piece of work so I'll be submitting another PR with tests shortly.

Secondly, there's a few things missing from the file manager that will be submitted as separate PRs:

* Move/copy files
* File/folder rename
* Sorting/ordering files within a folder


app/scripts/components/ipfs.js
```diff
@@ -0,0 +1,43 @@
...
+  return React.createClass({
+    displayName: `withIpfs(${Component.displayName || Component.name || 'Component'})`,
+
+    getInitialState () {
+      return {version, gateway}
+    },
+
+    componentDidMount () {
+      ipfs.version((err, res) => {
+        if (err) return console.error(err)
```
dignifiedquire: would be good to propagte these errors to the toastr state, rather than just logging them

app/scripts/components/ipfs.js
```diff
@@ -0,0 +1,43 @@
...
+        version = res.Version
+        this.setState({version: res.Version})
+      })
+      ipfs.config.get('Addresses.Gateway', (err, res) => {
+        if (err) {
+          return console.error(err)
+        }
+        if (res == null) {
+          return console.error(new Error('No Gateway found'))
```
dignifiedquire: same as above

app/scripts/components/ipfs.js
alanshaw: screen shot 2017-03-16 at 16 22 21

dignifiedquire: Nice :)

app/scripts/actions/router.js
```diff
@@ -0,0 +1,3 @@
+import {push, replace, go, goBack, goForward} from 'react-router-redux'
+
+export {push, replace, go, goBack, goForward}
```
dignifiedquire: This feels superfluous, lets replace the imports of these with direct imports to `react-router-redux`

app/scripts/containers/root.js
```diff
@@ -0,0 +1,27 @@
+import React, {Component, PropTypes} from 'react'
+import {Provider} from 'react-redux'
+import {Router} from 'react-router'
+
+window.React = React
```
dignifiedquire: This should probably be guarded and only be done non production envs as I think this is only used to attach the react devtools

dignifiedquire: This should probably be guarded and only be done non production envs as I think this is only used to attach the react devtools

d10r:
Please also provide a way of getting the hash of files. 
For me a frequent use case is to add a file to which I want to link.

I just made a modification to the old code which does this for newly added files, not realizing that a big update is in process: https://github.com/d10r/webui/commit/a95d430164f6f78ad8f0630befcf82e9f0c32668

PR #605, opened 2017-11-12
[WIP] Update Dependencies and Small Improvements

### LINKED ISSUE #603: Currently doesn't work for production build
Trying to use the newest code from webui (ea3664bde32490c3f082696726e4f334adc27901) together with go-ipfs. 

After installing the dependencies, I run `yarn build`, then add the hash to IPFS (`ipfs add -r dist/`) and replace the following line in go-ipfs with the root hash of the add: https://github.com/ipfs/go-ipfs/blob/0efdcadafb526ce0edce0a6ae4ca994740c7f844/core/corehttp/webui.go#L4 (then rebuild and start the newly built daemon)

After doing this, it quite doesn't work... First issue I hit was that there is a typo in the production configuration. See https://github.com/ipfs/webui/blob/ea3664bde32490c3f082696726e4f334adc27901/src/app/js/store/configure-store.prod.js#L3 Supposed to be `Middleware` instead of `Middlware`, notice the missing `e`.

After fixing that, things actually load but the CSS is not part of the HTML file that is loaded. See QmefLwZaGVCwRkdDMBNExWs5ahFERnyxDV7Ngnx9sgiefU

This is how the `index.html` ends up looking:

```
<!doctype html><head><meta charset="utf-8"/><meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no"/></head><body><div id="root"></div><script src="ipfs-webui.0.3.0.js"></script></body>
```

Started to dig into how `hjs-webpack` works and actually injects styles but gave up at this point.

hacdias:
### Update Dependencies and Other Improvements

- [x] Update Dependencies on package.json
- [x] Fix File's preview page
- [x] Replace safe-buffer by native Buffer
- [x] Fix Object's page
     - [x] View object by Hash (Path changed from `/objects/\\ipfs\\{hash}` to `/objects/ipfs/{hash}`.
     - [x] ~~View object by Path~~
- [x] Fix #603 
     - [x] Same-name directory issue
     - [x] Firefox Upload (**can't reproduce**)
- [x] Fix build
- [x] Review tests to work with new deps

### File Manager/Preview Improvements (secondary)

- [x] File Explorer and File Preview navigation is the same
- [x] File Preview no longer renders errors
- [x] File Explorer and File Preview now have different URLs for different files
- [x] Upload Button on File Manager
- [ ] Prettier Action Bar (?)
- [ ] Prettier File Listing (?)
- [ ] Preview PDFs too (?)
- [ ] Rename
     - [x] Context Menu
     - [x] Action Bar
     - [ ] Pretty Dialog Box
- [ ] "Move to" dialog
- [ ] "Copy to" dialog
- [x] Shortcuts
    - [x] F2 - rename
    - [x] ESC - clear selection
    - [x] CTRL + Click - select multiple files or directories
    - [x] SHIFT + Click - select a range of files or directories
    - [x] DEL - delete current selection


diasdavid:
@hacdias Is this working for you locally? Maybe I started reviewing way too early :) Sorry about that!

hacdias:
@diasdavid everything but file previews and objects' page.

hacdias:
@diasdavid:

1. Is there any specific reason to use HashHistory (e.g.: `address/#path`) instead of BrowserHistory (e.g.: `address/path`)?
2. What's the difference between 'containers' and "legacy" pages?

diasdavid:
Ah, the error was on my end, I was using another branch of go-ipfs. Now I see it working:

image

Great work with the deps update, @hacdias! 👏🏽

> re: Is there any specific reason to use HashHistory (e.g.: address/#path) instead of BrowserHistory (e.g.: address/path)?

When you use a `#` you are certain that information never goes to the server, it stays on the front end app.

> What's the difference between 'containers' and "legacy" pages?

Containers come from [Redux](https://redux.js.org/), are you familiar with it?

hacdias:
I'm not, but I'll get familiar to it ;) I'm still updating object's page.

Right now, on master, previews are not working but they're working now. Although, I'm receiving a very strange error which doesn't have any side effect apparently.

image

Another thing I noticed are the 'Unmapped range' errors I'm getting on Connections page. I already know their source but I'm not understanding why they are happening here.

image

Do you have any thoughts on this?

diasdavid:
I'm also checking what's happening there, it seems that here:

https://github.com/ipfs/webui/blob/master/src/app/js/pages/connections.js#L56

We call getLocation (which is lookupPretty) with a PeerId and a multiaddr, but then it returns an error "Unmapped range" coming from https://github.com/ipfs/ipfs-geoip/blob/master/src/pretty.js#L15-L17 which only happens when we pass an empty array.

Now checking what goes inside the Peer :)

diasdavid:
It's actually coming from https://github.com/ipfs/ipfs-geoip/blob/master/src/lookup.js#L50. Bad error names, let me change that. Nevertheless, this means that geoip doesn't have info from the location of that IP, this should not error to the user and only show that location is Unknown

hacdias:
Good to know you found the problem so fast 😮

diasdavid:
I've been around this codebase before 😉

geoip will need:
- [ ] IPv6 support https://github.com/ipfs/ipfs-geoip/issues/60
- [ ] Be tested with multiple multiaddrs now that we have things like Circuit Relay https://github.com/ipfs/ipfs-geoip/issues/61

Until then, webui should not fail on the user, it should just show n/a

diasdavid:
@hacdias mind looking into https://github.com/ipfs/webui/issues/603 as well?

hacdias:
@diasdavid sure thing.

hacdias:
@diasdavid sorry, there are some things I'm needing help with:

1. [This code](https://github.com/ipfs/webui/blob/update-dependencies/src/app/js/sagas/pages/preview.js#L13) is generating a `Unexpected token A in JSON at position 0` which I think it's because 'name' will be 'undefined' at first. [See more](https://github.com/reactjs/react-router-redux/issues/483).
2. Later, I will probably need some help with the tests. I'm not used to write tests and I haven't already understood how they work (I'll try to dig more into them though).
3. I'm not fully understanding why the objects page is not working as expected. Although, the current version on master also produces some errors but works better.

In the meanwhile, I'll make some improvements to the File Manager and the preview page and update this PR accordingly.

hacdias:
@diasdavid take a look at the checkboxes above 😄 There are a lot of new ones!! I'd like to know if you could help me with the unchecked checkboxes from the first section ("Update Dependencies and Other Improvements").

About the File Manager: do you think a download button for files/folders would be great? What about a slightly improved design? One more question: we're using [Bootstrap](https://getbootstrap.com/), right? 😃

diasdavid:
Still some deps to be updated

image

Here is a handy module to check things http://npmjs.com/npm-check-updates

hacdias:
Yeah, I've been using that module. Thanks 😄

diasdavid:
Hi @hacdias! Just got to review your work, lots of progress here! Congratz 👏🏽👏🏽👏🏽

This PR is a bit exploding in size, let's focus on wrapping it up so that then multiple things can be tackled at the same time.

From what I understand, the things that are essential to merge this PR are:

- [ ] Update tests and make sure CI is green
- [ ] Update remaining dependencies
- [ ] Fiz search by Object Path

> About the File Manager: do you think a download button for files/folders would be great? What about a slightly improved design? One more question: we're using Bootstrap, right? 😃

> Edit: what do you think about using Semantic UI instead of Bootstrap? They've got an official package for React and it doesn't seem that our current library (react-bootstrap) isn't getting much love recently!

We picked Bootstrap at the time because of its community support, documentation and stability. I understand that the world has changed and that new UI frameworks can fit us better.

Before we jump to a decision, let's create a issue (you create it :)) in this repo to discuss what is the best and then we ask our community to help us decide given their experience.

Btw, since you are also a designed, how experienced are you with Invision and Sketch? Wanna work on the next UI together? I can create a team and invite you there.

hacdias:
Thanks!

> This PR is a bit exploding in size, let's focus on wrapping it up so that then multiple things can be tackled at the same time.

Sure, let's focus on its initial topic and update the dependencies. I'll need some help with the tests and the object by path. The remaining dependencies are updated 😄 

> We picked Bootstrap at the time because of its community support, documentation and stability. I understand that the world has changed and that new UI frameworks can fit us better.
> 
> Before we jump to a decision, let's create a issue (you create it :)) in this repo to discuss what is the best and then we ask our community to help us decide given their experience.
> 
> Btw, since you are also a designed, how experienced are you with Invision and Sketch? Wanna work on the next UI together? I can create a team and invite you there.

I'll open an issue then. I'm not experienced with Invision nor Sketch. Yeah, I'd like to work on the next UI 😮 It seems to be a challenging work to do and I could learn more about design I guess. 😄

hacdias:
@diasdavid I think there is some kind of error on js-ipfs-api in `config.replace` which is returning `File argument 'file' is required` when we try to save the file. Could you check that, please?

diasdavid:
@hacdias I believe that was a bug introduced with https://github.com/ipfs/js-ipfs-api/pull/629 which changed how files are sent to the daemon, wanna try figure that one out and added a test to make sure it works well? @pgte can help you by answering your questions :)

hacdias:
I'll check that this evening. In the meanwhile, everything is greeen 😄

diasdavid:
# 👏🏽👏🏽👏🏽👏🏽👏🏽👏🏽 Excellent work @hacdias 



This means that we finally can ship new WebUIs to go-ipfs //cc @whyrusleeping

package.json
```diff
@@ -14,95 +14,95 @@
...
     "stream-http": "^2.7.2",
-    "style-loader": "^0.18.2",
-    "url-loader": "^0.5.9",
-    "webpack": "^3.0.0"
+    "style-loader": "^0.19.0",
+    "url-loader": "^0.6.2",
+    "webpack": "^3.8.1"
   },
   "pre-commit": [
     "lint"
```
diasdavid: Sooo greeeen! :D

src/app/js/containers/files-explorer.js
```diff
@@ -1,32 +1,127 @@
...
+      if (la > i) {
+        fi = i + 1
+      } else {
+        fi = la
+        la = i - 1
+      }
+      break
+    }
+
+    this.selectAllBetween(fi, la)
```
diasdavid: @hacdias can we convert this loop into something more JavaScript'y/Functional? Check out all the beautiful array methods https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array and try to use variable names that try to be more explicit (it is ok to use full words in JS land)

src/app/js/containers/files-explorer.js
```diff
@@ -1,32 +1,127 @@
...
+    const {
+      root,
+      list,
+      select
+    } = this.props
+
+    for (let i = fi; i <= la; i++) {
+      const filePath = join(root, list[i].Name)
+      select(filePath)
+    }
```
diasdavid: Yeah, let's make this more JavaScripty, I have no idea what is `fi` and `la`

package.json
hacdias: Am I supposed to change anything in here? 😮

diasdavid: just happy :)

PR #608, opened 2017-12-01
Adds copy hash to clipboard context menu item

alanshaw:
screen shot 2017-12-01 at 11 07 26

### DIFF
--- a/src/app/js/components/files/context-menu.js
+++ b/src/app/js/components/files/context-menu.js
@@ -15,7 +15,7 @@ class FilesContextMenu extends Component {
         </MenuItem>
         {selectedFiles.length === 1 && onCopyHash ? (
           <MenuItem onClick={onCopyHash}>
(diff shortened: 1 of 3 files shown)


src/app/js/components/files/context-menu.js
```diff
@@ -13,13 +13,19 @@ class FilesContextMenu extends Component {
         <MenuItem onClick={onMoveDir}>
           Rename
         </MenuItem>
+        {selectedFiles.length === 1 && onCopyHash ? (
+          <MenuItem onClick={onCopyHash}>
+            Copy hash to clipboard
```
hacdias: I would change to 'Copy Hash' only. I think it's better since the 'clipboard' part is implied. 😄

hacdias:
I think we can merge this @diasdavid! Awesome @alanshaw!

PR #620, opened 2018-01-19
fix: an error in webui-zh.json

alienmao:
There should not be an extra double quotation mark in line 3


hacdias:
Thank you @alienmao

Issue #625, opened 2018-01-31
Request over https not working


Sh4d1:
Hello!

Maybe it is a js-ipfs issue, but I'll post here first.

I have these erros on the JS console:
```
Mixed Content: The page at 'https://ipfs.ptrk.io/ipfs/QmQLXHs7K98JNQdWrBB2cQLJahPhmupbDjRuH1b9ibmwVa/#/home' was loaded over HTTPS, but requested an insecure resource 'http://ipfs.ptrk.io:443/api/v0/id?stream-channels=true'. This request has been blocked; the content must be served over HTTPS.
d._onFinish @ ipfs-webui.0.3.0.js:1
```

```
Mixed Content: The page at 'https://ipfs.ptrk.io/ipfs/QmQLXHs7K98JNQdWrBB2cQLJahPhmupbDjRuH1b9ibmwVa/#/home' was loaded over HTTPS, but requested an insecure resource 'http://ipfs.ptrk.io:443/api/v0/version?stream-channels=true'. This request has been blocked; the content must be served over HTTPS.
d._onFinish @ ipfs-webui.0.3.0.js:1
```

```
Mixed Content: The page at 'https://ipfs.ptrk.io/ipfs/QmQLXHs7K98JNQdWrBB2cQLJahPhmupbDjRuH1b9ibmwVa/#/home' was loaded over HTTPS, but requested an insecure resource 'http://ipfs.ptrk.io:443/api/v0/config?arg=Addresses.Gateway&stream-channels=true'. This request has been blocked; the content must be served over HTTPS.
d._onFinish @ ipfs-webui.0.3.0.js:1
```

When I try to load `https://ipfs.ptrk.io/api/v0/version?stream-channels=true` it is working, but the request does not seem to use HTTPS.

Any ideas where I can change that?

Cheers!

PR #626, opened 2018-01-31
Add HTTPS when calling ipfs API

Sh4d1:
Hey,

I saw that when NODE_ENV is set to production, even if the site is on https, the call is made over http. Thsi PR fixes it, taking the protocol of the site to send the requests.

See #625

### DIFF
--- a/src/app/js/components/ipfs.js
+++ b/src/app/js/components/ipfs.js
@@ -4,7 +4,9 @@ import createReactClass from 'create-react-class'
 
 const host = (process.env.NODE_ENV !== 'production') ? 'localhost' : window.location.hostname
 const port = (process.env.NODE_ENV !== 'production') ? '5001' : (window.location.port || (window.location.protocol === 'https:' ? 443 : 80))
-const ipfs = require('ipfs-api')(host, port)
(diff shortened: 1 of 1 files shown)


diasdavid:
@hacdias did you test this with localhost?

hacdias:
@diasdavid yeah, it defaults to HTTP. Although, there could be a problem if someone used HTTPS on localhost. But that's a matter of removing the first condition.

Sh4d1:
@diasdavid @hacidas also, I'm not sure it is a web UI issue though but it is not working.
I have a docker image of go-ipfs running on my server, and traefik as a reverse proxy. I expose the port 5001, and it is also accessible via HTTPS on ipfs.ptrk.io. When I try to load a request like API/v0/version, it loads the content but the request is not finished (still loading), same when I curl it. When I use the port 5001, it's all good. So on the WebUI, the requests made to https://ipfs.ptrk.io, are taking a long time, and when they are finished, nothing appears on the UI.

Sh4d1:
I'm pretty sure it is linked to [this](https://github.com/ipfs/go-ipfs/issues/4639)

PR #649, opened 2018-03-25
chore: update webpack & upgrade pages

hacdias:
**[WIP]**

Updating Webpack to version 4 and upgrading pages to use Redux reusing code from #229.


olizilla:
@hacdias is there any ideas in here we should save? If not can you close this out pls.

hacdias:
@olizilla just closing this. Let's focus on the revamp 😄

PR #654, opened 2018-04-15
Add Portuguese i18n support

mauriciomelo:
🌐 🇧🇷 🇵🇹🇦🇴 🇨🇻 🇹🇱 🇬🇶 🇬🇼 🇲🇿 🇸🇹 🌐


hacdias:
Hello @mauriciomelo. Thanks!! I appreciate your effort, but could I just ask you to change it to 'pt-br'? We still don't have an European variant, but it will be made soon 😄

mauriciomelo:
Sure, I'll update it ASAP.

olizilla:
What needs to happen to make this mergeable @hacdias ? is it just a file rename?

hacdias:
@olizilla I would just prefer to have it explicitly saying this was Brazilian variant but since we don't have an European translation, let's stick with this. Merging!!! ⛵️

PR #658, opened 2018-05-17
feat: files components

hacdias:
localhost_9009__selectedkind breadcrumbs selectedstory colors full 0 addons 1 stories 1 panelright 0 addonpanel storybook 2factions 2factions-panel

I was creating the breadcrumbs component for the files page but I'm a bit reluctant about how to do it and if this is going to be easily reused by another parts of the application.

What do you think @olizilla?


hacdias:
Now a big and beautiful checkbox:

image

I had to be a bit [hacky here](https://github.com/ipfs-shipyard/ipfs-webui/pull/658/files#diff-32d85b53aba5403fc7a7351272e484f4R27) and change the viewBox of the icon because it has a lot of padding. 

Another thing to mention: this tick isn't the same we can find on Invision.

src/components/breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import './Breadcrumbs.css'
+
+function makeBread (root) {
+  if (root.endsWith('/'))
+    root = root.substring(0, root.length - 1)
```
olizilla: We're using `standard` on this project, so this triggers a lint error. Can you update the .travis.yml to ensure that `npm run lint` is run on ci. Can your editor be configured to show lint errors as you type? I find that useful.

Also we now have a mix of unix and windows line endings. I'll add an editorconfig and a gitattributes file to the revamp branch to fix it.

- http://editorconfig.org/
- https://help.github.com/articles/dealing-with-line-endings/#platform-all

src/components/breadcrumbs/Breadcrumbs.stories.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { action } from '@storybook/addon-actions'
+import { checkA11y } from '@storybook/addon-a11y'
+import Breadcrumbs from './Breadcrumbs'
+
+storiesOf('Breadcrumbs', module)
+  .addDecorator(checkA11y)
```
olizilla: A11y checking right in the story. Rad!

src/components/breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
...
+    parts[i] = {
+      name: parts[i].name,
+      path: parts[i - 1].path + '/' + parts[i].path
+    }
+  }
+
+  return parts
+}
+
+export default function Breadcrumbs ({path, onClick, className = '', ...props}) {
```
olizilla: We need to think about about the API for this component. we'll have addresses like
`/ipfs/QmHash/foo/bar`. The simplest api would let us pass a valid ipfs address to the breadcrumb and have it do the right thing. I don't see the breadcrumb as a generic component like button or checkbox is, so let's move it to the files directory, and make it a bit less generic.

We'll probably represent the users MFS root with house icon. We won't show the label "MFS root" as that is an internal implementation detail.

src/components/checkbox/Checkbox.css
```diff
@@ -0,0 +1,8 @@
+.Checkbox {
```
olizilla: Checkbox is a separate issue. Whereever possible, keep PRs focused on 1 thing, even if it's tiny. Many small PRs are easier to merge, and it avoids issue where good work can't be merged when there are issues with the other parts.

src/components/checkbox/Checkbox.js
```diff
@@ -0,0 +1,43 @@
...
+    let content = null
+    className = `Checkbox pointer br1 w1 h1 ${className}`
+
+    if (this.state.checked) {
+      content = <Tick fill='#69C4CD' viewBox='25 25 50 50' />
+      className += ' checked'
+    }
+
+    return (
+      <div
```
olizilla: In general, this is nice code. One improvement would be to implement it as a hidden checkbox input wrapped in a label, then style the label as you have this div. That way you get all the behaviour of a real input element for free.

src/components/checkbox/Checkbox.js
```diff
@@ -0,0 +1,43 @@
...
+        role='checkbox'
+        aria-checked={this.state.checked}
```
olizilla: these `aria` attributes make me very happy ✨🎷🐩

src/components/breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
...
+  const cls = `Breadcrumbs sans-serif f4 ${className}`
+  const bread = makeBread(path)
+  const last = bread.pop()
+  const res = []
+
+  bread.forEach((link, index) => {
+    res.push(<a className='pointer' key={`${index}link`} onClick={() => { onClick(link.path) }}>{link.name}</a>)
+    res.push(<span className='mh1' key={`${index}divisor`}>></span>)
```
olizilla: `divisor` is a number that you divide another number by. I think you want `divider` here.

src/components/breadcrumbs/Breadcrumbs.js
hacdias: image

@olizilla yeah, I agree. So here we would remove the IPFS and MFS Root and put a home thumbnail right?

src/components/checkbox/Checkbox.css
hacdias: Yeah, I know it's a separate issue, although I needed it for the files. I can separate it though and create a PR just for it and then rebase this branch.

src/components/breadcrumbs/Breadcrumbs.js
hacdias: @olizilla yeah, thanks @olizilla 😆

src/components/checkbox/Checkbox.css
olizilla: No need to split this one, just a note for future ones.

hacdias:
image

Just to create some juice on your mouth. 😆

hacdias:
image

I'm not sure if I should put 1 and 2 together on the same component or not. On one side, I find that they should stick together because they're tightly related: 2 are the actions we can make when selecting from 1.

On the other hand, 1 could be a FileList and 2 a FilesAction or something. That would also make it easier to set the correct fixed position.

What are your thoughts on this?

olizilla:
@hacdias they should be separate components, and then you can create a third component that compose the two pieces.

src/files/file/File.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  return (
+    <div className={className}>
+      <div className='pa2 w2'>
+        <Checkbox checked={selected} onChange={select} />
+      </div>
+      <div className='name flex items-center flex-grow-1 pa2 w-40'>
+        <div className='dib icon'>
+          {extToPic(props)}
```
olizilla: @hacdias with a little tweak this can be a JSX element. if we rename it FileIcon we can

```jsx
<FileIcon name={filename} /> 
```

PR #659, opened 2018-05-18
chore: ensure consistent line endings

olizilla:
Adds editorconfig to keep our editors ending lines with LF and .gitattributes to get git to ensure LF line endings when the editor doesn't.

License: MIT

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,12 +3,14 @@ root = true
 
 [*]
 charset = utf-8
-indent_style = space
-indent_size = 2
-trim_trailing_whitespace = true
-insert_final_newline = true
 end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
 
-# Tab indentation (no size specified)
 [Makefile]
 indent_style = tab
+
+[*.md]
+trim_trailing_whitespace = false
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
```
lidel: Won't this break Github-style markdown where you indicate a `<br>` via two spaces at the end of line?

olizilla: oh my. _i did not know that was a thing_. I think github preserves newlines in markdown.
without
spaces
at
the
end
of
the
line
;-)

lidel: Yeah, Github uses simplified end line rules in comments :upside_down_face: 
But in `README.md` you need to put blank line for a new paragraph, otherwise it will be rendered as a single line.

olizilla: oh whut. every day a school day when nothing is consistent.

I copied this config from js-ipfs, so that's an interesting find @lidel

olizilla: @lidel i've added an override for markdown files

```
[*.md]
trim_trailing_whitespace = false
```

PR #662, opened 2018-05-18
feat: checkbox

hacdias:
image


src/components/checkbox/Checkbox.css
```diff
@@ -0,0 +1,21 @@
+.Checkbox span:first-of-type {
+  background-color: #DDE6EB;
+  box-shadow: inset 0 1px 1px 0px rgba(0,0,0,0.18), 0 1px 0 0px rgba(255,255,255,0.75);
+}
+
+.Checkbox input {
```
olizilla: You can replace this with tachyons atoms

```jsx
<input className='absolute o-0'  type='checkbox' disabled={disabled} onChange={change} />
```

our css is always more specific than tachyons because they are all single classes and we're after it in the load order.

src/components/checkbox/Checkbox.css
```diff
@@ -0,0 +1,21 @@
...
+  position: absolute;
+  opacity: 0;
+}
+
+.Checkbox svg {
+  opacity: 0;
```
olizilla: Same here

```jsx
  <Tick className='w1 h1 o-0' fill='#69C4CD' viewBox='25 25 50 50' />
```

src/components/checkbox/Checkbox.js
```diff
@@ -0,0 +1,24 @@
...
+
+  const change = (event) => {
+    onChange(event.target.checked)
+  }
+
+  return (
+    <label className={className} {...props}>
+      <input type='checkbox' disabled={disabled} onChange={change} />
+      <span className='dib v-mid br1 w1 h1 mr1'>
+        <Tick className='w1 h1' fill='#69C4CD' viewBox='25 25 50 50' />
```
olizilla: We've got a `.fill-aqua` atom in [ipfs-css](https://github.com/ipfs-shipyard/ipfs-css/blob/ed7fa0fdae2c0b241ba1d2195614b843a0fea077/src/theme.css#L87)

hacdias:
Sorry @olizilla, completely missed the tachyons!

PR #663, opened 2018-05-18
chore: run tests against dev server in dev

olizilla:
Our e2e test were pointing at the prod build while our unit tests
would test the current src tree. This quickly leads to annoyance.

This PR runs the e2e style tests against the dev server, and spins
up a test server in ci. This simplifes the run-scripts too.

The only trade off is the test will fail if a server isn't running,
but as the dev flow is to run the dev server and leave the tests
running in watch mode, this isn't an issue.

License: MIT


package.json
```diff
@@ -8,10 +8,10 @@
     "build": "react-scripts build",
     "build:icons": "svgr -d src/icons node_modules/ipfs-css/icons && standard --fix src/icons/*",
     "lint": "standard src",
-    "test": "run-p --race test:serve test:run",
-    "test:coverage": "run-p --race test:serve \"test:run -- --coverage\"",
+    "test": "react-scripts test",
+    "test:coverage": "npm run test:run -- --coverage\"",
```
alanshaw: `test:run` no longer exists...

src/App.js
```diff
@@ -17,6 +17,7 @@ export class App extends Component {
   }
 
   componentWillMount () {
+    // beeep
```
alanshaw: 👌

PR #639, opened 2018-03-17
fix: api host, port and protocol

hacdias:
WIP.

Issue #611, opened 2017-12-14
IPFS UI + HTTPs API are always pointing to port 80


koalalorenzo:
#### Description:
When I am exposing the API behind a reverse proxy via HTTPS (nginx), the API call coming from the `/webui` interface are all failing as they are trying to get to port `:80`. In the logs of the browser it looks like this:

```
GET https://ipfs-api.myawesomewebsite:80/api...
```

What should I do to allow the UI to talk to the HTTPS correctly?
If I am disabling HTTPS it works perfectly.

To reproduce you can spin up a container, set up nginx to point `:443` to `:5001`. Even with a fake certificate you will see the problem by browsing to `https://localhost/webui`.

(clones https://github.com/ipfs/go-ipfs/issues/4488 as this probably is the right place to submit the issue)

Stebalien:
Preliminary warning: Please do not expose your API to the public internet.

It looks like this may have already been fixed (https://github.com/ipfs/webui/pull/602). We've recently updated the webui version in the go-ipfs master; would you mind building go-ipfs from git and trying that?

danc86:
I tried master (commit 3d8aa07) built as `QmULQD1iBBkxWkzb7nsyxpLqivjdfp4D1w2ctC8wRz8jJD`. It's definitely different, but not fixed :-)

It seems to have the opposite problem now. I see Firefox blocking the AJAX requests like this:

```
Blocked loading mixed active content “http://ipfs.djc.id.au:443/api/v0/id?stream-channels=true”
```

Note port 443 but scheme http.

koalalorenzo:
Is it possible to define this value (port, protocol, domain/ip, hash of the UI) in a env variable or somewhere in the conf? That would solve the problem and allow us to test more versions easily

hacdias:
@danc86 @koalalorenzo could you please try the branch [api-local](https://github.com/ipfs-shipyard/ipfs-webui/tree/api-local) and let me know if it works correctly?

danc86:
I built commit b2dc3ad38092eb1479849775795b41f31907a9fa as QmZ8QuzSiF5mfJSYbAA6JWBENmYsYo3vzrSPtZxHKqPxSo. It still has the same problem, tries to access http://hostname:443/... It looks like you are intentionally slicing off the final char of window.location.protocol, what's up with that? Why not just use window.location.protocol as is?

hacdias:
[...]

Issue #653, opened 2018-04-15
IPFS deployment - Please use a browser with webcrypto support


leonprou:
I deployed my ipfs-go node to AWS instance. Opening the webui in the browser (chrome 65) I a have the following error:

```
ipfs-webui.0.3.0.js:1 Uncaught Error: Please use a browser with webcrypto support
    at r (ipfs-webui.0.3.0.js:1)
    at sha2256 (ipfs-webui.0.3.0.js:1)
    at Function.r.digest (ipfs-webui.0.3.0.js:1)
    at r (ipfs-webui.0.3.0.js:1)
    at ipfs-webui.0.3.0.js:1
    at serialize (ipfs-webui.0.3.0.js:1)
    at Function.e.exports [as create] (ipfs-webui.0.3.0.js:1)
    at ipfs-webui.0.3.0.js:1
    at t (ipfs-webui.0.3.0.js:1)
    at ipfs-webui.0.3.0.js:1
```

Looking into it I found that the browser can't load modules like `SHA-256` (a webcrypto API module).

The funny thing that we webui is working in Safari (11.1 version). As I understand the specs, both Chrome and Safari should support this feature.

Jonybang:
Same problem

VictorBjelkholm:
When using webcrypto, the origin needs to be secure. This means it cannot be served via http unless it's at localhost. If you're loading something external, it needs to be via https.

```
Secure origins are those that match at least one of the following (scheme, host, port) patterns:

    (https, *, *)
    (wss, *, *)
    (*, localhost, *)
    (*, 127/8, *)
    (*, ::1/128, *)
    (file, *, —)
    (chrome-extension, *, —)
```

From http://www.chromium.org/Home/chromium-security/security-faq#TOC-Which-origins-are-secure-

leonprou:
@VictorBjelkholm thanks for explaining this 👍

Jonybang:
But webui send requests via http and throw errors:
![2018-04-30 17 27 43](https://user-images.githubusercontent.com/4842007/39435660-3bdb755a-4c9c-11e8-8f9c-eca7a6d3609d.png)

PR #664, opened 2018-05-18
Fix support for use behind https

danieldaf:
I fix and test the use of https related to the pull request #639 and the issue #611


koalalorenzo:
Can't wait to get this merged (almost a month passed)

olizilla:
We're working on a rebuild of the webui, and we'll ensure it works behind https. As a regular reminder, don't expose your API port to the public internet. That said, the current UI should work via https, so this PR is very much welcome, let's get it in!

@hacdias can you test and review this change? I think it also fixes the issue in this comment https://github.com/ipfs-shipyard/ipfs-webui/issues/653#issuecomment-385435139

hacdias:
@olizilla gonna take a look!

hacdias:
Seems to be working. Thanks @danieldaf

PR #669, opened 2018-05-23
feat: add files page

hacdias:
Start building the files page. This is a minimal version with some controls working: delete, inspect, navigation and file upload. You can also preview music, pdf, video, pictures, audio and text files.

License: MIT


olizilla:
👀 now!

src/App.js
```diff
@@ -45,4 +45,10 @@ export class App extends Component {
   }
 }
 
-export default connect('selectRoute', 'doUpdateUrl', 'doInitIpfs', App)
+export default connect(
+  'selectRoute',
+  'doUpdateUrl',
+  'doInitIpfs',
+  'doGetGatewayUrl',
```
olizilla: `doGetGatewayUrl` can be removed here, it's not used.

src/bundles/ipfs.js
```diff
@@ -27,5 +38,21 @@ export default {
     }
 
     dispatch({ type: 'IPFS_INIT_FINISHED' })
+  },
+
+  doGetGatewayUrl: () => async ({ dispatch }) => {
```
olizilla: `doGetGatewayUrl` isn't called? I think it's a good plan to have something like this. We might end up with both a "publicGatewayUrl" that points to https://ipfs.io most of the time, and an "localGatewayUrl" that usually points to localhost. I think we've got something like that in companion already. 

What you have is good for now. We need to think about how we make this work against a js-ipfs instance, as we connect to window.ipfs, so we can't be sure that the user has http gateway. We could just use the public one in that case, but it would be neat if we could load the users files directly via window.ipfs rather than assuming an http gateway.

src/files/FilesPage.js
```diff
@@ -1,2 +1,114 @@
...
+
+const action = (name) => {
+  return (...args) => {
+    console.log(name, args)
+  }
+}
+
+const empty = (
+  <div>
+    <h2>It seems a bit lonely here :(</h2>
```
olizilla: @akrych when you get time, could you create some ideas for a "you have no files yet, why not add some" page for the files view? It could use some of the space to introduce some IPFS or at least some of the issues around "they are only on your machine until a peer fetches them from you", and "this is p2p. this is not a cloud"

src/files/FilesPage.js
```diff
@@ -1,2 +1,114 @@
...
+      this.props.doFilesRename(path, path.replace(oldName, newName))
+    }
+  }
+
+  onFilesUpload = (files) => {
+    this.props.doFilesWrite(this.props.files.path, files)
+  }
+
+  componentWillMount () {
+    console.log('WUOLL MOUNT')
```
olizilla: WULLY

src/files/file-input/FileInput.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Button from '../../components/button/Button'
+
+export function readAsBuffer (file) {
```
olizilla: Lets move this data wrangling logic to the redux bundle. It's something we'd need to do regardless of what UI library we used, so it let's pull it out.

src/files/file-input/FileInput.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  onChange = () => {
+    const raw = this.input.files
+    const { upload } = this.props
+    let promises = []
+
+    for (const file of raw) {
+      promises.push(readAsBuffer(file))
+      console.log(file)
```
olizilla: Let's use https://github.com/visionmedia/debug for this kind of logging.

src/files/file-preview/FilePreview.js
```diff
@@ -0,0 +1,85 @@
...
+        )
+      case 'pdf':
+        return (
+          <object width='100%' height='500px' data={src} type='application/pdf'>
+            If you're seeing this, is because your browser doesn't support previwing
+            PDF files.
+          </object>
+        )
+      case 'video':
+        console.log('Ivdeo')
```
olizilla: Ivdeo!

src/files/file-preview/FilePreview.js
```diff
@@ -0,0 +1,85 @@
...
+          <video controls className={className}>
+            <source src={src} type={`video/${ext}`} />
+          </video>
+        )
+      case 'image':
+        return <img className={className} alt={stats.name} src={src} />
+      default:
+        const cantPreview = (
+          <div>
+            Sorry, we can not preview this file..
```
olizilla: There are going to be a bunch of media types we can't preview in a browser... @akrych can you add designs for "preview page for file types where we can't show a preview of the content". We can still ask IPFS for a file size at least, but that isn't much.

More generally, I think that doing a good job of showing a preview for all types of files is a lot of work. Like, many files will say they are are "video" but that doesn't mean we have the codecs to play them. Same for audio. 

We could do extra work and unpack zip files and show their structure but then we've got tar,7zip,rar etc which would each need custom previews.

Then there is all the text flavours. Do we go full github and try and render all of them too? W

We need to at least catch all the scenarios where the current browser can't preview a file that might normally be playable. Showing a video element, that just shows a gray box is a bad user experience, and worse than just saying we can't preview it.

We could preview files in browser tabs for now, and let the browser decide what to do with it. It's not great, but it avoids a load of edge cases for until we're ready to do all the work that a robust, general purpose preview page will need.

@lidel @alanshaw @hacdias what are your thoughts on this?

src/files/file-preview/FilePreview.js
```diff
@@ -0,0 +1,85 @@
...
+          </div>
+        )
+
+        if (stats.size > 1024 * 1024 * 4) {
+          return cantPreview
+        }
+
+        if (!this.state.content) {
+          this.loadContent()
+          return <div>Loading</div>
```
olizilla: @akrych can you give us some design directions on "page loading" and "component loading" states please.

src/files/type-from-ext/extensions/audio.json
```diff
@@ -12,7 +12,6 @@
   "aac",
   "mp2",
   "mp3",
```
olizilla: As we want to look up a file type from an encoding, a map or object would be a better mechanism for storing these lookups.

extToType.js

```js
{
  amr: 'audio',
  bw: 'audio',
  jpg: 'image',
  jpeg: 'image',
  mp4: 'video'
  // ...and on and on
}
```

This ensures that we don't put the same extension in more than one file and keeps lookups as fast as possible and allows us to easily add new icon types for specific extensions if we need to.

olizilla:
The layout for the action items needs to be tweaked, the buttons jiggle depending on how many items are selected.

webui-action-bar-icon-wiggle

Also, I suggest changing the design so that when switching from 1 selected item to 2+, the options that are no longer applicable remain visible, but are disabled and have reduced opacity. We can tooltip them to say "only available for individual files". I'd reorder the options so that the ones that get disabled are to the right hand side. The position of the buttons should remain the same regardless of how many items are selected.

olizilla:
Uploading large files kills chrome. We have this issue in companion, and we solve it by blocking the upload and warning the user.

@lidel @alanshaw should we port that logic to window.ipfs? or at lease wrap window.ipfs with it, as any app that tries to do it is going to hit this problem. It's not ideal but I think it's better that we throw an error that the developer can catch rather than letting the tab fail.

olizilla:
@akrych the files list looks beautiful, but I'm concerned that the information density is too low for a file list. The user can't see many files in a single screen. Can we tighten it up a little without spoiling the feeling?

olizilla:
@hacdias being able to click through to the IPLD explorer is totally rad. I wonder if we can ask MFS if it know the local mfs path for a given CID... that way the IPLD page could offer a link back to the files view if a node is a unixfs one.

src/files/file-preview/FilePreview.js
lidel: I agree with @olizilla , not sure if we want to go into the rabbit hole of adding file previews of too many formats.  It will be time consuming and it *could easily be added later via PRs* (great opportunity for contributors). That being said, two additional ideas below.

#### Improving UX without real previews

It may have more sense to rename it to "File Details"  (as it does not imply preview is possible).

Even if we can't play a `.mov` video or open a `.docx` document, we can visually communicate to user if a file is a video, document, archive.

To roughly determine if preview can be rendered, we could do proper mime-type sniffing via something like [file-type](https://github.com/sindresorhus/file-type) (it only needs the first 4100 bytes of every file).

In absence of  real preview for specific format,  the screen could display file size, detected mime-type with some human-readable description of the file type, a different Icon for popular mime-types (or just generic image/sound/video/document/archive/unknown) and a <kbd>Download</kbd> button to enable user to open file in third-party handler provided by OS.

#### Previews for web-safe formats

If we want to have minimal support for previews we should aim at low hanging fruits: images, text files and basic audio and video player for [formats supported by the browser](https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats) with a fallback <kbd>Download</kbd> button if it can't be played for some reason.

src/files/FilesPage.js
hacdias: I don't even remember why I added this listeners... 😆

lidel:
> Uploading large files kills chrome. [..] should we port that logic to window.ipfs?

@olizilla  extracted this discussion into separate issue: https://github.com/tableflip/ipfs-postmsg-proxy/issues/33

src/bundles/ipfs.js
hacdias: Uhh, forgot to call it. Now it's called after doInitIPFS. 😄 It doesn't work when using companion though because it doesn't allow `.config.get`.

lidel: We should start thinking on adding `window.ipfs` in "privileged mode" for WebUI, which provides access to all APIs and has no sandboxing. 

Initial idea was to add a mechanism for pre-emptively requesting permissions, as suggested in https://github.com/ipfs-shipyard/ipfs-companion/issues/454#issuecomment-381245084 and https://github.com/ipfs-shipyard/ipfs-companion/issues/454#issuecomment-381278859, it could be extended with sandbox flag:

    window.ipfs.proxy.acl.whitelist(['config.get'], {sandboxing: false})

The problem is that other dapps could easily disable sandboxing this way. 
Not sure if we are able to secure it without some sort of pubkey crypto, which requires us to maintain key for signing webui reuest. 

Is this a blocker or just an inconvenience? 
Are you able to work with HTTP API until we sort this out?

hacdias:
@olizilla thanks for your review 😄 I don't think we can take an hash and get its path on MFS right now. There doesn't seem to be an API for that.

akrych:
> @akrych the files list looks beautiful, but I'm concerned that the information density is too low for a file list. The user can't see many files in a single screen. Can we tighten it up a little without spoiling the feeling?

Yeah, we can tight it up :) I think we can make 18px margins on the top and bottom of file icon

olizilla:
@hacdias can you fix up the action bar layout jiggle, then we can merge this.

PR #672, opened 2018-05-24
feat: IPLD graph for explore page

olizilla:
This PR adds a graph to the "explore" page to show the links from the current node and their target node type. It uses [cytoscape](http://js.cytoscape.org/) and [dagre layout](https://github.com/dagrejs/dagre).

screen shot 2018-06-11 at 12 55 45

|  CBOR node | Storybook  |
|-------------|----------------|
| screen shot 2018-06-11 at 12 56 19 | screen shot 2018-05-24 at 11 27 42

### Introducing GRAPH CRUMB!

It also adds a "graph crumb" component, which shows the node containment and path boundaries that have been traversed along the user provided path.

graph-crumb-explained

This change required upgrading the logic for resolving info from them the path. Previously we passed the `cid+path` to `dag.get` and tried to render whatever value came back. As dag-cbor node values don't provide the same info as dag-pb nodes _(no `links` or `multihash` properties)_ this pushed complexity on to the UI code of working out how to handle the differences. This PR fixes that by normalising data from `dag.get` into a common shape. The `normaliseDagNode` function currently handles `dag-pb` and `dag-cbor` but the plan is it can be extended to normalise _all the things._

`resolveIpldPath` gives the additional info about the path by recursively calling `dag.get` from the root cid to find all the path boundaries (links that have to be traversed) to resolve the entire path. The normalised array of `nodes` and `pathBoundaries` that were crossed are made available to the the UI so the Graph crumb component can visualise it.

License: MIT


whyrusleeping:
In the case where you have "favorites/{1,2,3}" it would be cool to represent "favorites" as its own node in the graph.

whyrusleeping:
@olizilla I think that showing internal structures (of every potential node type, not just cbor) would be very useful. cc @diasdavid

olizilla:
@whyrusleeping @diasdavid please do add ideas for IPLD graph visualisations over here https://github.com/ipfs-shipyard/pm-ipfs-gui/issues/11

olizilla:
Sorry, this PR has grown much larger than I would have liked, but there were a bunch of details to resolve in it around node normalisation, path resolving, and handling paths that land somewhere in the middle of a cbor node.

olizilla:
@lidel please can you take a look at this when you get a moment.

olizilla:
The dag node normalisation logic tries to nudge things towards a more coherent naming scheme; `normaliseDagNode` returns objects like:

```js
{
  cid: 'zHash1' // `cid` instead of `hash` or `multihash`
  type: 'dag-cbor' // or 'dag-pb'
  data: { /* the value of a dag-cbor node or the Data prop of a dag-pb */ }
  links: [
   {
      path: 'a/b/c' // in dag-cbor a link name may be a path, so standardise on `path`
      source: 'zHash1' // CID of node that contains this link.
      target: 'zHash2' // CID that the link points to.
  }
}
```
We map dag-cbor and dag-pb nodes returned from `dag.get` into this shape which then reduces the amount of conditional logic needed in the UI code. Adding the `source` cid to links is super useful as the link data can then be used in isolation from the parent node and still make sense, the source CID + path is a permanent-web alias for the target cid.

olizilla:
Am resolving the conflict now

src/explore/graph/IpldGraphCytoscape.js
```diff
@@ -0,0 +1,139 @@
...
+
+  ipfsLinksToCy (links) {
+    const edges = links.map(this.makeLink)
+    const nodes = links.map(this.makeNode)
+    return [...nodes, ...edges]
+  }
+
+  makeNode ({target, path}, index) {
+    // TODO: pass link type info.
+    const type = target.startsWith('Q') ? 'dag-pb' : 'dag-cbor'
```
lidel: Yeah, someone will take [CIDv0 of dag-pb](https://gateway.ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR) and [convert it to CIDv1](https://gateway.ipfs.io/ipfs/zb2rhk6GMPQF3hfzwXTaNYFLKomMeC6UXdUt6jZKPpeVirLtV) as a test and get [weird results](http://localhost:3000/#/explore/ipfs/zb2rhk6GMPQF3hfzwXTaNYFLKomMeC6UXdUt6jZKPpeVirLtV), like I did just a moment ago :upside_down_face:

src/lib/dag.js
```diff
@@ -19,62 +36,69 @@ export function explainDagPb (node) {
...
 
-export function explainDagCbor (obj) {
-  const links = findIpldLinks(obj)
+/*
+ * Convert DagLink shape into normalized form that can be used interchangeably
+ * with links found in dag-cbor
+ */
+export function normaliseDagPbLinks (node) {
+  return node.links.map(({name, size, multihash}) => ({
+    path: name,
```
lidel: :bike: :house:   since `path` is actually either a `path` (cbor) or a `name` (pb), perhaps `context` or `label` would be a less confusing term?

olizilla:
@lidel thanks for this, that's all really helpful. I strongly agree about making it super clear to the user what's happening and being robust when we find more exotic nodes. This PR gets things to a coherent point where it _can_ work and is large, so I vote we merge it and tackle the issues raised in subsequent PRs.

Totally up for discussing, but right now I think `path` is a good name for the property as a `name` is a `path` of length 1.  I see `path` as a superset of `name`, and we're using the term path for a "0 or more slash separated node traversal steps" ;-)

PR #678, opened 2018-06-12
chore: extract e2e test from unit tests

olizilla:
This change lets us run the unit tests without triggering the e2e tests.
This keeps the unit tests fast and light, to encourage developers to
run them in watch mode as they code.

This also adds react-helmet to control the page title and updates
the e2e navigation test to check for the page title value rather
than an h1 element which is not called for in the designs.

License: MIT


olizilla:
@hacdias good thing too... The e2e test didn't have a good way to wait for the js to update the page title element, but now it does!

PR #683, opened 2018-06-19
feat: add peer bandwidth bundle

alanshaw:
Not styled properly, just functionality. Ready for a review.

N.B. It could probably use some performance tweaking, but I think really what needs to happen is a better API. We can't do a single call to get a list of current bandwidth usage per peer so we have to update one by one, and for 800+ nodes that's a lot of API work. I'm open to suggestions!

screen shot 2018-06-25 at 17 03 45


olizilla:
👀

src/bundles/peer-bandwidth.js
```diff
@@ -0,0 +1,172 @@
...
+        const peerIds = (peers || []).reduce((ids, p) => {
+          const id = p.peer.toB58String()
+          if (ids.seen[id]) return ids
+          ids.seen[id] = true
+          ids.unique.push(id)
+          return ids
+        }, { seen: {}, unique: [] }).unique
+
+        const added = peerIds.filter(id => !bwPeers.some(p => p.id === id))
+        const removed = bwPeers.filter(p => peerIds.every(id => id !== p.id)).map(p => p.id)
```
olizilla: You can make `removed` more like `added` and get the stop early benefits of `some` again.
```js
const removed = bwPeers.filter(p => !peerIds.some(id => id === p.id)).map(p => p.id)
```

src/bundles/peer-bandwidth.test.js
```diff
@@ -0,0 +1,298 @@
...
+  opts = opts || {}
+  opts.minLatency = opts.minLatency || 1
+  opts.maxLatency = opts.maxLatency || 100
+
+  return {
+    stats: {
+      bw: () => new Promise((resolve, reject) => {
+        setTimeout(() => {
+          resolve(fakeBandwidth())
+        }, randomInt(opts.minLatency, opts.maxLatency))
```
olizilla: These tests are rad. Excellent work Chef. Bellissimo 💋👌✨

alanshaw:
@olizilla is good for merge?

PR #684, opened 2018-06-21
Settings page config editor

olizilla:
Adds the advanced mode config editor to the settings page. The editor expects valid JSON, and warns the user when it isn't.

webui-config-edit

We fetch the node config from the api. This is stashed in the redux
tree, and is only updated by re-fetching from the api.

The settings page manges a local mutable copy that is initialised
on first render. The save button becomes active when there is a
change in the editor. If the user reverts their changes the save
button becomes disabled again to indicate in a calm way that there
are no changes.

In the background we periodically refetch the config from the api.
If we detect that the config was changed since the user started
looking at it in the webui, we show a warning and invite them to
reset the editor to see the changes.

Adds a rad storybook with [knobs](https://github.com/storybooks/storybook/tree/master/addons/knobs) on, to visualise the numerous states

webui-config-edit-2


olizilla:
TODO

- [ ] add padding to top of editor
- Improve saving flow. 
  - [ ] handle time between hitting save and getting the updated config back from api
  - [ ] disable saving invalid JSON

olizilla:
@lidel Ace doesn't directly support full line error heighlighting, but I think it could be achieved with the setMarker api and some custom styling, but I'm going to leave that for a future PR.

I'm trying to come up with a nice way to handle the `config` api being blocked when using `window.ipfs` but I think I'll just have to start with the settings tab hidden and once we successfully load the config then fade it in... that'll have to do until we figure out how to support it.

olizilla:
@lidel @alanshaw can has re-review pls? We have 
- Formatting tweaks
- No save invalid json
- Ignore external config changes that happen within 1.5s of you saving config changes. Ideally the config would be identical to what you just saved, but I don't want to rely on that always being the case.
- Hide settings where we can't config.get, and check for window.ipfs blocking the api and alert the user specifically in that case if they happen to get to the settings page directly

olizilla:
> It may be subjective, but when there is an error in JSON the Save button looks too inviting: cursor changes to pointer on hover, but click does nothing, which kinda confuses user. I think we want Reset action to pop visually instead.

Yep, I wanted to use the transition to red to signal more clearly there was a problem. It'd be easy enough to just gray out the save button. I'll fix the pointer too. I think just graying out the save should be enough. I'll leave the reset button active and in it's default state.

> LAN IP seems to ignore window.ipfs(?) and connects to HTTP API port directly, showing Settings just fine, even if I have browser extension active

that's weird. I'll see if I can recreate it.

olizilla:
@lidel @alanshaw that is weird, `window.ipfs` is `undefined` when window.ipfs-fallback runs when the app is loaded over a LAN, ip rather than localhost.

alanshaw:
Is that a secure context thing? Localhost considered secure but IP addresses not?

olizilla:
ahhhh yes. that'll be it

olizilla:
aww, we've exceeded the arbitrary bundle size of 512KB that I set... I'll have to remove the settings page.

olizilla:
pew pew pew, I've added [react-loadable](https://github.com/jamiebuilds/react-loadable) so the settings page gets it's own bundle... total bundle size is still over 512kb, but the initial load is back under, which is nice.

olizilla:
@alanshaw @lidel @hacdias can i get a review pls.

src/settings/SettingsPage.js
```diff
@@ -1,11 +1,197 @@
...
+        <Helmet>
+          <title>Settings - IPFS</title>
+        </Helmet>
+        <Box>
+          <div className='dt dt--fixed pb3'>
+            <div className='dtc v-mid'>
+              <SettingsInfo config={config} isConfigBlocked={isConfigBlocked} isLoading={isLoading} hasExternalChanges={hasExternalChanges} hasSaveFailed={hasSaveFailed} hasSaveSucceded={hasSaveSucceded} />
+            </div>
+            <div className='dtc tr v-btm pt2' style={{width: 240}}>
+              { config ? (
```
fsdiogo: It's better to just:

```js
{ config && <div>...</div> }
```

src/settings/SettingsPage.js
```diff
@@ -1,11 +1,197 @@
...
+                <div>
+                  <Button minWidth={100} className='ml3' bg='bg-charcoal' disabled={isSaving || (!hasLocalChanges && !hasExternalChanges)} onClick={onReset}>Reset</Button>
+                  <SaveButton hasErrors={hasErrors} hasSaveFailed={hasSaveFailed} hasSaveSucceded={hasSaveSucceded} isSaving={isSaving} hasLocalChanges={hasLocalChanges} hasExternalChanges={hasExternalChanges} onClick={onSave} />
+                </div>
+              ) : null }
+            </div>
+          </div>
+          {config ? (
```
fsdiogo: Same here.

olizilla:
@alanshaw yep, i'm seeing that too now, I think this is an issue in the editor that seems to show up if the first character you type is a space, which annoyingly, is quite probable if you're editing formatted JSON. It's I think the browser is interpreting the first space keypress event as a "page down this document".

Can you confirm that you only see that behaviour with the first keypress, and only when its space?

src/settings/SettingsPage.js
olizilla: @fsdiogo I prefer the look of `&&` style too, but I now always use ternary to be explicit about what I want to be rendered if the value is falsey. This is due to doing a bunch of react-native work where most falsey values other than null cause an error.

olizilla:
I've updated ace-editor and it seems to be working now! I'm gonna merge this.

PR #686, opened 2018-06-25
feat: rename prompt

hacdias:


src/files/rename-prompt/RenamePrompt.js
```diff
@@ -0,0 +1,86 @@
...
+    this.props.cancel()
+    event.preventDefault()
+  }
+
+  render () {
+    let {cancel, className, action, filename, folder, ...props} = this.props
+    className = `${className} w-80 shadow-4 sans-serif relative`
+
+    return (
+      <form className={className} style={{maxWidth: '30em'}} {...props}>
```
olizilla: if you move the onSubmit handler to the form we should get "enter to submit" behaviour from the browser for free!

olizilla:
@hacdias not quite... I'm suggesting to add the handler as an `onSubmit` attribute to the form. That would be triggered when a user presses enter in the input, so we wouldn't need the `handleKeyPress` logic.

hacdias:
Oh, yes, makes sense, done now @olizilla

src/files/rename-prompt/RenamePrompt.stories.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { action } from '@storybook/addon-actions'
+import RenamePrompt from './RenamePrompt'
+
+storiesOf('Files', module)
+  .add('Rename Prompt', () => (
+    <div className='ma2'>
+      <RenamePrompt filename='my-agenda.markdown' cancel={action('Cancel')} action={action('Rename')} />
```
olizilla: `action` is a generic term; it doesn't give the reader much of an idea what it is used for. How about `onSubmit` and `onCancel` so you can see from the calling code what's expected.

hacdias:
np @olizilla. done now

PR #687, opened 2018-06-26
feat: add rename prompt to file browser

hacdias:
image

Ready to be reviewed 😄

### DIFF
--- a/src/files/FilesPage.js
+++ b/src/files/FilesPage.js
@@ -24,7 +24,7 @@ class FilesPage extends React.Component {
     clipboard: [],
     copy: false,
     rename: {
-      show: false,
+      isOpen: false,
       path: '',
       filename: ''
     }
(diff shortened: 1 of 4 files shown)


src/files/FilesPage.js
```diff
@@ -34,12 +41,27 @@ class FilesPage extends React.Component {
   }
 
   onRename = ([path]) => {
-    const oldName = path.split('/').pop()
-    const newName = window.prompt('Insert the new name:')
+    this.setState({
+      rename: {
+        show: true,
```
fsdiogo: Don't want to nitpick, but I would rename this to something like `isOpen`, that way it makes it clearer that it's a Boolean.

But I leave it up to you 🙂

hacdias: Yeah, that's a great idea. Thanks! 😄

PR #689, opened 2018-06-28
feat: add delete prompt to file browser

hacdias:
image

The commits on this PR are self contained and should be merged without squashing.

### DIFF
--- a/src/files/FilesPage.js
+++ b/src/files/FilesPage.js
@@ -7,6 +7,7 @@ import FilesList from './files-list/FilesList'
 import FilePreview from './file-preview/FilePreview'
 import FileInput from './file-input/FileInput'
 import RenamePrompt from './rename-prompt/RenamePrompt'
+import DeletePrompt from './delete-prompt/DeletePrompt'
 import { Modal } from 'react-overlays'
 
 const action = (name) => {
@@ -27,6 +28,12 @@ class FilesPage extends React.Component {
       isOpen: false,
       path: '',
       filename: ''
+    },
+    delete: {
+      isOpen: false,
+      paths: [],
+      files: 0,
+      folders: 0
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/files/FilesPage.js
```diff
@@ -35,17 +42,17 @@ class FilesPage extends React.Component {
     doUpdateHash(`/files${link}`)
   }
 
-  onInspect = (hash) => {
+  onInspect = ([file]) => {
```
fsdiogo: Why add the array here?

PS: I'm dropping in without much background.

src/files/FilesPage.js
```diff
@@ -70,6 +77,43 @@ class FilesPage extends React.Component {
     this.onRenameCancel()
   }
 
+  onDelete = (files) => {
+    let filesNumber = 0
+    let foldersNumber = 0
+
+    files.forEach(file => {
```
fsdiogo: Don't you prefer one liners when it's a simple conditional?

```js
files.forEach(file => file.type === 'directory' ? foldersNumber++ : filesNumber++)
```

src/files/FilesPage.js
hacdias: We were only sending the required info for each hook: the delete only needed paths, the inspect only needed the hash, etc.

But since now we also need the type of the file, we would have to send two different things for delete hook.

I then decided just to send everything to every hook and the hook uses the data it wants. You can see the changes on [this commit](https://github.com/ipfs-shipyard/ipfs-webui/pull/689/commits/f4e5800f8997ae3cc09df9a2534e9e13aa53302b).

src/files/FilesPage.js
hacdias: For me it's indifferent, I don't think one way is better than the other. I can change it though.

src/files/FilesPage.js
fsdiogo: Sounds good, thank you 👍

hacdias:
@olizilla let's merge this 😄

olizilla:
👀

olizilla:
screenshot 2018-06-29 10 41 03

There's an issue with the focus outline. The buttons are inheriting the focus state from the parent which looks weird. focus should generally only apply to a single element and isn't inherited. My bad, the `.focus-outline` selector is too general. Can you fix it and add it to the modal too please, the blue outline around the modal box looks out of place.

hacdias:
@olizilla just fixed 😄

PR #691, opened 2018-06-29
feat: download files

hacdias:
It allows you to download the files from the File Browser. Doesn't work with Companion yet though since it doesn't support `.config.get`.

### DIFF
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -184,10 +184,11 @@ function downloadMultiple (dispatch, getIpfs, store, files) {
       }
 
       dispatch({ type: 'FILES_DOWNLOAD_LINK_FINISHED' })
+      const multihash = node.toJSON().multihash
 
       return {
-        url: `${apiUrl}/api/v0/get?arg=${node.toJSON().multihash}&archive=true&compress=true`,
-        filename: 'download.tar.gz'
+        url: `${apiUrl}/api/v0/get?arg=${multihash}&archive=true&compress=true`,
+        filename: `download_${multihash}.tar.gz`
       }
     })
 }
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -144,14 +144,13 @@ bundle.doFilesWrite = (root, files) => ({dispatch, getIpfs, store}) => {
 function downloadSingle (dispatch, store, file) {
   dispatch({ type: 'FILES_DOWNLOAD_LINK_STARTED' })
 
-  const apiUrl = store.selectApiUrl()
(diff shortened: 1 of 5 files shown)


hacdias:
Thanks @lidel for your review. I hadn't noticed that problem yet (of having multiple files with the same hash on the same directory). I'll solve it ASAP on another PR because it's a separate subject **and** it will change more code than what I might have thought. With that, I'll also make some improvements to error handling on File Browser 😄

And it should be merged before this one.

hacdias:
@akrych @lidel @olizilla @alanshaw @fsdiogo how would you like to see the download progress?

lidel:
@hacdias if I had to guess I'd say `X-` prefix is just a legacy convention. We should be looking for both, I think, to make it future-proof.

As for the progress, perhaps for now (to keep this PR small) it will be enough to just replace label with progress %?


If you want to do more, there could be a subtle,  progress bar on the bottom  of the icon in style from [ut-kit](https://github.com/ipfs/ipfs-gui/issues/7#issuecomment-364501106):
icon with progress 50%

Otherwise we need to design something new, perhaps a variation on the round one (also from [ut-kit](https://github.com/ipfs/ipfs-gui/issues/7#issuecomment-364501106)):

hacdias:
@lidel I'll stick with the first for now. The second seems to occupy too much space. Let's wait for more opinions and leave this small yes.

hacdias:
@lidel check it out now 😄

lidel:
- Download of small file results is a [quick flash of label change](https://i.imgur.com/5DCoRMl.gif), which does not look useful. 
Not sure how to solve it, perhaps:
   -  we could skip label change for small files (eg. < 3MB)
   -  or change label to "Finished :heavy_check_mark: " and keep it that way up to  3 seconds after successful download?
- If download of bigger file in progress and I click on the button again it runs second download task and we get race between [status updates](https://i.imgur.com/MC8mwnX.gif) :upside_down_face: 
   - I think clicking on active download probably should cancel the download instead or starting it again.

hacdias:
@lidel done 😄

hacdias:
* when you move a PR to the wrong column on Waffle *

hacdias:
**Update:** apparently, using `event.total` works for single files (on `/ipfs/hash` links) and not for Tar files (on `/api/v0/get`) so I made some updates and simplified the thing.

hacdias:
Thanks @lidel. We should definitely add more animations and life to icons and actions. I'll open an issue to keept rack on that!

PR #697, opened 2018-06-30
feat: overlay component

hacdias:
Let's reuse code 😄

### DIFF
--- a/src/files/overlay/Overlay.js
+++ b/src/files/overlay/Overlay.js
@@ -1,0 +1,28 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Modal } from 'react-overlays'
+
+function Overlay ({children, show, onLeave, className, ...props}) {
+  return (
+    <Modal
+      {...props}
+      show={show}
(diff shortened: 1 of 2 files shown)


src/files/overlay/Overlay.js
```diff
@@ -0,0 +1,28 @@
...
+      onBackdropClick={onLeave}
+      onEscapeKeyDown={onLeave}>
+      {children}
+    </Modal>
+  )
+}
+
+Overlay.propTypes = {
+  show: PropTypes.bool.isRequired,
+  onLeave: PropTypes.func.isRequired
```
fsdiogo: I don't think this should be required. I imagine there'll be cases when we just want to close the modal without any other action!

hacdias: Hmm, so how would you change the `show` prop? If you don't have a `onLeave` trigger, you don't know the user wants to close the modal. And that way you can't change the `show` prop to `false`.

fsdiogo: You're right, my bad!

Don't know why, but I was thinking that the `onLeave` was from `react-overlay` and that it hid the modal by default!

PR #702, opened 2018-07-03
feat: add peer locations bundle

alanshaw:
Adds a bundle that fetches peer locations in parallel up to the configured `concurrency` level.

Use `selectPeerLocations` to get an object like `{ [peerId]: { /* location data */ } }`.

It depends on the peers bundle and will automatically get locations for all peers in the swarm.

Peers without a resolved location get the rainbow flag 🏳️‍🌈.

I added it to the peer bandwidth list on the status page:

screen shot 2018-07-17 at 10 03 22


hacdias:
@olizilla @alanshaw can this be merged?

Issue #676, opened 2018-06-06
Upload via WebUI (ipfs.files.write --create) produces different hash than CLI (ipfs.add)


lidel:
_(Originally reported in https://github.com/ipfs/support/issues/45, today it was raised on irc)_

**Discrepancy between `ipfs.add` and `ipfs.files.write --create` was discussed and explained in https://github.com/ipfs/support/issues/45, this issue is only about solving UX problem in WebUI.**

## How to Reproduce

1. Download [QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR](https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR) and save it as `file.jpg`
1. Add it back via CLI (`ipfs add file.jpg` 
1. Open `Files` screen  in [old webui](http://127.0.0.1:5001/ipfs/QmQLXHs7K98JNQdWrBB2cQLJahPhmupbDjRuH1b9ibmwVa/#/files/explorer)
1. Upload `file.jpg`
1. Right-click on it and select `Copy hash`

## Current behavior

Copied value is `QmWKsy9a3AU3Yn9SDuqUFNMnxFnsGq5KzkKgZREsr1cifv` and not `QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`.

In short `ipfs.files.write --create` (used by old webui) produces different hash than `ipfs.add`, for reasons notes in https://github.com/ipfs/support/issues/45#issuecomment-269296211:
>  The hashes are different because `ipfs files write` uses different linking structure than `ipfs add`. `ipfs files write`'s linking structure is optimized for random seeking and writing after initial creation and `ipfs add` structure is optimized for reduction of link count.
> 
> Underling data is still deduplicated as they use the same chunking, AFAIK.

### ipfs.files.write --create
```
$ curl -s https://ipfs.io/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR | ipfs files write --create /files.write.jpg

$ ipfs files stat /files.write.jpg
QmWKsy9a3AU3Yn9SDuqUFNMnxFnsGq5KzkKgZREsr1cifv
Size: 119762
CumulativeSize: 119832
ChildBlocks: 1
```

## Expected behavior

[...]

PR #703, opened 2018-07-03
feat: add folder

hacdias:
image

### DIFF
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -1,5 +1,5 @@
 import { createAsyncResourceBundle, createSelector } from 'redux-bundler'
-import { join } from '../lib/path'
+import { join, dirname } from '../lib/path'
 
 const bundle = createAsyncResourceBundle({
   name: 'files',
@@ -108,7 +108,7 @@ function readAsBuffer (file) {
     reader.onload = (event) => {
       resolve({
         content: Buffer.from(reader.result),
-        name: file.name
+        name: file.webkitRelativePath || file.name
       })
     }
     reader.onerror = (event) => {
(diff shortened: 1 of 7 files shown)


src/bundles/files.js
```diff
@@ -119,26 +119,31 @@ function readAsBuffer (file) {
...
 
-  let promises = []
+  try {
+    let promises = []
 
-  for (const file of files) {
-    promises.push(readAsBuffer(file))
-  }
+    for (const file of files) {
+      promises.push(readAsBuffer(file))
```
lidel: This may be a separate PR, but we may want to replace readAsBuffer with pull stream to improve performance when adding big files:
- https://www.npmjs.com/package/filereader-pull-stream
- `files2stream()`: https://github.com/ipfs-shipyard/ipfs-companion/blob/b1bf771a37fbf0d6d11fab06317a8144a8c4962b/add-on/src/popup/quick-upload.js#L29-L41

src/bundles/files.js
```diff
@@ -119,26 +119,31 @@ function readAsBuffer (file) {
...
+
+    for (const file of files) {
+      await getIpfs().files.mkdir(join(root, dirname(file.name)), { parents: true })
+    }
 
-  return Promise.all(promises).then(files => {
-    return Promise.all(files.map((file) => {
+    await Promise.all(files.map(file => {
       const target = join(root, file.name)
       return getIpfs().files.write(target, file.content, { create: true })
```
lidel: We should  avoid `files.write`, as it will produce different CID than `add`.
See details and workaround in: https://github.com/ipfs-shipyard/ipfs-webui/issues/676

hacdias:
@lidel I just made the files.write change and gave the dropdown some styling. I'll add each feature in a different PR to keep everything small. Could you see if the code's looking good?

src/bundles/files.js
```diff
@@ -119,26 +119,35 @@ function readAsBuffer (file) {
...
+      const path = join(root, dirname(file.name))
+      await getIpfs().files.mkdir(path, { parents: true })
+    }
 
-  return Promise.all(promises).then(files => {
-    return Promise.all(files.map((file) => {
-      const target = join(root, file.name)
-      return getIpfs().files.write(target, file.content, { create: true })
+    await Promise.all(files.map(async file => {
+      const res = await getIpfs().add([file.content])
```
lidel: I wonder if we should disable pinning when `add`ing for MFS. 
As long as files are in MFS, they won't be GC'd.
If we dont pass `{pin: false}` to `add` then removing file from MFS won't remove it from node.  
Is that a [bug or feature](https://ipfs.io/ipfs/QmVnLkRxWWv2xiPKeQy6re1zCepzX7YXUiq5fESkd1j8bC/bug_or_feature.jpg) ?
@olizilla  any thoughts on this?

hacdias: I'm adding that option then.

About the progress of uploading: don't worry. Won't be doing in this PR but it is on Invision how it is supposed to be.

hacdias: Done @lidel

hacdias:
Done @lidel

src/bundles/files.js
olizilla: Yes please. Is ok for separate PR, but we've gotta make this the best file uploader we can.

src/lib/path.js
```diff
@@ -104,3 +104,9 @@ export function join (a, b) {
   if (b.startsWith('/')) b = b.slice(1)
   return `${a}/${b}`
 }
+
+export function dirname (path) {
```
olizilla: Let's just use node core `path` stuff for this kind of thing, https://nodejs.org/api/path.html#path_path_dirname_path

olizilla: Of note this file was intended to be a lib for ipfs/ipld path helpers, rather than any sort of path.

hacdias:
Done @olizilla

olizilla:
if nothing else is using our `join` impl, can you remove it pls. If there is usages, pls update them.

hacdias:
Done @olizilla. Since I've already solved the issues you and @lidel mentioned, I'm going to merge this so I can continue to the next feature 😄

Issue #704, opened 2018-07-03
File Browser Revamp


hacdias:
### Tasks

- [x] #669 Add files
- [x] #703 Add folder
- [x] #705 Clickable icons 
- [x] #708 Add by IPFS Path
- [x] #709 Make dropdown more similar to [Invision design](https://projects.invisionapp.com/d/main/default/#/console/13924274/302170315/preview)
- [x] #713 Fix error: select files and click another folder
- [x] #714 Validate renaming
- [x] #717 Truncate text
- [x] #718 File Status (Postponed due to DHT api)
- [x] #719 Use pull streams; see https://github.com/ipfs-shipyard/ipfs-webui/pull/703#discussion_r199892990
- [x] #720 Context Menu
- [x] #721 Option to create a new folder
- [x] #723 Merge status with peers column
- [x] #723 Add hash below name 
- [x] #723 Progress on adding files/folders
- [x] #729 Drag'n'drop files
- [x] #736 Error feedback
- [x] Fix https://github.com/ipfs-shipyard/ipfs-webui/pull/736#issuecomment-408360348

### Notes

> From @lidel on #691 (comment):
>
> > ps . Idea for separate PR: I feel we are lacking some kind of visual feedback for user when the button/icon is clicked (fade-out+fade-id of the icon alone, or maybe other way around, making svg stroke more vibrant?).
>
> @fsdiogo: Agree, we should have some feedback while hovering too!

PR #708, opened 2018-07-05
feat: add by path

hacdias:
Add files by IPFS path.

image

image

image

image

On this PR, I also created a generic `TextInputModal` which was used for the modal above and for rename modal.


olizilla:
Nice! some minor issues when trying it out:

`dim` on the menu items feels kinda odd, as it makes the hovered thing feel _less_ active rather than more. It could look clearer with `bg-animate hover-bg-near-white` in place of `dim`. Your'd need to remove the horizontal padding from the container, and replace the margins with padding on the menu items.

See: http://tachyons.io/docs/themes/skins-pseudo

I just copied and pasted a path from  https://archives.ipfs.io/ and it told me it was not valid... turned out there was some rogue whitespace at the front of what I had pasted, but I couldn't see it as the input is narrow and paths are loooooong... can we make the input wider and the font smaller, and trim whitespace on paste pls?

More generally we need to focus on robustness and showing good loading info as the next priority, as the files section is working well, but the UI keeps feeling like it is stuck for me, when I try and preview large files (perhaps we should open large files in a browser tab rather than in line) and large dir listing can take a really long time to load and display.

hacdias:
> dim on the menu items feels kinda odd, as it makes the hovered thing feel less active rather than more.

Done 😄 

> can we make the input wider and the font smaller

Wouldn't it look odd side by side with other modals?

> and trim whitespace on paste pls?

Yes, we can.

> More generally we need to focus on robustness and showing good loading info as the next priority, as the files section is working well, but the UI keeps feeling like it is stuck for me, when I try and preview large files (perhaps we should open large files in a browser tab rather than in line) and large dir listing can take a really long time to load and display.

Yes, it's the next step. Already mentioned that on #704 (visual feedback).

Issue #711, opened 2018-07-06
IPFS node selector


olizilla:
Right now we try `window.ipfs` and fallback to js-ipfs-api with the default config. If that fails to connect, we just retry until it does, without telling the user whats going on. This could be so much better.

When we're first run, if `window.ipfs` is available, we should let the user know that we're using it and that they are clearly wizards for having it installed. If not, we instantiate a js-ipfs-api (as we do now) on the default URL request the the node id with a short timeout. If we get no response, we should communicate with the user... we can ask if they just need to restart their local ipfs daemon, or provide a different API url if they're running with a customised config, or would they like to run an embedded js-ipfs node?

The "would you like to run a js-ipfs node" could be framed as "Hey, install ipfs-companion, and it'll run a js-ipfs node for you." which'd be great for ipfs-companion visibility... and could give us a bunch more users. On the flipside, it's out of band and adds a little friction when and we could just ask if they want us to install a js-ipfs node in a service worker at the click of a button. Either option would be rad. I'm leaning towards "hey why not install companion" as I think it's a big win to be visible in the browser.

lidel:
I lean towards keeping options simple for now. 

We need to remember that regular user won't know what service worker is, and I don't think we should mention js-ipfs as a viable option until it has the same connectivity magic (dht, relay discovery) as go-ipfs.  Due to this installing Companion alone does not solve the problem of missing node yet (IMO).

A safe way to communicate things today would be to do basic heuristic to find IPFS API<sup>[1]</sup>, and if that fails, prompt user to  "manually start local node at port `:5001` or install IPFS Desktop".  

We can refine this after poc integrations with go-ipfs and companion land.

----

1. via old-school js-ipfs-api client and default port or by detecting it runs in privileged context such as WebExtension and getting API object directly, as noted in https://github.com/ipfs-shipyard/ipfs-companion/issues/514

olizilla:
We added the "Set your API address" feature in https://github.com/ipfs-shipyard/ipfs-webui/pull/715

PR #715, opened 2018-07-06
feat: guide the user when no api found

olizilla:
Inform the user when we can't connect to the API and give them the option to set a custom api address.

screenshot 2018-07-06 17 25 51

Early WIP on #711 

License: MIT


hacdias:
Still missing:

- [ ] Redirect
- [ ] Reload config

hacdias:
Requires:

- [ ] https://github.com/ipfs-shipyard/ipfs-redux-bundle/pull/1

hacdias:
@olizilla I've made some changes. Please take a look to [ipfs-redux-bundle](https://github.com/ipfs-shipyard/ipfs-redux-bundle/pull/1).

- If we can't connect to an instance, it redirects to `/welcome`.
- It won't let you open any other page if there isn't a valid IPFS instance.
	- This makes our navigation tests fail on CI because there isn't an IPFS instance running. Suggestions? We could start a js-ipfs instance so it connects and does everything else.
- Cache! So, you set an API address, reload the page, and it's still the same address.
- You can always go to `/welcome` and change the API url.

hacdias:
@olizilla 

> Redirecting the url when ipfsReady is false feels like something that we should set up in redux.

It is on `redirects` bundle.

> We should catch the condition that the welcome page bundle has loaded, but IPFS isn't ready and there is no error. It can just show a loading spinner on the welcome in that case.

Done.

## When failed to connect.

image

## After being successfully connected

image

olizilla:
@hacdias To complete this challenge please appease the CI gods. The tests should pass and the test coverage should not go down. Of note the navigation test that is failing was really a place holder for more exciting e2e tests. Do you accept this mission?

hacdias:
For this PR I can add tests with and without a node running. We can then think about more tests haha. Sounds good @olizilla?

olizilla:
@hacdias 💯

test/e2e/navigation.test.js
```diff
@@ -1,17 +1,40 @@
...
+  page.waitForFunction(`document.title === 'Welcome to IPFS'`, {timeout: 5000})
+}, ms.minutes(1))
+
+it('Navigation test: node running', async () => {
   // Save 3s per run by using the initial page rather than creating a new one!
   // const page = await browser.newPage()
   const page = (await browser.pages())[0]
 
+  await page.evaluate(() => {
+    window.ipfs = node
```
olizilla: @hacdias ok so, this is a rad idea, but I don't think it'll work. With the e2e tests things get complicated, as the test is running in jest in node, and firing up a browser instance with it's own seperate runtime. I don't think we'll ever be able to create an ipfs instance in a node runtime and smoosh it into a browser without very weird things happening.

Much more minor, the api for `page.evaluate` requires you to pass any args you want to hand to the page, and they get magically transferred as args to the callback function when it is executed in the context of the page instance. Even so, like going through a black hole, an ipfs instance won't survive the trip.

One solution is to create a fake window.ipfs object with an id function that resolves with a test identity info, and then check for that. We don't need to test that js-ipfs works in the webui tests, so it's ok to stub it out.

hacdias:
@olizilla made some updates and mocking but it seems to fail here: https://github.com/ipfs-shipyard/ipfs-webui/pull/715/commits/6a0f41dd6ff1caadef44c3e9a9c3910988415239#diff-0238dfb49a06ba796757bff9fcff0faaR41

There might be something on the files page that blocks the rendering? I can't see what it might be though. I tried adding empty functions for files.ls and files.stat but it didn't work.

olizilla:
@hacdias i'll take a look at this now

hacdias:
Thank you @olizilla

PR #717, opened 2018-07-09
feat: truncate text and add tooltip on hover

hacdias:
Truncates the text when it overflows and adds a tooltip on hover.

localhost_3000_

localhost_3000_ 1


hacdias:
I'm merging this because it seems okay. If you find something off or that could be better, just ping me here.

PR #720, opened 2018-07-09
feat: context menu

hacdias:
I was starting to create the context menu. However, it seems to be a lot intrusive with the bar below:

image

What's your opinion? 

/cc @alanshaw @fsdiogo @lidel @olizilla @akrych


lidel:
Is right-click on a file checking the checkbox and toggling the bar? 
My intuition for context-menu was to be ad-hoc command for a single file that does not touch checkbox logic.

hacdias:
@lidel yes, it's checking the checkbox

fsdiogo:
The right-click shouldn't tick the checkbox, that way I think it would work.

On the other hand, if the user ticks the checkbox and then right clicks, well, that's on him.

hacdias:
So, the context menu should must affect *one* file?

On Windows at least, it affects every selected file.

hacdias:
That's also how Google Drive and Onedrive work. Dropbox doesn't have context menu at all. It has a tree dots option like we have on Invision. Maybe we should stick the Context Menu for those three dots on the left of the icon and not have a right-click context menu. What's your thoughts?

hacdias:
image

Something like that ^

hacdias:
The [Invision designs](https://projects.invisionapp.com/share/AVGRKKJ75KT#/screens/297704376) have the three dots on the left side near the checkbox button but I thought it already had too many actions and places to click on the left side so I decided to put it on the right. Maybe if the tree dots were rotated 90 deg it would also look nice

image

hacdias:
localhost_3000_

fsdiogo:
Good call @hacdias, the three horizontal are pretty familiar and look good!

hacdias:
localhost_3000_

Also, updated the breadcrumbs to be more similar to IPLD Explorer's.

Issue #700, opened 2018-07-03
Peers page revamp


olizilla:
Initial design is here: https://projects.invisionapp.com/d/main#/console/13924274/302167366/preview

<img width="912" alt="screenshot 2018-07-03 08 45 33" src="https://user-images.githubusercontent.com/58871/42205887-8a031264-7e9d-11e8-88d0-91ba14b4bff6.png">

The design is based on the redux'n'sagas branch over here: https://github.com/ipfs-shipyard/ipfs-webui/pull/229#issuecomment-285348083

As a first pass please review the UI work from the `redux` branch and where it makes sense, port it to a branch off of `revamp`. The revamp branch has data in the store for peers already. It needs to be extended with logic for geoip look ups. That involves https://github.com/ipfs/ipfs-geoip and their is prior art for using it in both the master and redux branches.

In the revamp branch we're using redux-bundler to simplify our redux logic, see https://reduxbundler.com/

olizilla:
Of note @alanshaw is working on the [peers bandwidth graph](https://github.com/ipfs-shipyard/ipfs-webui/pull/688) for the stats page and there is significant overlap here with that work (flags and geoip plumbing). Please co-ordinate your efforts here.

olizilla:
- emoji country - flags https://github.com/danalloway/react-country-flag
- ipfs-geoip - https://github.com/ipfs/ipfs-geoip - would be great to re-import the data with the latest geoip mappings.

PR #726, opened 2018-07-16
feat: peers page

fsdiogo:
Tasks:
- [x] World map
- [x] Peers table ([react-virtualized](https://github.com/bvaughn/react-virtualized))
- [x] Dynamic locations in the map 
- [x] Flags in the table
- [x] Merge https://github.com/ipfs-shipyard/ipfs-webui/pull/702 first

<br>

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/700.


fsdiogo:
@hacdias Slow when scrolling or fetching the flags? I'm using `react-virtualized` to only render the visible rows, but you can scroll to see more. 

I can add the margin, but it was not supposed to appear `US` there, only flags 😅

hacdias:
@fsdiogo it seems that Windows Flag emojis are just the country code sadly. Yesterday it seemed more slower. When I resize the page and scroll it seems a bit slow

fsdiogo:
@hacdias ohh, so the flags aren't rendering on Windows? 😓

Can you test the following please (I don't have a windows machine):

In the `src/peers/PeersTable/PeersTable.js`, line `14`, can you change that to:

```js
{ rowData.flagCode ? <CountryFlag code={rowData.flagCode} svg /> : '🏳️‍🌈' }
```

That way we are force rendering the flags to SVG instead of emoji. Maybe it will work.

hacdias:
@fsdiogo 

Yes, that way they render:

image

Still, we need some margin there

hacdias:
@fsdiogo merged the #702

fsdiogo:
@hacdias awesome, thanks!

It doesn't look as good as with the emojis though:

screen shot 2018-07-19 at 11 04 05

We have two options here:
1. Always render the flags as SVG
2. Use something like [platform.js](https://github.com/bestiejs/platform.js/) to detect if it's Windows, and render as SVG accordingly

What do you think?

hacdias:
Yeah, the emojis are better even though they are platform dependent. You can render with SVGs on Windows or let it use the Windows "emojis". If we render SVGs conditionally, I don't think there's a need to add a dependency when we can just:

```js
svg={window.navigator.appVersion.indexOf('Win') !== -1}
```

src/bundles/peers.js
```diff
@@ -9,6 +9,43 @@ const bundle = createAsyncResourceBundle({
   checkIfOnline: false
 })
 
+bundle.selectTableData = state => {
+  const peers = state.peers.data
+
+  return peers && peers.map((peer, idx) => ({
+    'id': peer.peer.toB58String(),
+    'address': peer.addr.toString(),
+    'location': 'New York, United States'
```
hacdias: Why's New York here?

fsdiogo: Oops, that selector is not being used, I made a new below. I'll commit a fix.

src/peers/PeersTable/PeersTable.css
```diff
@@ -0,0 +1,4 @@
+
+.PeersTableContainer {
+  height: 220px;
```
olizilla: Minor, but I'd just use an inline style attributes for this as the value is duplicated in the definition of the table. I'd inline `.MapContainer` as well, just to keep all the definitions together.

fsdiogo: Agree, and that way we can delete those two CSS files 👍

fsdiogo:
Can we merge this? 🚀

hacdias:
@fsdiogo here we go 😄

PR #729, opened 2018-07-18
feat: drag'n'drop files

hacdias:
- [x] Drop Files and Folders
- [x] Move files and folders by drag and dropping them
- [x] Move multiple files ^^

License: MIT


fsdiogo:
@hacdias everything seems to be working fine 👌

I just think we should hint that it's possible to drag and drop files to upload them, but not sure how 🤔

hacdias:
I'm merging this to master. If there is anything that can be improved, we can do it on another PR 😄

Issue #730, opened 2018-07-19
Notes on Cache and Performance


hacdias:
So, we're building the new WebUI and we want it to be amazingly good and blazing **fast**. Some of the last changes introduced, such as the #702 and #718 in the future, can slow down the pages a bit (at least on my computer on Chrome).

Why? #702 because it is always fetching the locations for our peers. I have 800+ in this moment so as soon as I open the WebUI what happens? It starts fetching all of them, even when we're not on the peers page, with a maximum of 10 concurrently. During that time, opening the File Browser to add files or even just navigate can be a pain. After having the info about all peers, it gets normal.

On #718 the same thing would happen.

I've got two different suggestions that could be applied to solve this issues.

## 1 Caching

Why not cache the peer location's information for **x** days and then update it? Why not cache a file's health for **y** hours before updating it again? Those are not really node-dependent informations so it wouldn't matter much. It wouldn't harm anyone, and we'd have a more fluid WebUI. We could also provide some option to clean the cache or tell the user that there is a cache that can be cleaned through the browser.

Uh, and we could also cache information on Explore panel. After all, a hash is linked to a content. But I'm not sure if IPFS doesn't do this already internally when we fetch stuff from the DAG api. I still don't know how it will work for other IPLD things (like Git and Ethereum). Correct me if I'm wrong.

Cool links:
- https://reduxbundler.com/api/included-bundles.html#createcachebundlecachingfunction
- https://github.com/HenrikJoreteg/money-clip

## 2 Web Workers

We could perform secondary data fetching in Web Workers to make the interface more responsive while we're fetching peers, peer informations and other background tasks that aren't crucially needed.

Some background fetchings we could make:
- Peers
- Peer Locations
- Files find provs

Some things done in the main process:
- Files stuff
- Config stuff
- Everything else absolutely needed for a specific page.

---

So, this is it. I'd personally like to take a shot at the caching stuff if you think it's a great idea. /cc @ipfs-shipyard/gui

olizilla:
I'm very much in favour of this. It's on the roadmap, we just wanted to hit the problems first before adding caching... having IPFS running on the same machine we the webui should make lots of look ups so fast that we don't need caching... but the api for peers is so low-level that we end up having to do a lot of work at the app layer. The longer term fix that @alanshaw is thinking about is to propose a new peers api that does the work of tracking peers bandwidth over time **and** bandwidth usage by root CID.

[...]

PR #732, opened 2018-07-19
feat: cache peer locations for 1 week

hacdias:
Taking in account what I said on #730, I tried to do a little experiment with Peer Locations. It doesn't add much more code and it is clean and improves performance over time. I set the peer location cached for a week, but maybe we could set it to expire in more time or even don't expire at all.

It uses IndexedDB, which in case of not being available in that browser, just fails silently. This cache is supposed to be something to improve the app and not really required to be used.

License: MIT


olizilla:
@hacdias did you see https://reduxbundler.com/api/included-bundles.html#createcachebundlecachingfunction
and [`bundle.persistActions`](https://reduxbundler.com/api/bundle.html#bundlepersistactions) ?

With those configured we'd have a mechanism to cache entire bundle states after specified actions occur. That is a different idea to having certain data look ups cached, so it might be that we need both techniques, but I wanted to check that you'd seen the built in mechanism, as it may be that we can refactor things to just use that.

hacdias:
@olizilla I guess so

hacdias:
Some more thoughts: I have been trying with `createCacheBundler` and I've hit only one issue: on Peers Location bundle, for example, we have the following state:

```js
{
	locations: [...],
	queuingPeers: [...],
    resolvingPeers: [...]
}
```

And the cache bundle saves it all even though we just want to cache the locations themselves and not anything else. We would need to make some adaptations to the bundle.

Second point: if we're not using `createAsyncBundle` the expire date will be the same for all data. Otherwise we would need to create a mechanism to check that.

Third point: since it saves everything on one entry on IndexedDB, if we set the expire date for one week, all of the data might happen to expire at the same time or never expire if we're always updating that bundle data (hence updating the last modified date).

I hope I've been clear enough because I know I wasn't crystal clear with this message 😞 

To finish: I'd say the idea on this PR is much more flexible: if a bundle wants to have a cache, it will get it through a selector or update it to be something more generic like:

```

export bundleCreator(opts) {
	opts.cache = opts.cache || {}
	const cache = getCache(opts.cache)

	return {
		...
		// use the cache in the bundle
		...
	}
}
```

I'd like to hear your opinions.

src/bundles/peer-locations.js
```diff
@@ -25,9 +27,13 @@ export default function (opts) {
     resolvingPeers: []
   }
 
-  return {
+  return applyCache({
     name: 'peerLocations',
 
+    cacheOptions: {
+      maxAge: ms.weeks(1)
```
olizilla: So we can provide a version number here in the future if we want to invalidate existing caches, say if we totally change the structure of the data. Very cool.

https://github.com/HenrikJoreteg/redux-bundler-example/blob/0f23da51a32853a2d83aaf26e39e78f34997debc/src/utils/cache.js#L5-L14

That example also talks about adding in some unique identify where cached data is specific to a given user... we might want to cache info specific to a given ipfs node identity at some point so it's nice to know that there is a way to manage that. ✨

olizilla:
@hacdias I think this is a cool solution. Your write up of how this compares with the existing caching solution is super helpful. 

I think we can use both caching styles. The proposal in the PR makes sense for storing specific lookups, and doing so explicitly and procedurally, so the dev has full control of when it happens.

For simpler situations where the bundle is created with caching in mind, then we can just use the built in `persistActions` support and have it store the whole bundle.

As for "how long should we cache peer locations for" I think a week is fine. As a future PR if we cached the look up of IP address to location rather than peer address to location, we could cache them for much longer as the mapping will only change when the geoip db gets updated.

PR #734, opened 2018-07-20
[WIP] feat: peer country chart

alanshaw:
A pie chart showing the top 4 countries peers reside in.

screen shot 2018-07-18 at 15 54 07

### DIFF
--- a/src/status/CountryChart.js
+++ b/src/status/CountryChart.js
@@ -1,0 +1,65 @@
+import React, { Component } from 'react'
+import { Pie } from 'react-chartjs-2'
+import { connect } from 'redux-bundler-react'
+import PropTypes from 'prop-types'
+
+export class CountryChart extends Component {
+  static propTypes = {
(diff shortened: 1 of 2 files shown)


olizilla:
@alanshaw is there anything specific you wanted to add to this PR? I'll adopt it, if you wanna brain dump any todos you had in mind.

alanshaw:
I **think** all that was left here to do here was styling. I believe the design had one segment slightly larger than the rest and tooltips instead of a key.

hacdias:
I'll take a look at the styling then 😄

hacdias:
image

@olizilla @alanshaw I think we should remove the labels because we can get an enormous list of countries. Otherwise we can keep them at the bottom to match #764

olizilla:
@hacdias we need some kind of label on that graph without having to mouse over, otherwise it's just a mystery pie!

olizilla:
Let's merge this in, and merge it in to the status page PR.

PR #736, opened 2018-07-20
refactor: files page improvement

hacdias:
### On this PR:

- [x] Bundle Rewritten
- [x] Action bar changes width accordingly with the sidebar size
- [x] Update listing info every 1 minute
- [x] Better error handling

----

On behalf of what I said on https://github.com/ipfs-shipyard/ipfs-webui/issues/730#issuecomment-406648561, this little PR increases *a lot* the responsiveness of the File Browser page if we just dump the `reactX` function that is supposed to decide when to fetch files.

I doubt this is a very Reacty way of doing this stuff, but I'd like your opinion.

And I just noticed the Files Bundle has a lot of trash we don't need... such as lot of actions being dispatched for no reason.

The other actions don't work as expected. It's just a proof of concept that would make File Browser page a lot faster.

/cc @olizilla @lidel @alanshaw @fsdiogo


olizilla:
## Source of truth and stale data...

ipfs.files.ls('/', {l: true}) is going to return a snapshot of your MFS root.

We can't guarantee that what is in the redux tree is up to date. The user can edit the MFS contents from the command line or other apps. So we need to treat the data as a possibly stale snapshot of how things were, and code for the case where things have changed since we last read it.

There is some initial work towards that in the config bundle, where I store the response from the api, and have the user edit a copy of it. As it uses asyncResourceBundle, with a 60 second freshness, we poll the api to check that there hasn't been any out of band changes to it. If there has, we inform the user in where their local view is out of sync.

In the case of file browsing, we need to figure what situations might be worth explaining to the user, where remote changes have been made to the thing they are editing. If you're just browsing and files are added else where, it's probably not worth troubling them. If you're deleting a directory and something has added extra files to it while you were looking at it, that'd be worth catching and warning the user about. We could push this back to "future work" but we should think about it as we are making decisions on how to structure our files bundle code.


## Error handling and recording the active actions in state

All our ipfs.* calls can (and will) throw at some point. Lets be super robust to API errors. Use async/await and wrap those calls in a try/catch. Dispatch redux actions like

*_STARTED
*_FAILED
*_FINISHED

for those the lifecycles. 

We should be keeping track of all the pending tasks and their state in the redux tree, so where we call out to `ipfs.files.rm` we should be setting state in the redux tree that captures the fact that there is a `files.rm` action in progress... If we want to show a delete button as grayed out while a delete is in progress, we need to have a property in the redux state tree that captures the "delete in progress"^. It shouldn't be left to the UI to track. Calling a doX action should update some state that the UI can render. I'm gonna get a "UI as a function of state" tattoo.

https://rauchg.com/2015/pure-ui

If an action fails, we want to alert the user, so again, it should be updating state. We can't rely on return values from doX action creators in the UI.

This could lead to a pile of boilerplate, so we'd want to follow the example of createAsyncResource here, and consider creating a bundleCreator function that we use to wrap these calls to "do something with IPFS". We could compose bundles out of more bundles


^ This assumes that the callback / response from the call only returns when the action is completed... need to check that.


## Reacting vs Doing... user initiated changes vs background tasks.

It's preferable to use `reactX` so that drive our app state and transitions from redux. We could replace React with vue/svelte/yo-yo without touching the redux plumbing. That said, it's not a problem to trigger `doX` actions directly from the UI, but it's a coupling that we should avoid where we can.

In the specific case of the files perf, we have an interesting situation. The peers work is generating so many actions and reactions, that the files reactor is now having to wait for a long queue of requests to complete.

This strongly suggests we need to spec out a better api for getting bandwidth stats! It also raises the question of how we deal with priority of actions. We want user-initiated actions to run before "background tasks". We want the files list to come back quickly, but we don't really mind how long the bandwidth calculations take. I see 3 options right now:

1. Reduce the reactor queue size by making the peers bundle trigger fewer actions, by (long term) proposing a better ipfs api, or (short term) refactoring it to not rely on redux actions.

2. Use reactions for "background tasks" and call doX directly from the UI to trigger user initiated transitions, which gives us an informal mechanism for priorities... reactX is lower priority (wait your turn), while doX is high priority (user wants it now!). This is what's in this PR.

3. Refactor the reactor queue bundle to add in a notion of priority, so we can have low and high priority reactX functions in the queue, and have it pick the high priority ones first.

hacdias:
### Source of truth and stale data...

For this I think we could possibly re-fetch the data every minute. It's not that costly.

### Error handling and recording the active actions in state

Yes, I'll work on this. 😄 

### Reacting vs Doing... user initiated changes vs background tasks.

I believe the second or third options would be better. Altough I'm not sure how would we replace [create reactor bundle](https://github.com/HenrikJoreteg/redux-bundler/blob/master/src/bundles/create-reactor-bundle.js) easily.

hacdias:
I'm going with the 2nd.

hacdias:
image

An error that doesn't crash React.

src/files/FilesPage.js
```diff
@@ -51,72 +35,35 @@ class FilesPage extends React.Component {
...
-    const {filename, path} = this.state.rename
-    this.resetState('rename')
-
-    if (newName !== '' && newName !== filename) {
-      this.state.toggleOne(filename, false)
-      await this.props.doFilesMove(path, path.replace(filename, newName))
-      this.state.toggleOne(newName, true)
+  updateFiles = async () => {
+    if (this.props.ipfsReady) {
+      const path = decodeURIComponent(this.props.routeInfo.params.path)
```
olizilla: Let's move this to the redux bundle as a selector like `selectRoutePath`. I needed to do something similar in the ipld explorer: https://github.com/ipfs-shipyard/ipld-explorer/blob/76c6172899bda3fdd1b59be5684ced25d95ff371/src/bundles/explore.js#L29-L36

src/files/FilesPage.js
```diff
@@ -51,72 +35,35 @@ class FilesPage extends React.Component {
...
-      delete: {
-        isOpen: true,
-        files: filesCount,
-        folders: foldersCount,
-        paths: files.map(f => f.path)
-      }
-    })
+  componentDidMount () {
+    this.updateFiles()
+    setInterval(this.updateFiles, ms.minutes(1))
```
olizilla: Checking the files in the redux store are still up to date is a job for the bundle.

src/files/FilesPage.js
```diff
@@ -142,13 +89,12 @@ class FilesPage extends React.Component {
       root = files.path
     }
 
-    const updater = this.makeUpdater('addProgress')
-    await doFilesWrite(root, raw, updater)
+    await doFilesWrite(root, raw)
```
olizilla: Why `await` here?

src/files/FilesPage.js
```diff
@@ -142,13 +89,12 @@ class FilesPage extends React.Component {
...
   }
 
-  addByPath = (path) => {
+  addByPath = async (path) => {
     const { doFilesAddPath, files } = this.props
-    doFilesAddPath(files.path, path)
+    await doFilesAddPath(files.path, path)
```
olizilla: `await` but not using the return value, or catching errors. We shouldn't rely on the return values from redux actions... actions update the store and we render that. I should be fine to remove the `await`

src/files/download-file.js
```diff
@@ -25,6 +25,7 @@ export default (srcUrl, filename, updater) => {
     a.click()
 
     window.URL.revokeObjectURL(url)
+    updater(null)
```
olizilla: Can we rename this `progressCallback`.

src/files/file-input/FileInput.js
```diff
@@ -65,6 +66,15 @@ export default class FileInput extends React.Component {
     })
   }
 
+  componentDidUpdate (prev) {
+    if (this.props.addProgress === 100 && prev.addProgress !== 100) {
+      this.setState({ force100: true })
```
olizilla: Can you talk me through the issue that means `force100` is needed? Could we just drop the last call to `updater(null)` from the downloader and avoid the need for this?

src/lib/dom.js
```diff
@@ -0,0 +1,20 @@
+export function waitForElement (selector) {
```
olizilla: Is this used? Do we need it?

src/bundles/index.js
```diff
@@ -25,7 +25,7 @@ export default composeBundles(
   peerLocationsBundle(),
   routesBundle,
   redirectsBundle,
-  filesBundle,
+  filesBundle({ navbar: '#navbar' }),
```
olizilla: I'd like to find a css solution to this. Coupling the files bundle to the dom structure is a code smell.

hacdias: Yes, but for now I can't see any other way. If you have an idea, ping me.

src/files/FilesPage.js
hacdias: Will do!

src/files/FilesPage.js
hacdias: Sure 😄

src/files/FilesPage.js
hacdias: No reason, will remove.

src/files/FilesPage.js
hacdias: Yes!

src/files/file-input/FileInput.js
hacdias: We use `force100` so the transition from Finished to Add Files is not immediate and takes 2 seconds. The same for downloads. We don't want to change Finished to Download immediately or the user won't even see it.

src/lib/dom.js
hacdias: Here: https://github.com/ipfs-shipyard/ipfs-webui/blob/b25792935a3d34f0f2c3ee29310d72c44df5105d/src/bundles/files.js#L55

src/bundles/index.js
olizilla: When I'm faced with a problem like this, I create a minimal dom and css example on jsbin and try to figure out what is the simplest thing that could work with just css for layout, and js for the actions.

Failing that, redux is the single source of truth for the application state. If we have to do layout in js and we have to know the width of the navbar, then it should be pulled up into the redux store and updated and pulled from there. We could keep it just in React if you think that's prefeable, but it should till be capured as state and pulled up high enough in the tree, that it can be passed into the components that need it. 

Just keep repeating "data flows one way" and "UI as a function of state" until you either go mad or reach a zen like enlightenment 🎐

olizilla: @hacdias can we remove the ({ navbar: '#navbar'}) param from the filesBundle function call now?

hacdias: Done!

olizilla:
UX issue not related to this PR, it's really easy to click on a CID in the file list by mistake and then end up in the IPLD explorer. I think we should have both the filename and the CID do the same thing and drill in to the files.

olizilla:
When I try and drag and drop items that are already in the file list, the item being dragged appears about 200px to the right and 200px below the cursor. Do you see that on yours too or is a chrome+osx thing?

olizilla:
When i try and drop files into the web ui from the file browser, i can't easily drop items on the current directory if it has lots of folders in it. I either have to drop it on the file list header, which is a narrow target, or I have scroll down past the directories to some files, and the drop indication UI fades out file items so it feel like dropping a file on a file won't work, but it does what I'd expect which is add the file to the current folder.

hacdias:
> UX issue not related to this PR, it's really easy to click on a CID in the file list by mistake and then end up in the IPLD explorer. I think we should have both the filename and the CID do the same thing and drill in to the files.

Sure.

> When I try and drag and drop items that are already in the file list, the item being dragged appears about 200px to the right and 200px below the cursor. Do you see that on yours too or is a chrome+osx thing?

Yes, thanks for noticing this. I was wondering where was the dragged item and now I see it. Initially it worked, but then it started just disappearing. I'll take a look and see what the issue is.

---

I think both of these issues could be solved on a different PR though. They're not related to the bundle.

PR #741, opened 2018-07-30
💄: more file browser tweaks

hacdias:
With the previous folder visible as '..' we can drop files there too so it's easier to move files around.

localhost_3000_ 1


hacdias:
localhost_3000_ 1

A bit less off.

hacdias:
Show checkboxes on hover on bigger screens (> 60em):

localhost_3000_ 3

Always show on smaller screens:

localhost_3000_ 4

hacdias:
You can now drop files to 'Add to IPFS' button so you've got a little more dropping area haha

PR #753, opened 2018-08-17
feat: arrow navigation

hacdias:
This introduces the following shortcuts:

- ESC - Unselect everything.
- F2 - Rename selected.
- Del - Delete selected.
- Enter or Meta+Arrow Right - Open selected.
- Arrow up/down - Focus files.

@olizilla I see an issue here I can't find a way to solve right one: if we have a modal opened, we don't want this shortcuts to trigger. I'm just unsure of how to know if a modal is opened. There are some that are completely independent and I would need to make their components fully controlled which would require a lot of changes. There must be a simpler way to handle this.

### DIFF
--- a/src/components/overlay/Overlay.js
+++ b/src/components/overlay/Overlay.js
@@ -3,14 +3,23 @@ import PropTypes from 'prop-types'
 import { Modal } from 'react-overlays'
 
 function Overlay ({children, show, onLeave, className, ...props}) {
+  const stopPropagation = (e) => {
+    e.stopPropagation()
+    e.nativeEvent.stopImmediatePropagation()
+
+    if (e.key === 'Escape') {
+      onLeave()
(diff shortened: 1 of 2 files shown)


olizilla:
@hacdias the modal component can catch all keyboard events that bubble up to it, and call `evt.stopPropagation()` on it to stop it triggering something on the parent page. 

Ideally the ESC key would trigger closing the modal.

hacdias:
Yeah, the ESC key closes the modal already. I'll add the keyboard events and stop the propagation, yes.

hacdias:
@olizilla done and it seems to be working. I added the `stopImmediatePropagation()` on `nativeEvent` because React delegates everything to the root element hence everything has bubbled to the top. Source: https://stackoverflow.com/a/24421834/2681565

PR #759, opened 2018-08-21
feat: files focus mode

hacdias:
**Focused:**

image

**Some selected and a focused:**

image

**Selected and focused:**

image

Now you can also focus the upper directory so it's easier to navigate.

License: MIT

### DIFF
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -22,9 +22,11 @@ export const sorts = {
 }
 
 function compare (a, b, asc) {
-  if (a.toLowerCase() > b.toLowerCase()) {
+  const strings = typeof a === 'string' && typeof b === 'string'
+
+  if (strings ? a.toLowerCase() > b.toLowerCase() : a > b) {
     return asc ? 1 : -1
-  } else if (a < b) {
+  } else if (strings ? a.toLowerCase() < b.toLowerCase() : a < b) {
     return asc ? -1 : 1
   } else {
     return 0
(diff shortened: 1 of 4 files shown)


olizilla:
@hacdias This looks good and works really well. One fiddly issue is that we should be updating the dom element focus with [el.focus()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus) which means keeping a ref around. 

I think it should be the checkbox that gets has focus() called on it, as that is the interactive element. The checkbox should have a focused outline applied, while the containing file component can be styled up as you have it to show clearly the association between the checkbox and the file.

hacdias:
> @hacdias This looks good and works really well. One fiddly issue is that we should be updating the dom element focus with el.focus() which means keeping a ref around.

We already keep refs for files.

---

@olizilla could you take a look at 76f9544?

olizilla:
It blows up if you arrow up past the .. entry. The domElement ref is null at that point

olizilla:
@hacdias also I'm seeing the file order change and the focus skip some items as I press the down arrow, see:

files-reorder-on-keyboard-nav

olizilla:
i can't recreate that sorting / jumping issue now. gonna put it down to some dev setup madness and just keep an eye out for it.

olizilla:
@hacdias there is minor issue when selecting the last item in the list covers it up with the actions bar with using the arrow keys to navigate all the way to the bottom of a list that has more items than fits in the available screen height.

A quick fix is to add a margin-bottom of 80px to the `FilesList` element.

Issue #748, opened 2018-08-16
File browser polishing


olizilla:
- [x] Ignore common system files when adding directories. Specifically `.DS_Store` files on osx. If you know of an similar windows os metadata files that should generally be ignored, then those too. (56106551425b72e1af05c6f18172c6fc7d0e3469)

- ~~Sort by name should only sort by name. We can sort by kind (folders first then files) by default, but if the user selects sort by name it should do that regardless of their kind. We may need to introduce a "kind" column so that the user can return the sort to "by kind" if they want, but I'd love to find a less redundant way.~~ Skip for now. See https://github.com/ipfs-shipyard/ipfs-webui/issues/748#issuecomment-413955567

- [x] fix breadcrumb position jumping up a few pixels when navigating to a file preview page. (acdd1dd438ab5e1968b5fc7d34b08c4dcc6ad43d)

- [x] Display cumulative sizes for directories (9218fea291359f08066f59a6177fbd96185be6a4)

- [x] focus the first item in the file list and add a focus state. (95ea3c7 - just click arrow up or down when nothing is focused).

- [x] add keyboard navigation for up and down arrow keys to navigate up and down the list, and enter to drill down into a folder or file. Meta key + down should also drill down. (95ea3c7)

- [x] replace "Root" in breadcrumb with a home icon (like https://icon.now.sh/home) (39b5de0)

- [x] Add a quick slide-in from bottom animation to the multi select actions bar when it appears (e95d95fdb81d11f00343f9fa41ef42da4b5b74c8)

- [x] When previewing images and videos, don't show the media larger than it's natural resolution. Use `max-width:100%`. Right now they always take up the full width, which looks neat, but small images get super distorted, and most images and videos end up much taller than the available screen height, on my laptop screen at least. (333eb883567de1e0cebd58ea82abe706cf53dc83)

- [x] match breadcrumb underline color to the dag-pb color. See: https://github.com/ipfs-shipyard/ipld-explorer/issues/27 (81e58024e12fe70902d21d60ea8ec263c128d94c)

olizilla:
@hacdias let me know if any of this isn't doable or you have other ideas.

hacdias:
> Ignore common system files when adding directories. Specifically .DS_Store files on osx. If you know of an similar windows os metadata files that should generally be ignored, then those too.

And what if the user wants to upload them even if it's unusual? This is easy to do though: just have a list and filter by name.

[...]

PR #761, opened 2018-08-21
More files polishing

hacdias:
From @olizilla (https://github.com/ipfs-shipyard/ipfs-webui/issues/748#issuecomment-414671819)

> Getting really good now! A couple of things from using the new fancy files:
> 
> - Now we've tightened up the file list, it feels like there is a little too much space at the top of the page. I don't have specific suggestions there, but what do you think?
> - The home icon with the underline seems a little to "strong" when we're at the root. It's all on it's own on the left and really dark and underlined, but it's not interesting until we start drilling down into the folders and files. Perhaps we should hide it or make it less visually distracting when we're at the root.
> - There are a few little gaps in the clickable areas on the file items. I keep managing to click right between the filename and the cid, and it's not part of the hit area so nothing happens. The is another between the icon and the file name/cid. Would be nicer to have that whole part of the box be clickable.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7117,11 +7117,13 @@
         },
         "balanced-match": {
           "version": "1.0.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
@@ -7134,15 +7136,18 @@
         },
         "code-point-at": {
           "version": "1.1.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "concat-map": {
           "version": "0.0.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "console-control-strings": {
           "version": "1.1.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "core-util-is": {
           "version": "1.0.2",
@@ -7245,7 +7250,8 @@
         },
         "inherits": {
           "version": "2.0.3",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "ini": {
           "version": "1.3.5",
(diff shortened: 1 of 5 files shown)


hacdias:
@olizilla yes, there's a lot of space in the top of the page. I noticed we had two margins stacking and removed one of them.

Already fixed the clicking area.

About the home icon: I agree and personally I find the whole breadcrumbs thing too strong. Maybe a lighter color, what do you think?

olizilla:
S C R E E N S H O T S
or
it
didn't
happen.

hacdias:
@olizilla 😆 

image

hacdias:
I personally liked it more saying 'Root' or 'Home' instead of having an icon

olizilla:
@hacdias add screenshots of proposals too! Seeing the icon version side-by-side with the text version will help us decide which works best.

hacdias:
image

image

image

image

olizilla:
@hacdias I don't feel strongly about the icon vs the word, you pick. 

Thinking out loud, is it odd to make the last section of the breadcrumb stand out while the initial sections are faded? the initial sections are the clickable ones, so I think they should look more "clickable" than the last breadcrumb, which should look less clickable, as it's the current location. What do you think @hacdias

hacdias:
@olizilla maybe let's stick with 'Home' for now.

Yesss, the color between faded and not faded has too much contrast. Let me see

hacdias:
@olizilla I made the links behave and look like they do on IPLD Explorer. Every path part is a link. If you click on the path you're in, it will just do a forced refresh. It might be useful sometimes since it just updates every minute.

image

olizilla:
@hacdias nice! Using the word Home + the colour change looks good! The breadcrumb position still seems a little off. It'll look neater if the left edge lines up with the Files colunm header, and it should match the ipld explorer. I know this deviates from the original layouts, but now we can quickly jump between the two pages it'll looks sharp if they line up across the pages too

breadcrumb-position

hacdias:
@olizilla before that, I'm wondering whether of not is this supposed to be aligned:

image

olizilla:
@hacdias ok, let's get this merged and I'll have a look at making making those bits more consistent.

PR #763, opened 2018-08-24
feat: change can't preview message and add loading animation

hacdias:
image

License: MIT


src/files/file-preview/FilePreview.js
```diff
@@ -52,8 +51,8 @@ class FilesPreview extends React.Component {
         return <img className={className} alt={stats.name} src={src} />
       default:
         const cantPreview = (
-          <div>
-            Sorry, we can not preview this file..
+          <div class='b'>
```
fsdiogo:
```js
<div className='b'>
```

hacdias: uuuh, thanks. Good catch 😄

PR #766, opened 2018-08-24
feat: node stats page

olizilla:
Adds a node stats bundle.

License: MIT


hacdias:
image


What about adding a copy button for the ID and Public Keys and some interesting graphs on the right side?

olizilla:
Yep, that sounds good. 

I'm not convinced we need to show the public key in that box. Maybe there is an advanced info section that is collapsed, that has public key and the addresses the node is bound to in.

I think peer id, connected peer count, and the agent version, which we can just call "version" info is worth showing up top.

hacdias:
Current state:

localhost_3000_

And it is responsive!

hacdias:
@olizilla more screenshots:

image

image

hacdias:
@olizilla couldn't find a component but I found a very simple plugin: https://github.com/rstacruz/details-polyfill

I will make a Reacty component though.

hacdias:
@olizilla the tests started failing really randomly. Need to figure out why.

DO you agree with the three graphs on Invision?
- Speed In
- Speed Out
- Storage

I can't see to which compare the speed to to fill the graphs.
The storage by default is 10 GB maximum which will create a graph where you have barely nothing filled.

olizilla:
@hacdias good point. Like a speedometer, we can pick a max value, and if it's over the "top speed" we can just show the bar as full. We could just say that the full is 100MB/s. 

If we want it to be more visually exciting, then we could use a non-linear scale, that gives more space to lower numbers and bunches up higher numbers like http://www.speedtest.net/ does. I'm not sure we want to play those games though.

I do think it's a good thing to have the storage graph look empty if there is very little in your repo!

olizilla:
How about adding `f7 lh-copy pa2 bg-light-gray` to the two advanced blocks and removing `monospace` and adding `blue` to the details summary link?

screenshot 2018-08-30 13 02 46

hacdias:
@olizilla done! 😉

olizilla:
I get duplicate react key errors, as it seems I have the same address twice in my list!

screenshot 2018-08-30 15 09 16

@hacdias this is clearly weird, but we should make the UI robust to that. Can you filter the list to contain only unique entries. I think they should be in monospace too, but they are showing up as sans-serif now. Same for the public key.

In another PR it'd be great to group them by ip4 and ip6, and pick out the ip address section in a different color, so users can quickly pick out what network interfaces their node is bound to.

hacdias:
@olizilla just made some updates. If I recall correctly, you told me you were making some changes this afternoon? Did you commit them?

PR #767, opened 2018-08-24
[wip] dial down the redux actions

olizilla:
Only fire peer-locations events on the peer page.
Upgrade explore bundle to fix react loop on home page.

License: MIT


hacdias:
@olizilla is this PR still valid as we're now using ipfs-explore-components?

olizilla:
Nerp, this can go.

Issue #762, opened 2018-08-24
Upload Large File bug


gitGksgk:
in components\files\action-bar.js    OnFilesChange:  It reads the entire file first and then do things. This cause a huge memory consumption when uploading large file. In Chrome it will crash the tab and in Firefox it almost causes my system to crash..... Any suggestions? thanks.

gitGksgk:
weired thing is, the online ipfs release using the same version doesn't have crash problem. although still consume a lot of memory.

gitGksgk:
I test the problem using 40MB and 700MB file, 40 works fine and 700 crashs. it might differ because of the amount of total memory

hacdias:
Hello @gitGksgk! The current version of WebUI is going to be replaced soon. Could you try and see if the error persists on [revamp branch](https://github.com/ipfs-shipyard/ipfs-webui/tree/revamp)? We know there are some issues with large files.

gitGksgk:
it turns out that it's probably because of some kind of memory leak. In chrome dev tools a lot of 1MB JSArrayBufferData is observed when uploading.  The last sentence traceable is In node_modules\pull-file-reader\index.js . It seems to be designed sending while reading the uploading file but the buffer persists so that the memory consumption becomes larger and larger until it crashes.
 Any ideas?

hacdias:
@gitGksgk I was searching and this seems to be a known issue. Not sure if it's the same one, but I'll take a look at WebUI's code to see if something might be missing. Although, the most probable thing is that this is the same issue as https://github.com/ipfs/js-ipfs-api/issues/654.

gitGksgk:
hi @hacdias Thanks for your reply. yeah I've reached that issue before. I don't know if this project use some part of js-ipfs-api component implicitly, but I can't find http-api in this project, which that issue has modified to fix it's uploading problem.

hacdias:
@gitGksgk we're using `ipfs-react-bundle`, which required `window.ipfs-fallback`. So, we use `window.ipfs` if available or connect through `js-ipfs-api` otherwise. On either way, we use the js-ipfs-api because that's what window.ipfs also uses.

gitGksgk:
hi @hacdias Thanks for your reply. Further investigation shows that in src\bundles\files.js , function
  const res = await ipfs.add(file.content, {
          pin: false,
          // eslint-disable-next-line
          progress: (bytes) => {
            sent = sent + bytes - alreadySent
            alreadySent = bytes
[...]

PR #769, opened 2018-08-29
refactor: simplify file adding

hacdias:
@lidel with your comment yesterday (https://github.com/ipfs-shipyard/ipfs-webui/issues/762#issuecomment-416643973) I ended up changing the code and simplifying it a bit.

Although, it happened once or twice to upload like 10 streams and it would only return 1 hash which I think it might be related to https://github.com/ipfs/js-ipfs-api/issues/797!

It also happened before, but since it was only one by one, they wouldn't all fail.

License: MIT

### DIFF
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -240,6 +240,11 @@ export default (opts = {}) => {
         progress: updateProgress
       })
 
+      if (res.length !== streams.length + 1) {
+        // See https://github.com/ipfs/js-ipfs-api/issues/797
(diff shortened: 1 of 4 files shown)


hacdias:
Fact is: the same set of files errors when using go-ipfs, but it always works on js-ipfs.

hacdias:
Also, adding one by one would avoid errors like this:

image

hacdias:
@lidel thanks! Just added it here!

PR #770, opened 2018-08-29
feat: suggest download when preview not possible

lidel:
If something goes wrong, we should try providing the next action user can pick.
In case of being unable to see the preview, people will appreciate an option of direct download without the need for going back to file list.

This PR adds a plain text link, but the UI can be improved in separate PRs:

screenshot

### DIFF
--- a/src/files/file-preview/FilePreview.js
+++ b/src/files/file-preview/FilePreview.js
@@ -37,8 +37,8 @@ class FilesPreview extends React.Component {
       case 'pdf':
         return (
           <object width='100%' height='500px' data={src} type='application/pdf'>
-            If you're seeing this, is because your browser doesn't support previwing
-            PDF files.
+            Your browser does not support PDFs. Please download the PDF to view it:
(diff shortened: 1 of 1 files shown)


src/files/file-preview/FilePreview.js
```diff
@@ -51,8 +51,9 @@ class FilesPreview extends React.Component {
         return <img className={className} alt={stats.name} src={src} />
       default:
         const cantPreview = (
-          <div className='b'>
-            Sorry, this file can't be previewed <span role='img' aria-label='sad'>😢</span>
+          <div>
+            <p className='b'>Sorry, this file can't be previewed <span role='img' aria-label='sad'>😢</span></p>
+            <p>Try <a href={src} download target='_blank' className='underline-hover navy-muted' >downloading</a> it instead.</p>
```
lidel: Note: `download` won't work for the most of time because `{src}` points at a different Origin, so I've added `target='_blank'` as an universal fallback.

hacdias:
@lidel thanks! I just changed the link classes to match the others around WebUI 😄 Will merge

PR #771, opened 2018-08-29
chore: reuse ipld-explorer as git submodule

olizilla:
This is the shortest path to re-using the ipld explorer code
in webui. Publishing ipld explorer as a reuseable module needs
more thought, and right now we want to be able to build and
edit all of the webui code from a single code base, while we
do the polishing for release.

To test it out, checkout `ipld-explore-as-git-submodule` and do the submodule dance

```
git checkout ipld-explore-as-git-submodule 
git submodule init
git submodule update
```
You only need to do that the once.

If you then need to edit ipld explorer code for the web ui project, you can `cd` into the explore directory, and you are now working in the scope of the ipld-explore repo. You'd make your changes there, as if you were working directly against the ipld-explorer repo (you are) and branch and push from inside the `explore` sub dir. Remeber that your changes need to make sense for both the standalone app and the webui. Once your happy with and you've pushed your ipld explorer changes, you can `cd` back up to the webui root, and run `git status` and you'll see something like

```console
$ git status
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   src/explore (new commits)
```

Which tells us that the ipld explore submodule in your local workspace has your new changes, but the submodule pointer from the parent webui project needs updating. If you add, commit and PR that change to webui, then you share that "update ipld explorer to this commit" with the team.

I already hate this, so will be working on a better solution asap.

In the future ipld explorer may become one app in a learna repo
for all the webui sections, or we might focus on a post-message
api to co-ordinate ipfs apps via iframes with webui orchestrating
things and providing the ipfs instance to child frames.

License: MIT


hacdias:
I believe we can close this @olizilla

olizilla:
Better solution in #781

Issue #675, opened 2018-06-06
Add support for multiple languages since the beginning


hacdias:
We should add a library to support multiple languages right from the beginning to make it easier to expand after this revamp is finished. What do you think?

/cc @olizilla @lidel @alanshaw

lidel:
Keep in mind that setting up a unified crowdsourcing website for all GUI apps should be done at the end of revamp endeavor to decrease noise/breaking changes until we have all screens implemented.

Until then, there are things we should do first:
- [ ] research and pick i18n library that uses **future-proof format**
  - the most popular seems to be [yahoo/react-intl](https://github.com/yahoo/react-intl) and they use ICU format:
     - http://userguide.icu-project.org/formatparse/messages
     - http://icu-project.org/apiref/icu4j/com/ibm/icu/text/MessageFormat.html
     - https://blog.crowdin.com/2016/11/09/icu-syntax-in-crowdin
  - are there better alternatives?
- [ ] create English translation by moving existing text from app to locale files
  - from this point forward all new labels will be added to English locale

Digression: The key is to respect time of translators and do not ask from contributions until we are sure locales work as expected and _won't change drastically_. For example, when initial revamp work is done, we could smoke-test translation process by adding Polish and Portuguese locales to iron out all kinks before opening project to accepting external translations via crowdsourcing site.

lidel:
I think we've done it all :sparkles: :rocket:

We now have a fancy crowdsourcing website for translations (context: https://github.com/ipfs/ipfs-gui/issues/50):
- https://www.transifex.com/ipfs/public/

PR #772, opened 2018-09-01
feat: i18n

hacdias:
Ref.: #675 

This doesn't add translations for Explore view since it will be used as an external component. /cc @olizilla


hacdias:
So, I've started using react-i18next for this WebUI and I'm using [namespaces](https://www.i18next.com/principles/namespaces) to divide the translations into files so each page has its own translation file. It might be useful if we subdivide each part into a different web app.

@ipfs-shipyard/gui could you take a look before I keep going? Do you agree with this approach?

hacdias:
Well, the biggest file would be `files.json` and some, like peers and status would have barely ten lines. Maybe it's actually better to merge them. Opinions?

fsdiogo:
I'd say to keep it in separate files anyway. In the future it could explode translation-wise and that way we'll be ready 🙂

hacdias:
@fsdiogo yup, that could also happen! I'll keep them separate then 😄 It will also be easier to split the Web UI into multiple smaller web apps.

hacdias:
I think this is basically done. Doesn't seem to be any string left.

lidel:
I think the layout is good: all namespaces are kept in `public/locales/en` which works well as a source for PR-based translations, and namespacing makes it easier for people to submit translations via PRs.

There is something to be aware when it comes to `i18next`: IIUC the project works out-of-the-box only with  their own translation website: https://github.com/i18next/react-i18next/issues/472#issuecomment-404186104.

I was unable to find native support for `i18next` translation format at Transifex (which I am leaning towards, [here is why](https://github.com/ipfs/ipfs-gui/issues/50#issuecomment-418106712)). People were successful with using Transifex+[i18next-gettext-converter](https://github.com/i18next/i18next-gettext-converter) which has a  support for [plurals](https://github.com/i18next/i18next-gettext-converter/blob/e587b5fd52ed9af4191cc0b6ebdcd566bb4c743c/src/lib/plurals.js) for most of languages, but there are edge cases with [known issues](https://github.com/i18next/i18next-gettext-converter/issues/66) which may produce frustrating bugs.


The [react-intl](https://github.com/yahoo/react-intl) uses industry standard [ICU format](http://userguide.icu-project.org/formatparse/messages) which does not require additional conversion step and is supported out of the box by Transifex and other  crowdsourcing websites.

hacdias:
@lidel @olizilla should I start working on moving this to ICU format already?

olizilla:
@hacdias what are your thoughts on @lidel's question:

>  did i18next made a big improvement over react-int for you as a dev?
If we consolidate translations under GUI of Transifex, then the "ugliness" of ICU format may not be that big of a deal – translators will not use it directly, and use of industry standard may save us some headache in the long run.

hacdias:
@lidel @olizilla just moved the formats to ICU. It was actually a simple migration since we didn't use much plurals. Take a look at [the locales directory](https://github.com/ipfs-shipyard/ipfs-webui/tree/i18n/public/locales/en) and see how they are.

I'm up to merging this PR. Feedback?

**EXPLORE is not translated.**

I'd like to do a PT translation after this is merged.

lidel:
Thank you @hacdias !  It is :ok_hand:  and I managed to set up Transifex as well:  added the configuration file for Transifex  (`.tx/config`) and uploaded existing source files to the website: https://www.transifex.com/ipfs/ipfs-webui

If you want we can smoke-test translation process before we merge this.:

- Start Translating 
  1. [Create Transifex account](https://www.transifex.com/signup/?join_project=ipfs-webui) 
  1. Go to https://www.transifex.com/ipfs/ipfs-webui/translate/#pt  and try translating some strings for PT locale 
       -  ping me if you don't have access to the team 
 - Sync  Translations
    1. Install and set up command-line client (`tx`) – see docs below 
    1. To download new translations from Transifex: `tx pull -a` 
         - this should create/update files in `public/locales/*` that need to be commited
         - if a new language is created, remember to add it to `src/i18n.js`
    1. (optional) upload updated source locale (`en`): `tx push -s` 
         - (I already did that for this branch, so you can skip it)
         - This will be automated after `revamp` is moved to `master`, for now it is a manual step

## Transifex 101

- [Installing the Transifex Client](https://docs.transifex.com/client/installing-the-client)
- [Understanding `.tx/config` file](https://docs.transifex.com/client/client-configuration#section-tx-config)
- Manual sync via Transifex Client 
  -  [Using Transifex with GitHub in Your Development Workflow](https://docs.transifex.com/integrations/github)
     - [Syncing a local project to Transifex with the Transifex Client](https://docs.transifex.com/integrations/github#section-using-the-client)

public/locales/en/files.json
```diff
@@ -0,0 +1,62 @@
...
+    "description": "Copy the link below and share it with your friends."
+  },
+  "renameModal": {
+    "titleFile": "Rename File",
+    "titleFolder": "Rename Folder",
+    "descriptionFile": "Choose a new name for this file.",
+    "descriptionFolder": "Choose a new name for this folder."
+  },
+  "deleteModal": {
+    "titleItem": "{count, plural, one {Delete Item} other {Delete {count} Items}}",
```
lidel: FYI I had to tweak the ICU format used for plurals because Transifex supports only a subset of ICU when JSON files are used (I think the full support is in gettext):

- https://docs.transifex.com/formats/json#plurals-support

Anyway, it now renders correctly. Below is an  example in Polish, where  we have more than two plural forms :upside_down_face: , and the UI at Transifex now displays them all: 

> screenshot_18

public/locales/en/files.json
```diff
@@ -0,0 +1,62 @@
+{
+  "title": "Files",
+  "home": "Home",
+  "breadcrumbs": "Breadcrumbs",
+  "finished": "Finished!",
+  "totalSize": "Total size: {size}",
+  "filesSelected": "{count, plural, one {File selected} other {Files selected}}",
```
lidel: I also changed this one, because it supported only 2 plural forms, now it is universal :)

olizilla:
💯 🌟

PR #777, opened 2018-09-05
wip: full screen bandwidth chart

hacdias:
Expand the bandwidth chart to full screen.

License: MIT

### DIFF
--- a/src/status/NodeBandwidthChart.js
+++ b/src/status/NodeBandwidthChart.js
@@ -6,6 +6,9 @@ import filesize from 'filesize'
 import { Title } from './Commons'
 import Box from '../components/box/Box'
 
+import Modal from '../components/modal/Modal'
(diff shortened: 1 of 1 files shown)


hacdias:
@ipfs-shipyard/gui this is the current state:

image

Personally, I don't find it much 'attractive'. It seems top much for so little. What do you think?

fsdiogo:
What about instead of showing it like a modal, just expand it to full width in the dashboard?

hacdias:
@fsdiogo I also thought about that, but then what would we do with the circular graph for peers? It would be strange to have a simple circle filling 100% width of the page too

fsdiogo:
Good question, didn't think about that 😅

hacdias:
I'll close this one for now until we get a better idea of how to handle this. Maybe some user has ideas 😱

PR #780, opened 2018-09-05
💄: node bandwidth graph

### LINKED ISSUE #778: Tweaks on Status page
image

The current status page doesn't yet look like the one from Invision. What's missing:

- [x] Banwidth graph:
	- [x] Background
	- [x] Gradient fullfillment
	- [x] Simpler tooltip
	- [x] Simpler axis
- [ ] Peer Graph:
	- [ ] Display percentage on circle - have some feelings about this: what if the slice is small enough the number doesn't even fit?
	- [ ] FLags on tooltip

hacdias:
image

This closes #778

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,12 +31,6 @@
           "resolved": "https://registry.npmjs.org/jsesc/-/jsesc-2.5.1.tgz",
           "integrity": "sha1-5CGiqOINawgZ3yiQj3glJrlt0f4=",
           "dev": true
-        },
-        "source-map": {
-          "version": "0.5.7",
-          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.7.tgz",
-          "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w=",
-          "dev": true
         }
       }
     },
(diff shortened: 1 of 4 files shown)


hacdias:
Current state:

image

@olizilla there is an issue with this: the tooltip won't go away because [this line](https://github.com/ipfs-shipyard/ipfs-webui/pull/780/files#diff-d6319b33d86e4c6a2573c8eb0ac9d901R131) is never `true`. If I don't [do this](https://github.com/ipfs-shipyard/ipfs-webui/pull/780/files#diff-d6319b33d86e4c6a2573c8eb0ac9d901R131), it will be `true` when I get my mouse off the point. I think it might be because the setState triggers a re-render and then the canvas is changed and the 'hover' state over the point is lost so it won't call that function again with `opacity` 0. What do you think?

hacdias:
image

olizilla:
@hacdias did you see https://reactjs.org/docs/portals.html

hacdias:
This closes #778.

olizilla:
Just a minor tweak to have fewer ticks on the y-axis.
screenshot 2018-09-12 13 48 19

PR #786, opened 2018-09-08
Rename Locale for Chinese (China) to zh-CN 

lidel:
Received some useful hints from our friends at Transifex and I think this approach is the  way to go forward.

This PR applies revised naming convention from https://github.com/ipfs/ipfs-gui/issues/50#issuecomment-419430520 
- Renames Chinese locale from `zh` to `zh-CN`,  which is supported by `i18next` and match value from `Accept-Language`:
     > screenshot

I smoke tested this PR and strings that are already translated rendered correctly:
>  screenshot_5

ps. In the future we can add `pt-BR` and  `en-GB` using the same convention :)


lidel:
This is basically https://github.com/ipfs-shipyard/ipfs-webui/pull/784 but with changed code, so safe to merge.

PR #793, opened 2018-09-13
feat: peers page polish

### LINKED ISSUE #790: PeerIDs are force-uppercased
http://localhost:3000/#/peers: 
> screenshot_30

We probably don't want to change case of PeerID, those are in Base58 (case-sensitive) also multiaddr looks weird in uppercase, but that is not a bug, just ux. 

Update: this seems to be just css transform, if i select and copy one, case is correct (big and small letters). Still, bit confusing.

olizilla:
- Fix #790 no more uppercase all the things
- Map color, graticule and point tweaks
- Table colors and typography improvements
- Use the Box so background color matches other sections

screenshot 2018-09-13 13 40 27

License: MIT

### DIFF
--- a/public/locales/en/peers.json
+++ b/public/locales/en/peers.json
@@ -1,7 +1,7 @@
 {
   "title": "Peers",
-  "id": "ID",
+  "peerId": "Peer ID",
   "peers": "Peers",
-  "addresses": "Addresses",
+  "address": "Address",
   "location": "Location"
 }
--- a/src/components/address/Address.js
+++ b/src/components/address/Address.js
@@ -1,0 +1,17 @@
+import React from 'react'
+
+const Address = ({ value }) => {
+  const parts = value.split('/')
+  return (
+    <div className='charcoal-muted pb1 monospace'>
+      {parts.map((chunk, i) => (
+        <span>
+          <span className={i % 2 || i > 4 ? '' : 'charcoal'}>{chunk}</span>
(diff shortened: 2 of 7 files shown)


olizilla:
thanks @lidel I'll take a look. On tweaking the behaviour of the user selection, isn't that a bit surprising for the user? Until everyone is using multiaddrs, it seems like there would be a use-case for being able to just select the ip address from the string.

As an aside, I am sensitive to selection meddling. The web app for my online bank sets all the account numbers and transaction details to `user-select:none` and I have no words for how angry it makes me. This anecdote is more to surface a wtf. We'd never use that power for terrible things.

lidel:
I agree getting the IP instead of entire multiaddr may be a better default.
Looking at that use case,  aiming and selecting entire IP chunk is a chore. 

My point is: right now [clicking selects only a single octet](https://user-images.githubusercontent.com/157609/45561697-a0d0e180-b848-11e8-8601-2d04c6b53a75.png), we could tweak it to [select entire IP](https://user-images.githubusercontent.com/157609/45561773-d249ad00-b848-11e8-8bfe-d8ed56480856.png) instead.
That is probably what everyone wants.

olizilla:
You win. https://github.com/ipfs-shipyard/ipfs-css/pull/26

Also now with unknown locations:
screenshot 2018-09-14 17 10 25

olizilla:
ah rats, gotta i18n them all.

PR #801, opened 2018-09-17
feat: make the peers map fast

olizilla:
- Split out the world rendering from the pins rendering.
- Stop asking ipfs for stats we're not using.
- Ask for stats less often

| Before  | After  |
|-------- |------ |
| peer-map-before | peer-map-after |

License: MIT

### DIFF
--- a/src/bundles/stats.js
+++ b/src/bundles/stats.js
@@ -3,21 +3,17 @@ import { createAsyncResourceBundle, createSelector } from 'redux-bundler'
 const bundle = createAsyncResourceBundle({
   name: 'stats',
   getPromise: async ({ getIpfs }) => {
-    const [bitswap, repo, bw] = await Promise.all([
-      getIpfs().stats.bitswap(),
-      getIpfs().stats.repo(),
+    const [bw] = await Promise.all([
+      // getIpfs().stats.bitswap(),
+      // getIpfs().stats.repo(),
       getIpfs().stats.bw()
     ])
-    return { bitswap, repo, bw }
+    return { bw }
   },
-  staleAfter: 2000,
+  staleAfter: 3000,
   persist: false,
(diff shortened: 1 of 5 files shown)


src/bundles/stats.js
```diff
@@ -3,22 +3,18 @@ import { createAsyncResourceBundle, createSelector } from 'redux-bundler'
 const bundle = createAsyncResourceBundle({
   name: 'stats',
   getPromise: async ({ getIpfs }) => {
-    const [bitswap, repo, bw] = await Promise.all([
-      getIpfs().stats.bitswap(),
-      getIpfs().stats.repo(),
+    const [bw] = await Promise.all([
```
fsdiogo: We should remove `Promise.all` as we're only waiting for one promise now.

olizilla:
I spotted an issue where the map pins are drawn from all the locations rather than the locations for the current peers list.

The quick fix is to do a map of the curernt peers to the their location where available. Longer term we should refactor the locations logic to just look up locs for the current swarm rather than all locations we've ever seen, as we cache that info in the async storage, so we dont need to keep it memory.

src/peers/WorldMap/WorldMap.js
```diff
@@ -9,100 +9,118 @@ import * as topojson from 'topojson'
...
+    .attr('d', path)
+    .attr('fill', 'none')
+    .attr('stroke', 'none')
+
+  return el.node().toReact()
+}
+
+// Just the dots on the map, this gets called a lot.
+const MapPins = connect('selectPeerCoordinates', ({ width, height, path, peerCoordinates }) => {
+  // https://github.com/d3/d3-geo/blob/master/README.md#geoGraticule
```
fsdiogo: Remove the comments below.

PR #799, opened 2018-09-17
[WIP] feat: add api and api provider info

olizilla:
- show the api address when connect via http
- otherwise show api provider (window.ipfs, embedded js-ipfs)

**Connected to external js-ipfs node**
screenshot 2018-09-17 14 40 22

**Connected to window.ipfs running embedded js-ifps**
screenshot 2018-09-17 14 38 03

todo:
- [ ] update with new version of ipfs-redux-bundle with support for `selectIpfsProvider`

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>

PR #802, opened 2018-09-19
feat: add ipfs connection indicator and notifications

olizilla:
This is an initial sketch of an idea on how to inform users about IPFS connection issues. It works for both api errors and window.ipfs errors, but it is very high level; just enough info to let the user know that we're not able to talk to ifps at the moment.

- Add an IPFS connection status indicator to the top right. It's yellow/orange when when we get errors back from the stats api call, which we poll every 3s, so is a useful indicator of connection health.
- Add a dismissible toast style notification to let you know that there is a problem.
- The IPFS request error notification is replaced with a "Normal IPFS service has resumed" if it wasn't dismissed, when the connection comes back, otherwise it just stays hidden.

notify-error

TODO 
- [x]  listen for errors from files tab.
- [x] this is branched off of the work in #799 so that needs merging first, which in turn needs https://github.com/ipfs-shipyard/ipfs-redux-bundle/pull/8


hacdias:
@olizilla oops: found a bug: when going to /welcome it doesn't show the API address in input box. It might be an issue on ipfs-redux-bundle. Will take a look.

hacdias:
@olizilla I think that issue is fixed by https://github.com/ipfs-shipyard/ipfs-redux-bundle/pull/12

olizilla:
Live action speed play here https://ipfs.io/ipfs/QmevJYiF5yFGxz68kHWwyT62sEDVtTJnn8TssqD3XGEYFM/ipfs-webui-connection-indicator-sneak-peak.mp4

olizilla:
I'd like to merge this to revamp and then move revamp to master 🚢💨 
@lidel @alanshaw @fsdiogo pls take a look

src/bundles/notify.js
```diff
@@ -0,0 +1,93 @@
+import { createSelector } from 'redux-bundler'
+
+/*
+# Notify
+- show error when ipfs goes away.
+- show ok when it comes back.
+- dismiss the the ok after 2s
```
fsdiogo: dismiss ~the~ the ok after ~2s~ 3s

src/bundles/notify.js
```diff
@@ -0,0 +1,93 @@
...
+
+const notify = {
+  name: 'notify',
+
+  reducer: (state = defaultState, action) => {
+    if (action.type === 'NOTIFY_DISMISSED') {
+      return { ...state, show: false }
+    }
+
+    if (action.type.match(/_FETCH_FAILED$/) || action.type.match(/^FILE_\w+_FAILED$/)) {
```
fsdiogo: The actions are `FILES_*`, so this should be:

```js
action.type.match(/^FILES_\w+_FAILED$/)`
```

src/bundles/notify.js
```diff
@@ -0,0 +1,93 @@
...
+        return {
+          ...state,
+          show: true,
+          error: true,
+          eventId: 'FETCH_FAILED'
+        }
+      }
+    }
+
+    if (action.type.match(/_FETCH_FINISHED$/) || action.type.match(/^FILE_\w+_FINISHED$/)) {
```
fsdiogo: Same here: `FILES_`.

src/components/connected/Connected.js
```diff
@@ -0,0 +1,34 @@
...
+
+const pickColor = (ipfsReady, ipfsConnected) => {
+  if (ipfsReady && ipfsConnected) return 'aqua'
+  if (ipfsReady) return 'yellow'
+  return 'gray'
+}
+
+const pickTitleKey = (ipfsReady, ipfsConnected) => {
+  if (ipfsReady && ipfsConnected) return 'connectedToIpfs'
+  if (ipfsReady) return 'ipfsApiRequestFailed'
```
fsdiogo: Shouldn't we put these keys into some constants file and import them from there?

olizilla: maybeeeee. I'm going to leave that out for now, as in general, different parts of the app should uses different i18n resource file. The crime here is that I'm reusing the same label in two, only mildly related things, when really each should use it's own resource file.

olizilla:
@lidel very good plan! I have made it so. The spinner is gone. The connection indicator is a link. 

For bonus points we now also have auto-retry ipfs init when you get stuck on the welcome page if you open the app with no daemon running. If you go start your daemon, the page will now auto update to show you are now connected, which feels nice.

olizilla:
retry-ipfs-init

olizilla:
Gonna merge this and see if i can do anything about the bundlesize in a separate PR

PR #808, opened 2018-09-24
feat: make peers page first render fast

### LINKED ISSUE #807: d3 generated peer map svg is 2.8MiB
The lovely peer world map on the peers page is created by d3, which means it's (relatively) simple to tweak programatically. That's nice, but the svg it generates is almost 3 MiB. That is large. The cause seems to be the precision that you get in the svg paths from the d3 calculations. Every point is rendered to ~16 decimal places of precision.

This is a known issue https://github.com/d3/d3-path/issues/10

Right now, we're not doing anything dynamic with the world map part of the svg, so we can capture it as a static svg to spare the browser all that hard calculatin` 

with some `svgo` magic, the map can be reduced to ~71KiB (19KiB gzip).

```console
$ svgo StaticMap.svg -p 0

Done in 1173 ms!
2905.256 KiB - 97.5% = 71.235 KiB
```

olizilla:
The peers page was taking >2s to load on my machine. Turns out the
d3 generated svg of the world map was around 3MiB

This PR makes extracts the world map to a static svg that is used as
a background image. It's ~79KiB / ~20KiB gzipped and cached as a
static resource rather than generated by js.

The peers page now loads faster than all the others. There is still
too much fan spinning caused by the geo look ups, but this is an
improvement.

fixes #807 

License: MIT


olizilla:
fast-peers-page-load

fsdiogo:
In big screens the background image repeats:
pr

In master it doesn't:
master

It should behave the same way imo!

src/peers/WorldMap/WorldMap.js
```diff
@@ -4,23 +4,16 @@ import { connect } from 'redux-bundler-react'
...
-import worldData from './world.json'
+import staticMapSrc from './StaticMap.svg'
 
 const WorldMap = ({ t }) => {
   return (
-    <div className='flex w-100 mb4' style={{ 'height': '550px' }}>
-      <AutoSizer>
-        { ({ height, width }) => (
-          <GeoPath width={width} height={height}>
+    <div className='flex w-100 mb4' style={{ 'height': '550px', background: `transparent url(${staticMapSrc}) center` }}>
```
fsdiogo: Adding `no-repeat` fixes the https://github.com/ipfs-shipyard/ipfs-webui/pull/808#issuecomment-424281275 issue:

```js
style={{ 'height': '550px', background: `transparent url(${staticMapSrc}) center no-repeat` }}
```

olizilla:
I'm going to give the table a max width and make it line up with the map

olizilla:
screenshot 2018-09-25 14 53 16

PR #812, opened 2018-09-25
fix: misc ui fixes

fsdiogo:
- Fixed inputs to have the same border-radius
- Fixed api address input focus to behave like the others:

| Before | After |
|--------|-------|
|before|after|

### DIFF
--- a/src/components/text-input-modal/TextInputModal.js
+++ b/src/components/text-input-modal/TextInputModal.js
@@ -108,7 +108,7 @@ class TextInputModal extends React.Component {
             value={this.state.value}
             required
             autoFocus
-            className={`input-reset charcoal ba b--black-20 pa2 mb2 db w-75 center focus-outline ${this.inputClass}`}
+            className={`input-reset charcoal ba b--black-20 br1 pa2 mb2 db w-75 center focus-outline ${this.inputClass}`}
             type='text' />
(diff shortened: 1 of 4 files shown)


fsdiogo:
There is another issue: when loading the page, the settings page only appears when we have access to the config, that takes a glimpse of time but we can see the icon animation changing colors incorrectly. That is because of the [`link` tachyons class](https://github.com/tachyons-css/tachyons-links#the-css), that has an animation of `.15s` for the color. Without that it works and the lack of the animation is unnoticeable, at least to my eyes.

| Before | After |
|--------|-------|
|bef|af|

☝️ 	the quality of the gifs is not great.

Want me to commit this change or it goes in another PR with a nicer solution?

olizilla:
@fsdiogo I think last time it was discussed @lidel and I were thinking that settings tab should to be visible but to appear disabled until we know it can be used, rather than hidden. In general it seems better UX to show what's possible even if it's not currently available. I'm ok with it fading in. Is the problem that the icon fill starts off black and transitions to gray?

fsdiogo:
Agreed, that sounds like a better UX.

Yes, that is the main problem.

fsdiogo:
When settings are not available the link appears disabled and with no pointer-events:

set

Can this be merged?

Issue #815, opened 2018-09-25
How do users discover the Web UI?


olizilla:
It's mentioned in the docs

- https://docs.ipfs.io/introduction/usage/#fancy-web-console

as:

> We also have a web console you can use to check the state of your node. On your favorite web browser, go to:
> http://localhost:5001/webui

but if you don't read the docs that far (or at all) then I think the Web UI is basically hidden from you.

What if we added a log line to go-ipfs and js-ipfs start up like:

```console
$ jsipfs daemon
Initializing daemon...
Swarm listening on /ip4/127.0.0.1/tcp/4003/ws/ipfs/QmNTAZYQ5rtaoFtryAX2h9dycuBjhVgXtjPVZNYuMHMBw8
Swarm listening on /ip4/127.0.0.1/tcp/4002/ipfs/QmNTAZYQ5rtaoFtryAX2h9dycuBjhVgXtjPVZNYuMHMBw8
Swarm listening on /ip4/192.168.1.244/tcp/4002/ipfs/QmNTAZYQ5rtaoFtryAX2h9dycuBjhVgXtjPVZNYuMHMBw8
API is listening on: /ip4/127.0.0.1/tcp/5002
Gateway (readonly) is listening on: /ip4/127.0.0.1/tcp/9090
Web UI available at http://localhost:5002/webui
Daemon is ready
```

meiqimichelle:
If this is straighforward to add, I don't see any reason not to! That info seems to fit with all the other info listed -- aka, here are all the things you now have available to you at these locations.

While we're there, we should standardize the language for all the sentences (colon versus no-colon). So, it could be:

```console
$ jsipfs daemon
Initializing daemon...
Swarm listening on: /ip4/127.0.0.1/tcp/4003/ws/ipfs/QmNTAZYQ5rtaoFtryAX2h9dycuBjhVgXtjPVZNYuMHMBw8
Swarm listening on: /ip4/127.0.0.1/tcp/4002/ipfs/QmNTAZYQ5rtaoFtryAX2h9dycuBjhVgXtjPVZNYuMHMBw8
Swarm listening on: /ip4/192.168.1.244/tcp/4002/ipfs/QmNTAZYQ5rtaoFtryAX2h9dycuBjhVgXtjPVZNYuMHMBw8
API is listening on: /ip4/127.0.0.1/tcp/5002
Gateway (readonly) is listening on: /ip4/127.0.0.1/tcp/9090
[...]

PR #816, opened 2018-09-25
docs: add first draft of usability test script

olizilla:
I'm not sure this is a great usability test _yet_, but it's been productive to think about what the app should make easy vs possible, and just thinking through how you might test for that has uncovered a few significant UX issues like #815 

License: MIT

### DIFF
--- a/docs/usability/ipfs-webui-useabilty-test.md
+++ b/docs/usability/ipfs-webui-useabilty-test.md
@@ -1,0 +1,82 @@
+# IPFS Web UI Usability Testing
+
+The app should make the following actions **easy** to complete
+
+- Add a file and share it with a friend
+- See the files in your repository
+- Determine if the node is connected to the network and operating normally
+- Diagnose when your IPFS daemon isn't running and fix it
+- Find the peer ID of the node you are connected to
+- Find IPLD info about an IPFS path and browse a link
+- Discover and try out experimental IPFS features _missing_ [ipfs-webui#814](https://github.com/ipfs-shipyard/ipfs-webui/issues/814)
+- Find the total size of your IPFS repo _missing_
+
+More holistically, it should give a feeling of being connected to a network of peers.
+It must not give a sense of "uploading to the cloud", as that is a significant cause
+of confusion about how IPFS works, and it's dangerously misleading.
+
+The app ensure that the following actions are **possible** to complete
+
+- Determine the peer ID of the node
+- Determine how it is connected to IPFS
+- Find which network addresses it's bound to
+
+## Goal of the usability test
+
+- Find any problems prevent any of the participants from completing the tasks
+- Gather structured feedback on the v2 redesign
+
+## Methods
+
+- 45 min interview with PL folk, some developers, some not.
+- Notes, no recording
+- Remote or in-person
+- Include a retrospective design crit
+
+## Test Script
+
+- Visit webui.ipfs.io and find the peer ID of your node
+  _Do they have a daemon running already?_
+  _Do they follow the prompts to start their daemon?_
+  _Do they have to go and download go-ipfs first?_
+
+- Add a picture you'd like to share to your IPFS node
+- Copy the share link and send it to me via (whatever works best for both)
+  - _Do they figure out where the add files feature is?_
+  - _Do they manage to upload the file?_
+  - _Do they find and copy the share link?_
+  - _Do they try drag-and-drop? or do they use the add button or..._
+  - _Are you able to get the file from ipfs webui at your end_
+
+- I'll send you a share link. Download the file to your desktop
+  - _Do they find the download link?_
+  - _Do they get the file?_
+
+- Add the file and and explore it the IPLD data behind it. What node type is it?
+
+- Explore the "Project Apollo Archives" IPLD dataset, follow the "albums" link, and find picture to view on the IPFS gateway
+  - _Do they click through the album link?_
+  - _Do they find a jpg?_
+  _ _Do they find the View on IPFS Gateway link?_
+
+- Pick a dataset from https://archives.ipfs.io/ and explore it the IPLD structure behind it.
+  - _Do they? ...this requires copy and pasting the cid from the website into the explore form_
+
+- How many peers is your node connected to?
+  _Do they find out? How do they feel about that?_
+
+
+## Questions / Musings
+
+- How do we test for discoverability? (https://github.com/ipfs-shipyard/ipfs-webui/issues/815)
+  - how does a new user even find the webui? The daemon doesn't log out an address for it on start up.
+  - If we test for that with new users, I think we'd find most wouldn't know where to look.
+  - this is probably just another UX issue, and we need to fix it. I think I'm loathed to test it right now as the UX is so bad it's not worth testing yet.
+
+- How do we check for mis-conceptions the app might dispell or reinforce?
+  - IPFS is not a Backup / upload to the cloud service. It's still up to you to ensure your files are on more than one machine.
+  - Files are not private. It's up to you to encrypt them... but they are obscure, and only those you share the link with, or those that already have them will be able to find them from you.
+
+---
+
+Credit: cribbed from https://github.com/18F/myusa-ux/edit/master/research/usability/sprint22_research-plan.md


olizilla:
@meiqimichelle it's a usabiltity test script! Cribbed from https://github.com/18F/myusa-ux/edit/master/research/usability/sprint22_research-plan.md

I'm keen to get your thoughts / corrections on this. Perhaps we should try it out?

olizilla:
The rendered preview is here https://github.com/ipfs-shipyard/ipfs-webui/blob/d3efcd3c272cec299d820faba70be9e489ffb686/docs/usability/ipfs-webui-useabilty-test.md

docs/usability/ipfs-webui-useabilty-test.md
```diff
@@ -0,0 +1,82 @@
+# IPFS Web UI Usability Testing
+
+The app should make the following actions **easy** to complete
+
+- Add a file and share it with a friend
+- See the files in your repository
```
lidel: "Files"? Do we mean MFS or all objects in IPFS repo?
It would be much easier if we had better language to talk about those things.
MFS is a really bad acronym and even worse full name (civilians just shut down). 

Digression:
Any thoughts on UX change to use "IPFS Drive" instead of "MFS" (https://github.com/ipfs/specs/issues/174#issuecomment-412912829) across the stack, docs etc? It is a so much better name.

docs/usability/ipfs-webui-useabilty-test.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Questions / Musings
+
+- How do we test for discoverability? (https://github.com/ipfs-shipyard/ipfs-webui/issues/815)
+  - how does a new user even find the webui? The daemon doesn't log out an address for it on start up.
+  - If we test for that with new users, I think we'd find most wouldn't know where to look.
+  - this is probably just another UX issue, and we need to fix it. I think I'm loathed to test it right now as the UX is so bad it's not worth testing yet.
+
+- How do we check for mis-conceptions the app might dispell or reinforce?
+  - IPFS is not a Backup / upload to the cloud service. It's still up to you to ensure your files are on more than one machine.
```
lidel: It would be interesting to ask people (1) to describe how long data uploaded via webui or share.ipfs.io will be available [on public gateway], and (2) why they think what they think.

meiqimichelle:
This is great @olizilla and I'm excited this is happening! I am sure we will learn alot a lot.

screenshot 2018-09-27 14 07 56

Ok, let's see -- how to make it even better 💃 

Does your test script (which I quite like!) cover all the things you want to make easy (the list at the top)? It seems like you've focused on a sub-set of the problem space in lines 3 - 22, which isn't a bad thing, but! I think you've got two or three documents all rolled together into this one that each deserve fuller exploration.

You might consider:
- Pulling out the kinda intro section from lines 3 to 22, and the `Questions/musings` section at the bottom, into a background doc, and/or use a version of that as the README to the `usability` folder. It can become something the sets up your problem space for user research and usability testing priorities in general.
- Then, you're free to really focus on the specific features you're testing in this round in this doc. You could scope the intro into the parts of lines 3 to 22 that you think the questions you're asking will inform. This doc becomes a very practical guide to this specific test that you can come back to months from now when you need to remember wth you were trying to do in these specific weeks of testing. Because, memory.

I think the musings you brought up are important, and we should explore those more, and in the meantime you can separate and 🚢 the first, scoped usability test. IMHO I would scope the doc, and then do what you said -- try it out, and see how it goes!

docs/usability/ipfs-webui-useabilty-test.md
meiqimichelle: These are really good areas of understanding that we need to dive into! Yays.

I really like your suggested question (1) @lidel ! We should find a different way of getting at (2) tho -- humans aren't generally very good at accurately describing why they do something or know something because they'll naturally try to give you an answer that makes them look smart, or gives you the answer they think *you* want, because they want to seem likable. Human nature n such. So we shall have to be sneaky.

In any case tho....hmmmmmm........as a first step, I like the idea of adding a version of @lidel 's question to the test script. After they send the link to the interviewer, could ask something like:
- How long do you think that picture will still be available/at that address/at that link?
- _If they answer so that it seems like they understand about the close-the-computer-its-unavailable thing, could ask_ How would you make sure that picture is always available?
- Can anyone find that photo, or just me?

You can likely improve on those questions, but that's a start -- what do you think?

For bigger questions of understanding in general, it might be good to decide what the Mission of the App is (you might know this already, and I just don't know the latest!). Is it meant to be someone's first toe-dip into IPFS-land? Or...? Having this target in mind will help understand what sort of research we need to understand potential misconceptions, or even at the start, what the app should be helping folks understand. ie, I could see a different sort of module or illustration that we should develop, or metric, if one of the key communication points we want to convey is that IPFS is not a backup service (something around time, perhaps, and how that interacts with IPFS). And if we want to convey that 'uploads' are technically available to anyone on the network, I could see some different ways of illustrating that, too.

But maybe those shouldn't be main 'talking points' for the app in the first place. But! It could be really cool to make the app the place people go to understand how IPFS works, soooooo.....hmmmmm... 🤔

docs/usability/ipfs-webui-useabilty-test.md
meiqimichelle: We could at least test that language in the app, if it appears there. Does it appear in the app? It would be great to see first if that makes more sense to people before changing everything across all our docs. (I'm sure it will be clearer than MFS, but I wonder if through testing we'd come up with something even better!)

Would 'IPFS Drive' replace.....repository? No, that's not quite it. Hmm. Do you think we're at a place where we can talk at Team Week about how lots of our concepts/terms fit together? ie, to a Normal Human, what `ipfs add` means, what `ipfs get` means, how that relates to, say, an IPFS Drive.....??? It would be great to start to align on an illustration or metaphor for all this that the team could use since often when you change one word, you find out later that it doesn't make sense once *other* terms are changed. Might be asking too much at this stage of development tho :)

olizilla: @lidel I used files here as I don't think that user will typically know there is a difference between "things in the MFS/Drive" and "things in the repo". I think the name "Drive" is better than "MFS", and I think better still would be for all files in your repo to be in you MFS and we just refer to them as "your files".

@meiqimichelle Strong yes from me about making time in team week to figure this out! We have to find a story and a set of terms that work. Not having it slowing us down.

The biggest ones that are causing me trouble today are
- Files in repo vs Files in MFS / Drive. Right now, the list of files in your MFS are a subset of all the files in your repo. We could lobby for that to not be the case in the future, but in the mean time we have to deal with it.

- Upgrade "pinning" to co-hosting. We need to start pushing for apis that let us build apps that let people co-host content and set a duration (1week, 1month, until i need space, forever, kind of thing.) The concept of pinning doesn't seem to work for everyone, and it's an all or nothing thing; If I pin it, it's forever until I unpin it. If I don't pin it, I share it anyway until my repo is full, at which point it may get deleted.

docs/usability/ipfs-webui-useabilty-test.md
```diff
@@ -1,82 +1,49 @@
...
-- Discover and try out experimental IPFS features _missing_ [ipfs-webui#814](https://github.com/ipfs-shipyard/ipfs-webui/issues/814)
-- Find the total size of your IPFS repo _missing_
+  - _Do they have a daemon running already?_
+  - _Do they follow the prompts to start their daemon?_
+  - _Do they have to go and download go-ipfs first?_
 
-More holistically, it should give a feeling of being connected to a network of peers.
-It must not give a sense of "uploading to the cloud", as that is a significant cause
-of confusion about how IPFS works, and it's dangerously misleading.
+> Add a picture you'd like to share to your IPFS node
```
lidel: What if we change this to (or add it as a separate item):
- Add two files/pictures and share then with a friend

This could produce more data, for example: Are people sending one by one and sending two links, or do they notice that multiple files could be selected and shared together? Are our GUIs communicating the fact multiple files can be bundled and shared together?

We could ask follow-up questions about content-addressing and the meaning of root CID (wrapping directory) etc.  (I suspect having more than one file make it easier to reason about wrapping directory and related concepts)

PR #831, opened 2018-09-26
fix: add quick-config script

fsdiogo:
Adds the missing `quick-config.sh` script as noted in https://github.com/ipfs-shipyard/ipfs-webui#config-your-ipfs-daemon.


olizilla:
See #829

fsdiogo:
Damn, missed that completely 😔

PR #832, opened 2018-09-27
fix: sort folder by size

### LINKED ISSUE #826: Files - Sorting by size not working
screen shot 2018-09-26 at 09 58 17

Not sure what's happening here, looks like the folders are grouped, but they don't appear to be in any particular order...

hacdias:
Closes #826.

@alanshaw can you verify if it fixes the issue on your machine?


olizilla:
Works for me, on OSX, same as Al, so I'm happy for this to go in.

Issue #824, opened 2018-09-26
Should be able to "add by path" with a CID


alanshaw:
I currently have to prefix them with `/ipfs/`, would be good to just be able to paste `QmVaVXbLdw4R5NqAiiQoTWtitxo5g7FS31PQmCLbH9p8Fu`, especially seeing as you can copy a hash via the context menu, and it might be shared to someone who'll then want to add it.

PR #838, opened 2018-09-28
fix: add by path

fsdiogo:
Adds the ability to `add by path` with only the CID. Also works with `/ipfs/` prefixed.

path

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/824.

### DIFF
--- a/src/files/file-input/ByPathModal.js
+++ b/src/files/file-input/ByPathModal.js
@@ -11,7 +11,7 @@ function ByPathModal ({ t, tReady, onCancel, onSubmit, className, ...props }) {
       p = `/ipfs/${p}`
     }
 
-    return isIPFS.ipfsPath(p.trim())
+    return isIPFS.ipfsPath(p)
   }
 
   return (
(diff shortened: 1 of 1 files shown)


src/files/file-input/ByPathModal.js
```diff
@@ -6,9 +6,17 @@ import Icon from '../../icons/StrokeDecentralization'
 import TextInputModal from '../../components/text-input-modal/TextInputModal'
 
 function ByPathModal ({ t, tReady, onCancel, onSubmit, className, ...props }) {
+  const validatePath = (p) => {
+    if (!p.startsWith('/ipfs/')) {
+      p = `/ipfs/${p}`
+    }
+
+    return isIPFS.ipfsPath(p.trim())
```
alanshaw: `trim` before `if` in case I paste "   /ipfs/QMhash"?

fsdiogo: No need, because we have:

```js
onChange={(p) => p.trimStart()}
```

Now that I'm looking at it, we don't even need that trim you pointed out.

alanshaw: Less code FTW!

Issue #817, opened 2018-09-25
Explain how to set up CORS headers on welcome page


olizilla:
We can't differentiate "api not running" from "api running but no CORS headers configured" so we'll have to add CORS header config messaging to the welcome page, as it's likely to be the second problem that users hit if they try and use the webui from https://webui.ipfs.io

olizilla:
Pervious discussion here https://github.com/ipfs-shipyard/ipfs-webui/issues/578

> We could try requesting the empty dir CID from a gateway running on the default port http://127.0.0.1:8080/ipfs/QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn

> That way we can infer if you have a daemon running or not.

We can use that info to focus on 
- no gateway - "download ipfs, set up cors, start the daemon"
- gateway but no api - configure cors, or set the address your api is on

There is the case that we don't detect at gateway because it's on a custom port or disabled, so arguably we should just show all the steps that need to be done to get this working.

PR #839, opened 2018-09-28
chore: add cors info on welcome page

fsdiogo:
cors

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/817.

### DIFF
--- a/src/welcome/WelcomePage.js
+++ b/src/welcome/WelcomePage.js
@@ -31,13 +31,20 @@ const WelcomePage = ({ doUpdateIpfsApiAddress, ipfsInitFailed, ipfsConnected, ip
             ) : (
               <div>
                 <h1 className='montserrat fw2 navy ma0 f3 yellow'>Is your IPFS daemon running?</h1>
+                <p>Failed to connect to the API.</p>
                 <p>
-                  Failed to connect to the API. Please check the IPFS daemon is running.
-                  Run <code className='f6'>ipfs daemon</code> in a terminal window, as shown below.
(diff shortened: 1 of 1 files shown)


olizilla:
This is great! Let's only show the CORS info if the current location isn't `localhost:5001`, then this is good to go.

fsdiogo:
> Let's only show the CORS info if the current location isn't localhost:5001

@olizilla why is that? 🤔

olizilla:
Just an small UX improvement to only show the CORS config info where the origin for the page doesn't match the origin for the api.

olizilla:
How about pulling the CORS info below the api address form? By far the most likely problem is that your daemon isn't running.

src/welcome/WelcomePage.js
```diff
@@ -5,11 +5,51 @@ import Box from '../components/box/Box'
 import Button from '../components/button/Button'
 import ComponentLoader from '../loader/ComponentLoader.js'
 
-const WelcomePage = ({ doUpdateIpfsApiAddress, ipfsInitFailed, ipfsConnected, ipfsReady, ipfsApiAddress }) => {
+const WelcomePage = ({ doUpdateIpfsApiAddress, apiUrl, ipfsInitFailed, ipfsConnected, ipfsReady, ipfsApiAddress }) => {
   if (!ipfsInitFailed && !ipfsReady) {
     return <ComponentLoader pastDelay />
   }
 
+  const renderConnectionStatus = () => {
```
olizilla: Minor but I'd find this easier to read if `renderConnectionStatus` was another component rather than an extra render function, like `<ConnectionStatus connected sameOrigin />`

Issue #842, opened 2018-10-01
Publish GIT Revision of running build


lidel:
Right now we are unable to tell what build is running, which will make bug reporting a real problem.

It would be extremely useful to:
- At least put it in meta header ([example](https://user-images.githubusercontent.com/157609/46286101-25786b00-c57e-11e8-8807-abecc8993861.png))
- Additionally,  put a faded-out version somewhere in Settings screen, or near 'See source'/'Report a bug', so it is visible on screenshots and/or non-technical user is able to check it and provide it with a bug report

This needs to happen before we ask general public for feedback. 

Eg.
- https://stackoverflow.com/questions/48391897/add-git-information-to-create-react-app
- https://github.com/facebook/create-react-app/issues/2486

fsdiogo:
This is great @lidel 👌

Here's a thought, as we have the running daemon version, why not have the WebUI version there too:

<img width="1271" alt="version" src="https://user-images.githubusercontent.com/33324750/46292349-9d4b9300-c588-11e8-87c3-c62daa421455.png">

lidel:
Hm.. I'd rather not put it there: webui version is not really a part of "node info".

What if we add it above links added in https://github.com/ipfs-shipyard/ipfs-webui/pull/809#issuecomment-425911220 ?
To save space, the git hash  (`git rev-parse --short HEAD`) could double as a link to sources at github eg: 
[See sources: `1e2c58fe`](https://github.com/ipfs-shipyard/ipfs-webui/tree/1e2c58fee899009cf68bc4c903d4eed9dbbe1f0b)

fsdiogo:
> webui version is not really a part of "node info".

Yeah, you're right.

> What if we add it above links added in #809 (comment)?

That could work, although personally I wouldn't want that info showing all the time. I'd prefer as you said somewhere in Settings. Or in the status page maybe?

olizilla:
Very cool! What if we go more subtle and replace "see the code" with the revision number?

fsdiogo:
I have mixed feelings about that 🤔

For one I like it because it would have -1 link. On the other hand, it's not very clear for the average user that the revision number will link to the code, and I think we should always have a direct link to the latest code. 

Do you agree?

fsdiogo:
Cool, I'll open a PR.

What will be the plan to update the git revision?

PR #844, opened 2018-10-02
feat: add git revision

fsdiogo:
Adds the git revision.

In a meta tag:

meta

In the UI:

ui

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/842.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,9 @@
   "version": "2.0.0",
   "private": true,
   "scripts": {
-    "start": "react-scripts start",
+    "start": "cross-env REACT_APP_GIT_REV=`git rev-parse --short HEAD` react-scripts start",
     "prebuild": "shx cp config/webpack.config.prod.js node_modules/react-scripts/config/",
-    "build": "react-scripts build",
+    "build": "cross-env REACT_APP_GIT_REV=`git rev-parse --short HEAD` react-scripts build",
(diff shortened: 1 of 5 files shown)


lidel:
@olizilla I'd really like to have revision number present when I ship bundled webui with a new beta companion.
Please review when you have a spare moment :)

olizilla:
minor, but I'm trying to move us towards using consistent names for the app...

- IPFS Web UI - it's human friendly name
- ipfs-webui - it's machine friendly name

as an aside, I went looking for a more structured way to declare the git revision that a page was built from and found

https://github.com/krallin/meta-revision

which is listed on the WHATWG MetaExtensions proposals https://wiki.whatwg.org/wiki/MetaExtensions

So we could do something like

```html
<meta name="generator" content="ipfs-webui https://github.com/ipfs-shipyard/ipfs-webui" />
<meta name="revision" content="[revision]" />
```

It's kinda mind blowing that there is no accepted structured way for a web page to declare what version it is.

olizilla:
cooOooOoL. The e2e tests are failing because the robot is trying to click the settings link, but due to the default screen size that puppet opens pages at 800x600 the revision, the revision link appears over the settings link!

screenshot 2018-10-04 11 47 53

now, my first instinct was to update the tests to use a large scree size, but I think this is actually our bug... @fsdiogo could you add a custom media query to hide the revision link where the screen height would mean that it would be covering the nav items.

PR #845, opened 2018-10-02
fix: skip window.ipfs

### LINKED ISSUE #787: Problem with window.ipfs and Files/Settings
Something that we should address before shipping the first iteration:

## Background

When WebUI is bundled with IPFS Companion it detects and uses original API object available in WebExtension context, but if WebUI is loaded over HTTP it tries to use `window.ipfs` injected by IPFS Companion.

## Problem

The problem with `window.ipfs` in current form is a limited subset of IPFS API and sensitive things like access to `ipfs.config.*` is blocked and access to `ipfs.files.*` is sandboxed. 

Current  UX is problematic. `window.ipfs` not only disables the Settings screen, but the sandboxing [make it look like Files screen is broken](https://user-images.githubusercontent.com/157609/45311602-e8dec400-b529-11e8-96a8-44db26cb6562.png). 

User will be really confused if `ipfs files ls` shows files in the terminal, but WebUI claims it is empty.

## Solution

- Files screen should not log entire stacktrace in code below (quick fix: `String(error)`):
  https://github.com/ipfs-shipyard/ipfs-webui/blob/a67135c0a5041ed5c7a6d5216e9b664cb5336726/src/bundles/config.js#L64
- There should be  a note at the top of Files screen informing user  that listed files come from "window.ipfs sandbox at /path/to/sandbox/root". 
  It may be a good UX to display faded-out breadcrumbs prefixing  root with the path of sandbox root.

Thoughts?

### LINKED ISSUE #837: Smart API detection when loaded from non-standard API port in go-ipfs
This is something came to me when I was looking at https://github.com/ipfs/js-ipfs/pull/1595  and https://github.com/ipfs-shipyard/ipfs-webui/issues/785. 

If go-ipfs runs API on a different port than `:5001`, then webui will not be able to connect to it, even tho it was loaded from the API port :upside_down_face: 

Instead, it will display [error page  where user needs to manually provide the port](https://user-images.githubusercontent.com/157609/46203682-d088ea80-c31a-11e8-9683-c6c3342c4124.png).  

Its not bad, and may be enough for now, however I feel we could make it work automatically:
- **(A)** by appending parameter with custom API port to the webui URL printed to the console  (see https://github.com/ipfs/js-ipfs/pull/1595 and  discussion in https://github.com/ipfs-shipyard/ipfs-webui/issues/785)
- **(B)** by adding a simple fallback to webui: **if default port fails, check the same hostname but under the port  from `window.location.port`** (`""` means 80/443)

I think (B) makes more sense, it would fix the issue for both go-ipfs and js-ipfs without adding any cruft to URL. 

@olizilla  thoughts?

olizilla:
**blocked on**
- [x] release ipfs-redux-bundle with https://github.com/ipfs-shipyard/ipfs-redux-bundle/pull/15

Web UI wants all the privileges. With window.ipfs the files view
gets silo'd to the files added _in this origin_ so it's gonna
cause confusion.

This PR updates thing to use the new ipfs-redux-bundle and skips
window.ipfs in the provider checks.

Also adds an identity bundle as we no longer use /id as the
ipfs connection test, to move us towards supporting the read-only
api where it is all that is available.

fixes #787 
fixes #837 

License: MIT

### DIFF
--- a/src/components/provider-link/ProviderLink.js
+++ b/src/components/provider-link/ProviderLink.js
@@ -9,6 +9,9 @@ const providers = {
   },
   'window.ipfs': {
     url: 'https://github.com/ipfs-shipyard/ipfs-companion'
+  },
+  'ipfs-companion': {
+    url: 'https://github.com/ipfs-shipyard/ipfs-companion'
   }
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
-  "name": "ipfs-webui-next",
-  "version": "0.0.0",
+  "name": "ipfs-webui",
+  "version": "2.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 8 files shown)


hacdias:
@olizilla I'm just wondering: going directly to IPFS Companion's instance of IPFS is easy. Won't other apps be able to do the same?

olizilla:
@hacdias the "direct from companion" option only works if your app as been bundled with companion.

hacdias:
@olizilla awesome then 😄

src/components/provider-link/ProviderLink.js
```diff
@@ -9,6 +9,9 @@ const providers = {
   },
   'window.ipfs': {
     url: 'https://github.com/ipfs-shipyard/ipfs-companion'
+  },
+  'ipfs-companion': {
```
olizilla: @lidel this should bring back the provider linking from Web UI's node info.

Issue #827, opened 2018-09-26
Show count and percentage in distribution pie chart tooltip


alanshaw:
<img width="381" alt="screen shot 2018-09-26 at 10 02 13" src="https://user-images.githubusercontent.com/152863/46069379-5833f500-c173-11e8-9d46-d2bcc4661f0b.png">

- [x] Show the count and the percentage value in the toolip like "United States 60% (450 peers)
- [x] Make title a link to the peers page

olizilla:
This is causing confusion, see: https://github.com/ipfs-shipyard/ipfs-webui/issues/846
I am bumping up the priority.

olizilla:
@fsdiogo this one is an easy win

PR #859, opened 2018-10-16
fix: count in pie charts

fsdiogo:
pie

Fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/827.


src/status/CountryChart.js
```diff
@@ -44,6 +44,19 @@ const CountryChart = ({ t, peerLocations, className }) => {
...
       position: 'bottom'
+    },
+    tooltips: {
+      displayColors: false,
+      callbacks: {
+        title: (tooltipItem, data) => data.labels[tooltipItem[0].index],
+        label: (tooltipItem, data) => {
+          const percent = data['datasets'][0]['data'][tooltipItem['index']]
+          const count = Math.round((percent * totalCountries) / 100)
+          const peerStr = count === 1 ? t('peer') : t('peers')
```
alanshaw: Does the translations module allow us to do plurals?

fsdiogo: Yes, it does. But as we already had those two keys I used those for convenience.

I change that though. What do you think?

olizilla: @fsdiogo better to use the i18next pluralisation magic as languages can have multiple plural forms depending on the number https://www.i18next.com/translation-function/plurals#languages-with-multiple-plurals

fsdiogo: Hmm, strangely I can't get that to work 🤔

```js
// public/locales/en/status.json

{
  ...
  "peersCount": "Peer",
  "peersCount_plural": "Peers",
  ...
}
```

```js
// src/status/CountryChart.js

const peerStr = t('peersCount', { count: count })
```

This should work, but it's always returning the singular form, even when the count is > 1.

Any ideas?

olizilla: We've configured i18next to use the ICU format for locales as it's a standard, rather than the default i18next format. See https://react.i18next.com/misc/using-with-icu-format

src/status/CountryChart.js
```diff
@@ -44,6 +44,19 @@ const CountryChart = ({ t, peerLocations, className }) => {
...
+    tooltips: {
+      displayColors: false,
+      callbacks: {
+        title: (tooltipItem, data) => data.labels[tooltipItem[0].index],
+        label: (tooltipItem, data) => {
+          const percent = data['datasets'][0]['data'][tooltipItem['index']]
+          const count = Math.round((percent * totalCountries) / 100)
+          const peerStr = t('peersCount', { count: count })
+
+          return `${percent}% (${count} ${peerStr})`
```
alanshaw: I'd translate this whole string as it wouldn't make sense in a right-to-left language.

fsdiogo: rollsafe

Issue #810, opened 2018-09-25
Fix storybooks


olizilla:
Some of our storybooks don't render currently

- [x] Files
- [x] Nav
- [x] Settings page

Would be great to get these fixed, so we can publish our story books with the deployment and add more.

PR #862, opened 2018-10-18
fix: storybook

fsdiogo:
Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/810.


fsdiogo:
Can't seem to get the files components working in storybook because of `react-dnd`.

@hacdias any ideas of a work around?

hacdias:
@fsdiogo we might need to wrap the Files component with the required Drag and Drop components like here: https://github.com/ipfs-shipyard/ipfs-webui/blob/master/src/App.js#L67

olizilla:
@hacdias optional props don't generally need default values. Required props should have a sensible default where possible to make the api more plesant to use.

src/files/files-list/FilesList.js
```diff
@@ -350,7 +351,9 @@ const dropCollect = (connect, monitor) => ({
   canDrop: monitor.canDrop()
 })
 
+export const StoryFilesList = DropTarget(NativeTypes.FILE, dropTarget, dropCollect)(translate('files')(FilesList))
+
 export default connect(
   'selectNavbarWidth',
-  DropTarget(NativeTypes.FILE, dropTarget, dropCollect)(translate('files')(FileList))
+  DropTarget(NativeTypes.FILE, dropTarget, dropCollect)(translate('files')(FilesList))
```
olizilla: Let's rename `StoryFilesList` to be `FilesListWithDropTarget` and reuse it in the default export here.

src/files/selected-actions/SelectedActions.stories.js
```diff
@@ -15,12 +18,8 @@ storiesOf('Files', module)
...
         rename={action('Rename Files')}
-        inspect={action('Inspect Files')}
-      />
-    </div>
-  ))
-  .add('Actions for Multiple Files', () => (
-    <div className='ma2'>
+        inspect={action('Inspect Files')} />
+
       <SelectedActions
```
olizilla: Why merge these into 1 story? Seems more natural to have a story per state, or 1 story with knobs on so the user can try out different states https://github.com/storybooks/storybook/tree/master/addons/knobs

hacdias:
> @hacdias optional props don't generally need default values. Required props should have a sensible default where possible to make the api more plesant to use.

Tweaked @olizilla!

fsdiogo:
@olizilla I thought the settings page story was some kind of duplication of the json editor one. As none of those were working I couldn't see the difference so I deleted it.

I find it a bit strange to have a full page as a story, but strongly agree that is pretty useful to check the different states on the UI.

I'll resurrect it 👼

PR #865, opened 2018-10-19
chore: update locales from transifex

olizilla:
- new locales
- Remove fixed width from settings page buttons, as it breaks the layout
for langs where the text is longer than the max-width.

License: MIT


olizilla:
Tests pass, and no major changes here, so I'm merging it so we can tag a release.

Issue #795, opened 2018-09-13
i18n: Language selector


olizilla:
The new webui is using the value returned from `navigator.language` to pick the locale to load. There is no way to override that from the UI. It'd be great to have a language selector that lets you change the current locale manually.

This would expose the number of languages that the UI is available in, which is a nice side effect, and would give us a place to add a "contribute a translation" call to action to get more eyes on https://www.transifex.com/ipfs/

This would also help with developing and testing the UI in different languages. Setting up multiple profiles for different languages is a faff.

olizilla:
@fsdiogo we need to sketch out what this should look like, if you have any thoughts, or can grab some good examples from elsewhere that'd be rad. Maybe, just maybe, there is already a good react component for this floating around out there in npmspace somewhere.

olizilla:
yep, I think ideally we'd get the language name in the local language instead of the shortcode. The shortcode is always in ascii, so isn't super friendly to non-latin-alphabet users. Would be good to get some screenshots of some existing language selectors that do this well. Should we include flags or are they divisive in some locales?

Can we just expose or query the list of locales we set up here? https://github.com/ipfs-shipyard/ipfs-webui/blob/1774658fb277ea6b6d32238b43d0a1f00f3919bb/src/i18n.js#L26

fsdiogo:
Yeah, we should have the language name instead of the shortcode for sure!

I've been poking around a bit and to my surprise the dropdown isn't the tool of choice, but rather the modal:

|Google|Dropbox|Soundcloud|
|------|-------|----------|
|![google](https://user-images.githubusercontent.com/33324750/47080847-fceca400-d200-11e8-8d4e-e55664e166b9.gif)|![dropbox](https://user-images.githubusercontent.com/33324750/47080848-fceca400-d200-11e8-8e2e-bd948748c874.gif)|![soundcloud](https://user-images.githubusercontent.com/33324750/47080849-fd853a80-d200-11e8-8702-83abe5f7d33f.gif)|

An explanation to this can be the vast number of languages supported, in a dropdown it wouldn't be that much user friendly.

I didn't think the flags were divisive and I thought we should exclude them, but after scanning through a few articles I'm not sure anymore:
- https://usersnap.com/blog/design-language-switch/
- http://www.flagsarenotlanguages.com/blog/best-practice-for-presenting-languages/
- http://daily.unitedlanguagegroup.com/stories/editorials/inside-design-language-selector-no-flags


[...]

PR #866, opened 2018-10-19
[WIP] feat: add language selector

fsdiogo:
What's missing:

- [ ] Get all the available locales
- [ ] Show full language name instead of the short code

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/795.


fsdiogo:
lang

Don't know why I only get `en` when querying the i18n languages.

fsdiogo:
Current status:

i18n

Why use modal instead of dropdown: https://github.com/ipfs-shipyard/ipfs-webui/issues/795#issuecomment-430580143

src/settings/SettingsPage.js
```diff
@@ -2,23 +2,43 @@ import React from 'react'
...
 }) => (
   <div data-id='SettingsPage'>
     <Helmet>
       <title>{t('title')} - IPFS</title>
     </Helmet>
+
+    <Box className='mb3 pa4'>
+      <div className='flex justify-between items-center'>
+        <p className='ma0 lh-copy charcoal-muted f6'>
+          {t('language', { language: isoLangs[i18n.language].nativeName })}
```
olizilla: `isoLangs[i18n.language].nativeName` is risky. It assumes that whatever comes back from i18n.language will definitly be a key the isoLangs object. 

As an example, `i18n.language` reports `en-GB` on my machine becuase I am so very british i have a locale for it. As such the settings page blows up for me.

olizilla:
Why is `svenska` listed all lowercase in that gif? Is it all officially an all lowercase language?

src/settings/SettingsPage.js
fsdiogo: 😮 gotta figure out a work around for that then.

fsdiogo:
We now have a `languages.json` that includes the language codes with the country variants.

Unfortunately the process is not automatic, when we add a new language we have to manually import it in the `i18n.js` and if it has a country variant we need to add it in the same file.

src/settings/SettingsPage.js
```diff
@@ -2,23 +2,40 @@ import React from 'react'
...
 }) => (
   <div data-id='SettingsPage'>
     <Helmet>
       <title>{t('title')} - IPFS</title>
     </Helmet>
+
+    <Box className='mb3 pa4'>
+      <div className='flex justify-between items-center'>
+        <p className='ma0 lh-copy charcoal-muted f6'>
+          {t('language', { language: getCurrentLanguage() })}
```
olizilla: UI suggestion, make the label look like our block labels, make the language name larger and more obvious... if you need to change it you're gonna need it big and clear, and bring the button next to it.

screenshot 2018-10-31 at 11 06 03

It could also be it's own compoent, that opens and closes the model, so the parent page doesn't need to know abou it.

src/settings/language-modal/LanguageModal.js
```diff
@@ -0,0 +1,52 @@
...
+    onLeave()
+  }
+
+  return (
+    <Modal {...props} className={className} onCancel={onLeave} >
+      <ModalBody icon={SpeakerIcon}>
+        <p className='gray w-80 center'>{t('languageModal.description')}</p>
+        <div className='pa2 flex flex-wrap'>
+          { localesList.map((lang) =>
+            <button
```
olizilla: For some reason the button styles need to include a background color to reset them on FF, and we need to give the text more space. Perhaps the modal could be wider.

screenshot 2018-10-31 at 11 12 35

fsdiogo:
Lastest commit introduces two changes:

- Updated the language selector UI
- Updated the IPFS Config layout from display table to flex and upgraded its responsiveness

## Language selector

lang

## IPFS Config responsiveness

resp

src/lib/languages.json
```diff
@@ -0,0 +1,433 @@
+{
+  "af-NA": "Afrikaans (Namibia)",
```
olizilla: I think it'd be more friendly to display the languages in their local name, rather than english for all the things.

screenshot 2018-11-02 at 12 09 34

olizilla: @lidel have you got any sources for languages in their local language. How do we deal with the multiple Chinese varients? I'm stuggling to find a simple process for this. I have pinged transifex.

olizilla: More here https://github.com/ipfs/i18n/issues/4

lidel: I was about to link https://github.com/mozilla/language-mapping-list/blob/master/language-mapping-list.js but you already found it :) (`nativeName` is what we want to show to end user)

olizilla:
woo haa

screenshot 2018-11-02 at 17 48 46

olizilla:
@lidel pls can you eyeball this when you get a moment

fsdiogo:
Can this be merged @olizilla @lidel?

The language code was being displayed when it wasn't available in the `languages.json`. In the latest commit if the language and country code isn't available, I fallback to showing only the language code if possible, if not I fallback to English because is the default language of the Web UI.

public/locales/en/settings.json
```diff
@@ -3,6 +3,16 @@
   "save": "Save",
   "saving": "Saving...",
   "reset": "Reset",
+  "language": "Language",
```
lidel: Would it be helpful to clarify what is the target of language picker? 
For example:
```suggestion
  "language": "Web UI Language",
```

fsdiogo: I don't think it's necessary. It's expected to change the Web UI language, as we are in the Web UI settings no? 🤔

olizilla: I think `Language` is ok.

src/components/language-selector/language-modal/LanguageModal.js
```diff
@@ -0,0 +1,52 @@
...
+import { getLanguage } from '../../../lib/i18n'
+
+// Components
+import { Modal, ModalBody, ModalActions } from '../../modal/Modal'
+import SpeakerIcon from '../../../icons/StrokeSpeaker'
+import Button from '../../button/Button'
+
+const LanguageModal = ({ t, tReady, onLeave, link, className, ...props }) => {
+  const handleClick = (lang) => {
+    console.log(lang)
```
olizilla: rm `console.log(lang)`

src/i18n.js
```diff
@@ -22,18 +22,32 @@ import sv from 'i18next-icu/locale-data/sv'
...
-    localeData: [cs, da, de, en, eo, es, fr, it, ko, nl, no, pl, pt, ru, sl, sv, uk, zh]
-  }))
+  .use(new ICU({ localeData: localeData }))
   .use(XHR)
   .use(LanguageDetector)
   .init({
     ns: ['welcome', 'status', 'files', 'explore', 'peers', 'settings', 'notify'],
-    fallbackLng: 'en',
+    fallbackLng: {
+      'zh-HANS': ['zh-CN', 'en'],
```
olizilla: After much digging, i think `Hans` and `Hant` should only have the first letter capitalized

```suggestion
      'zh-Hans': ['zh-CN', 'en'],
```

src/i18n.js
```diff
@@ -22,18 +22,32 @@ import sv from 'i18next-icu/locale-data/sv'
...
+      'zh-HANT': ['zh-TW', 'en'],
```
olizilla:
```suggestion
      'zh-Hant': ['zh-TW', 'en'],
```

src/i18n.js
olizilla: See:
- Mozilla langmap https://github.com/mozilla/language-mapping-list/blob/6a1ba99f9625486705f0872750df228f52498bea/language-mapping-list.js#L837
- Spec http://www.ietf.org/rfc/bcp/bcp47.txt

PR #870, opened 2018-10-26
 fix: add missing leveldown dependency

leoparis89:
Unable to start or build the project because of missing dependecy to leveldown so I added it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6205,6 +6205,33 @@
         "leveldown": "^3.0.2",
         "levelup": "^2.0.2",
         "pull-stream": "^3.6.1"
+      },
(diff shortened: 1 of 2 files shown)


olizilla:
Hi! Thanks for contributing. Can you give me any more info? If I delete my node_modules, I can run

```console
$ npm install
# much npm logging

$ npm start
```
and all works as expected. The same process is working on our CI build.

What npm and node versions are you using when you hit this issue? It's expected to work on npm > 6 and node > 8

leoparis89:
I don't have the issue with npm 6.4.1
I probably had an older version at the time where I had the bug

Issue #871, opened 2018-10-26
Responsive Layout for Mobile Devices


lidel:
I've been playing with:
- [Sweet IPFS](https://discuss.ipfs.io/t/sweet-ipfs-full-ipfs-node-for-android/3779?u=lidel) (go-ipfs for Android packaged by a community member)
- [Firefox for Android Beta](https://play.google.com/store/apps/details?id=org.mozilla.firefox_beta) (runs the same webextensions as desktop version)
- ipfs-companion beta [v2.5.1.11870](https://github.com/ipfs-shipyard/ipfs-companion/releases/tag/v2.5.1.11870) (bundles new ipfs-webui)

I opened bundled Web UI and it works as expected :1st_place_medal:   \o/ but is missing some low hanging fruits such as collapsing  left menu on narrow viewport:

| Status | Files | Settings |
| ---- | ---- | ---- | 
| ![screenshot_20181026-163859](https://user-images.githubusercontent.com/157609/47574159-434bac80-d93f-11e8-90fc-bec917c1a8be.png) |  ![screenshot_20181026-165424](https://user-images.githubusercontent.com/157609/47574541-18ae2380-d940-11e8-8da3-645d940d964b.png) | ![screenshot_20181026-165435](https://user-images.githubusercontent.com/157609/47574558-219ef500-d940-11e8-99ff-5bb7f9ec1b53.png) |

olizilla:
Rad! That'd be a pretty simple fix.

olizilla:
_Scientists was so preoccupied with if they could, they didn't stop to think if they should_

![webui-responsive-navbar](https://user-images.githubusercontent.com/58871/47913392-d0d55200-de93-11e8-8784-f68a54cba67c.gif)

olizilla:
it was not simple.

fsdiogo:
Fixed in #877.

olizilla:
I'm happy to close this for now, but there is a ton of polishing to do to actually have a good reponsive layout, but that is a low pritority for now.

fsdiogo:
Yes, I jumped the gun closing this one because I thought it was only referring to the navbar responsiveness. 

We can reopen it later when someone starts working on it!

PR #877, opened 2018-11-02
feat: responsive nav

olizilla:
This is a minimal implementation of a responsive nav as part of #871 

We can go fancier in the future, but this is a minimal change to behave reasonably on small screens.

webui-responsive-navbar

On small screens, the navbar has a horizontal scroll, so you can swipe across to see the other nav items.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -30,12 +30,12 @@ export class App extends Component {
 
     return (
       <div className='sans-serif' onClick={navHelper(this.props.doUpdateUrl)}>
-        <div className='flex' style={{ minHeight: '100vh' }}>
-          <div className='flex-none bg-navy' style={{ width: navbarWidth }}>
+        <div className='flex-ns' style={{ minHeight: '100vh' }}>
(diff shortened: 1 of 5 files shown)


lidel:
Switched to desktop and noticed that on my linux box there is a strange "gap" that was not present before: 

|v2.1.1 |   this PR |
| ---- | ---- | 
 |   |

olizilla:
@lidel wat is that bar. Can you inspect it and give me any more info? I don't see it on Firefox or Chrome on mac.

fsdiogo:
I can't reproduce it either (macOS).

fsdiogo:
Yup, adding the class `overflow-x-hidden-l` fixes that.

I was able to reproduce that on a mac applying this settings (General):
 
scroll

fsdiogo:
And a wild pokemon appears:

poke

fsdiogo:
## Desktop



## Mobile

olizilla:
@lidel the last commit from @fsdiogo should fix the scrollbar on linux, can you eyeball!

PR #883, opened 2018-11-09
feat: IPFS Desktop settings

hacdias:
More info see: https://github.com/ipfs-shipyard/ipfs-desktop/issues/674.

IPFS Desktop injects a bundle on `window.ipfsDesktop`. If it's present, we show IPFS Desktop specific settings.

image

(Screenshot taken with IPFS Desktop)

License: MIT

### DIFF
--- a/src/bundles/ipfs-desktop.js
+++ b/src/bundles/ipfs-desktop.js
@@ -11,16 +11,16 @@ if (window.ipfsDesktop) {
       if (!action.type.startsWith('DESKTOP_')) {
         return state
       }
-  
+
       if (action.type === 'DESKTOP_SETTINGS_CHANGED') {
         return action.payload
       }
-  
+
       return state
     },
 
     selectDesktopSettings: state => state.ipfsDesktop,
-  
+
     doDesktopStartListening: () => async ({ dispatch }) => {
       window.ipfsDesktop.onConfigChanged(config => {
         dispatch({
@@ -29,7 +29,7 @@ if (window.ipfsDesktop) {
         })
       })
     },
-  
+
     doDesktopSettingsToggle: (setting) => () => {
       window.ipfsDesktop.toggleSetting(setting)
     },
--- a/src/bundles/ipfs-desktop.js
+++ b/src/bundles/ipfs-desktop.js
@@ -1,11 +1,18 @@
-const bundle = {
-  name: 'ipfs-desktop',
+let bundle = {
+  name: 'ipfsDesktop',
   reducer: (state = {}) => state,
-  selectIsIpfsDesktop: () => !!window.ipfsDesktop,
-  selectDesktopSettings: window.ipfsDesktop.selectDesktopSettings,
-  doDesktopStartListening: window.ipfsDesktop.doDesktopStartListening,
-  doDesktopSettingsToggle: window.ipfsDesktop.doDesktopSettingsToggle,
-  init: window.ipfsDesktop.init
+  selectIsIpfsDesktop: () => !!window.ipfsDesktop
+}
+
+if (window.ipfsDesktop) {
+  bundle = {
+    ...bundle,
+    reducer: window.ipfsDesktop.reducer,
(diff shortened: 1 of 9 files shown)


hacdias:
The injected code can be found here: https://github.com/ipfs-shipyard/ipfs-desktop/blob/new/src/hooks/webui/preload.js .,

src/settings/DesktopSettings.js
```diff
@@ -0,0 +1,55 @@
...
+      <Title>IPFS Desktop</Title>
+
+      <CheckboxSetting checked={desktopSettings['autoLaunch'] || false}
+        title='Launch on startup'
+        onChange={() => doDesktopSettingsToggle('autoLaunch')} />
+      <CheckboxSetting checked={desktopSettings['screenshotShortcut'] || false}
+        title='Auto add screenshots'
+        onChange={() => doDesktopSettingsToggle('screenshotShortcut')}>
+        <p className='mb0 mt1 lh-copy'>
+          Use <Key>CTRL/CMD</Key>+<Key>ALT</Key>+<Key>S</Key> to take screenshots and add them to the repository.
```
olizilla: Help the user out here by best-guessing the operating system and showing the key combo that's relevent to them.

src/bundles/ipfs-desktop.js
```diff
@@ -0,0 +1,8 @@
+const bundle = {
+  name: 'ipfs-desktop',
+  reducer: (state = {}) => state,
+  selectHasIpfsDesktop: () => !!window.ipfsDesktop,
```
olizilla: minor but I think the api for the rest of the app should be `selectIsIpfsDesktop`. I can see that it's named becuase it's checking for the presence of `window.ipfsDesktop`, but that is an implementation detail that may change, while the rest of the app only cares if it **is** "desktop" or not.

src/bundles/ipfs-desktop.js
```diff
@@ -0,0 +1,8 @@
...
+  ...(window.ipfsDesktop || {})
```
olizilla: It'd be a duplication, but I think we should spread the functions from ipfsDesktop that we expect to see here, so we can read the api of this bundle from the webui source code, rather than having to check elsewhere.

hacdias:
@olizilla working and with a storybook 😄

src/settings/DesktopSettings.js
```diff
@@ -0,0 +1,57 @@
...
+      <Title>IPFS Desktop</Title>
+
+      <CheckboxSetting checked={desktopSettings['autoLaunch'] || false}
+        title='Launch on startup'
+        onChange={() => doDesktopSettingsToggle('autoLaunch')} />
+      <CheckboxSetting checked={desktopSettings['screenshotShortcut'] || false}
+        title='Auto add screenshots'
+        onChange={() => doDesktopSettingsToggle('screenshotShortcut')}>
+        <p className='mb0 mt1 lh-copy'>
+          Use <Key>{isMac ? 'CMD' : 'CTRL'}</Key>+<Key>ALT</Key>+<Key>S</Key> to take screenshots and add them to the repository.
```
olizilla: - <kbd>ALT</kbd> is the <kbd>⌥ option</kbd> key on mac. 
- <kbd>CTL</kbd> is the <kbd> ⌘ command</kbd> key on mac

src/settings/DesktopSettings.js
```diff
@@ -0,0 +1,57 @@
...
+    </div>
+  </div>
+)
+
+const Key = ({ children }) => <span className='monospace br2 bg-snow ph1'>{ children }</span>
+
+export function DesktopSettings ({ doDesktopSettingsToggle, desktopSettings }) {
+  return (
+    <Box className='mb3 pa4'>
+      <Title>IPFS Desktop</Title>
```
olizilla: Please replace the hardcoded strings with i18n keys

olizilla:
@hacdias please revert the change to the package-lock.json. We need to upgrade ipld and ipld-explorer-components together or we hit breaking changes that causes errors when trying to view some ipld types.

Issue #878, opened 2018-11-03
Bug: missing protocol breaks the peers page


jbenet:
Missing protocol (quic) breaks execution in the peers page, and causes it all to fail. 

may want to:
- add quic (known protocol) to codec.js
- isolate failures (one failing protocol shouldnt break everything. assume protocols may exist that are not yet registered in the codec.js table)
- level up codec.js to deal with or admit unknown protocols.

```
protocols-table.js:17 Uncaught Error: no protocol with name: quic
    at i (protocols-table.js:17)
    at u (codec.js:45)
    at y (codec.js:170)
    at Object.fromString (codec.js:178)
    at o.l.proto.className (index.js:43)
    at new Multiaddr (index.js:40)
    at l.proto.className (index.js:26)
    at Multiaddr (index.js:40)
    at t.Strings.r.map.e (peers.js:48)
    at Array.map (<anonymous>)
```

![image](https://user-images.githubusercontent.com/138401/47946475-8969b880-dec9-11e8-8add-bb87294ecc7f.png)

olizilla:
I think webui has handled this as gracefully as it can, the call to `ipfsApi.swarm.peers()` has thrown and uncaught error, so webui recieves no information to about the current peers list.

`js-ipfs-api` needs to handle multiaddr errors when unpacking the the api response. I'll issue and PR that.

olizilla:
Issue for how to handle unvalidatable peers in the js-ipfs-api client https://github.com/ipfs/js-ipfs-api/issues/885

lidel:
Quick update  for drive-by visitors: 
- js-multiaddr v5.0.2 shipped with  [the fix](https://github.com/multiformats/js-multiaddr/pull/71), needs to be bumped in all libs that use it
- ipfs-companion [v2.5.1.12010](https://github.com/ipfs-shipyard/ipfs-companion/releases/tag/v2.5.1.12010) (beta) applies the above fix when bundled ipfs-webui is used

olizilla:
There is a PR to make js-ipfs-api robust to peer validation errors https://github.com/ipfs/js-ipfs-api/pull/887

and also of note there is an open issue to get multiaddr to support unverifiable addresses https://github.com/multiformats/multiaddr/issues/70

olizilla:
This was fixed  in  https://github.com/ipfs/js-ipfs-http-client/pull/887 and #884

PR #884, opened 2018-11-13
fix: support quic protocol in multiaddrs

olizilla:
Bump the multiaddr dep so we support the quic protocol in multiaddrs.

See:  #878


olizilla:
@lidel this is the minimal fix for quic support that I think you were hoping to see. I'm going to merge this and tag a release. I'll PR all the things.

PR #895, opened 2018-12-04
WIP: Feat/backend toggle switch for desktop settings

phoniks:
Seeking feedback.  I'm working on enabling the user to try out the Javascript implementation of IPFS by toggling a switch on the settings page.  This PR simply ads a checkbox on the DesktopSettings page.  I'm particularly interested in knowing whether having this as a checkbox makes sense. Originally I had in mind, based on the original discussions of this feature, to make this setting a select component, but during coding it became obvious that doing so would add a lot of code for not very much change in user experience. Perhaps in the future, if there are other implementations that can be chosen, it may make sense to do the dropdown, but for now it doesn't seem prudent.  But, perhaps since it potentially takes your node offline there should be an intermediate step between flipping the switch and actually shutting down the node and re-spawning.  Maybe a quick confirmation alert?


jessicaschilling:
@phoniks - please accept sincere apologies for this PR getting lost in the fray. Is this something you'd still like to discuss?

phoniks:
If it's still something that'd be valuable to the project, I'm happy to follow up on adding the functionality.  The last time I had checked in on this, I think some things about the project had changed pretty significantly, so it was going to require some additional work that I didn't have the bandwidth for at the time. I put it on the back burner and kind of lost track of it, but since we're all on mandatory Quarantine and Chill mode right now I could take another look.

jessicaschilling:
We're ramping back up on GUI-related work for this quarter, so any work is welcome -- though if you'd be interested in taking a look at the current state of the landscape while you quarantine and chill, have a look at https://github.com/ipfs/ipfs-gui/blob/update-readme/README.md (as you can see, it's in a PR so not final yet, ha!). There's info there about the general methodology we're taking, and a link to a unified ZenHub board for a glimpse into current open needs ... so there might be something there that you're more interested in these days? Happy to chat further if anything catches your eye. 😊

lidel:
Thank you for bringing this up @phoniks!

Since this PR was opened we decided to decouple ipfs-desktop-specific settings from ipfs-webui (https://github.com/ipfs-shipyard/ipfs-desktop/issues/1389, https://github.com/ipfs-shipyard/ipfs-webui/pull/1445) and move them to a submenu (https://github.com/ipfs-shipyard/ipfs-desktop/pull/1425):

  


---- 

That is why I am closing this PR, but the idea of giving a choice to switch to js-ipfs is still valid and is tracked under https://github.com/ipfs-shipyard/ipfs-desktop/issues/601

PR #909, opened 2018-12-13
Peers tab: Natural address sorting

terorie:
The peers in the tab are currently lexicographically sorted, meaning that instead of a listing like this:
```
/ipfs/12.2.3.4/…
/ipfs/21.2.3.4/…
/ipfs/100.2.3.4/…
```
this is displayed instead:
```
/ipfs/100.2.3.4/…
/ipfs/12.2.3.4/…
/ipfs/21.2.3.4/…
```

This patch sorts the peers using the `localeCompare` function with `numeric: true` to solve this.


anacrolix:
Is that an IPFS timeout in the CI?

terorie:
@anacrolix Yes, seems like the other PRs also get this error.

olizilla:
Neat! I've nudged CI and all is well. I agree, having a numeric sensitive sort on multiaddrs makes the list feel more coherent.

screenshot 2018-12-14 at 10 20 30

PR #914, opened 2018-12-18
fix: minimum translation percentage

fsdiogo:
The minimum percentage for pulling translations was `1%`; that pulled files with almost no strings translated.

Updated the minimum percentage to `75%`.


olizilla:
We need to clear our all the locales other than en from the `public/locales` dir, and then pull from transifex to see which translations dont meet the new threshhold. Then check which of the [localData imports](https://github.com/ipfs-shipyard/ipfs-webui/blob/master/src/i18n.js#L6-L25) we still need, and recreate the l[anguages.json](https://github.com/ipfs-shipyard/ipfs-webui/blob/master/src/lib/languages.json) which you can re-generate via [`lol`](https://www.npmjs.com/package/@olizilla/lol)

fsdiogo:
Done @olizilla!

Issue #922, opened 2019-01-03
Add folder always seem to fail


Gozala:
I think issue originally reported as https://github.com/ipfs/ipfs-gui/issues/73 probably belongs here.

PR #924, opened 2019-01-04
fix: add folder

fsdiogo:
Adding a folder was failing because of the length of the api response. It would also fail and give an incoherent error if adding a folder with an existing folder name. 

folder

Fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/922.

### DIFF
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -241,9 +241,9 @@ export default (opts = {}) => {
         progress: updateProgress
       })
 
-      if (res.length !== streams.length + 1) {
+      if (res.length !== streams.length + 2) {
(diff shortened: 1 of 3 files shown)


src/bundles/files.js
```diff
@@ -241,17 +241,22 @@ export default (opts = {}) => {
         progress: updateProgress
       })
 
-      if (res.length !== streams.length + 1) {
+      if (res.length !== streams.length + 2) {
```
fsdiogo: Not sure when this changed, but it was always failing because of this!

olizilla: interesting, it's returning an entry for the folder, and the parent of the folder. This fixes our logic around it, so +1

Issue #928, opened 2019-01-07
Failed to add files to IPFS


maooricio:
I updated the ipfs-webui and follow the instructions in order to install and run the tool.

I had the old version of this module and I could worked with it fine. I decided to update it because I noticed that many things changed both here and in js-ipfs. All works fine but when I try to upload a file, The module shows me an error that I don't understand why because I studied and analyzed all the connections and I think all are set just fine. In fact, the console doesn't throw me any error when I make the upload. What I am missing?

Here is the error:

![screenshot from 2019-01-07 11-20-15](https://user-images.githubusercontent.com/6587647/50779453-61525900-126e-11e9-81b3-450760bdade0.png)

And here is the console log

![screenshot from 2019-01-07 11-20-44](https://user-images.githubusercontent.com/6587647/50779470-6f07de80-126e-11e9-8cb5-65bb2d544586.png)

olizilla:
Yep, that should work. Can you give some more info?

- what commit version is listed in the bottom of the nav bar?
- what browser and version and OS are you using?
- did you drag and drop or did you click the button and choose add files or add folder?
- How many files / folders are you trying to add, and what is the total file size?

olizilla:
i'm seeing an error on master with drag and drop of directories on Firefox, which I'm looking into now

maooricio:
Hi @olizilla, look the answer to your questions:

1. commit/261f4f4
2. Google Chrome Version 68.0.3440.75 (Official Build) (64-bit). OS: Ubuntu 18.04.1 LTS
3. I clicked the button and I chose add files
4. One file with 200kb of total size

maooricio:
I see that the POST error that we can see in the console log image doesn't have a logical URL (http://dnsaddr/localhost/tcp/3000/http:5001/api/v0/get?arg=QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn&stream-channels=true). This doesn't have something related with the issue?, maybe

lidel:
Ok, I got the same result with ["Add file"](https://user-images.githubusercontent.com/157609/50865730-41728080-13a7-11e9-94cf-af629b220388.png) under both prebuilt `QmenEBWcAk3tN94fSKpKFtUMwty1qNwSYw3DMDFV6cPBXA` or dev version (`npm start`) tagged as `v2.3.2`:


[...]

PR #931, opened 2019-01-09
fix: add files and folders

fsdiogo:
Need your input about this one guys:

We can now add files and folders through the `+ Add to IPFS` button, I did that by saying if it's files or folders depending on what option we clicked on the dropdown. So far so good.

The problem is when we drag and drop. It seems there is not a standard way of checking if we're dropping a file or a folder in `react-dnd`.

I did this (it's not pretty, I know):
https://github.com/ipfs-shipyard/ipfs-webui/blob/11710148d1aba2f91106ca1d7722e896bcbe30ad/src/files/files-list/FilesList.js#L343

but after searching a bit and testing with some files I found that some file types will fail this check. [This issue talks about this](https://github.com/react-dnd/react-dnd/issues/840).

So, what options do we have? 
1. Disable drag and drop ~~for folders~~
2. Keep the code this way, knowing that some files may fail if dragging and dropping
3. ?

We'll need to refactor `+ Add to IPFS` dropdown and the drag and drop in the Files List, that will be a good time to think about how to circumvent this. 

Thoughts? (we need to ship the fix for adding files ASAP)

Fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/928.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ipfs-webui",
-  "version": "2.3.1",
+  "version": "2.3.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -227,7 +227,7 @@ export default (opts = {}) => {
       }
     },
 
-    doFilesWrite: make(actions.WRITE, async (ipfs, root, rawFiles, type, id, { dispatch }) => {
+    doFilesWrite: make(actions.WRITE, async (ipfs, root, rawFiles, id, { dispatch }) => {
       const { streams, totalSize } = await filesToStreams(rawFiles)
 
       // Normalise all paths to be relative. Dropped files come as absolute,
@@ -253,9 +253,6 @@ export default (opts = {}) => {
       const numberOfFiles = streams.length
       const numberOfDirs = countDirs(streams)
       const expectedResponseCount = numberOfFiles + numberOfDirs
-      console.log('added files', res.length, expectedResponseCount, { numberOfFiles, numberOfDirs, expectedResponseCount })
-      console.log(streams.map(s => s.path))
-      console.log('res', res)
 
       if (res.length !== expectedResponseCount) {
         // See https://github.com/ipfs/js-ipfs-api/issues/797
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -230,10 +230,11 @@ export default (opts = {}) => {
     doFilesWrite: make(actions.WRITE, async (ipfs, root, rawFiles, type, id, { dispatch }) => {
       const { streams, totalSize } = await filesToStreams(rawFiles)
 
-      // TODO: make file paths absolute before we get here
+      // Normalise all paths to be relative. Dropped files come as absolute,
+      // those added by the file input come as relative paths, so normalise them.
       streams.forEach(s => {
-        if (s.path[0] !== '/') {
-          s.path = '/' + s.path
+        if (s.path[0] === '/') {
+          s.path = s.path.slice(1)
         }
       })
 
@@ -245,7 +246,7 @@ export default (opts = {}) => {
 
       const res = await ipfs.add(streams, {
         pin: false,
-        wrapWithDirectory: true,
+        wrapWithDirectory: false,
         progress: updateProgress
       })
 
@@ -271,6 +272,7 @@ export default (opts = {}) => {
           try {
             await ipfs.files.cp([src, dst])
           } catch (err) {
+            console.log(err, { root, path, src, dst })
             throw Object.assign(new Error(`Folder already exists.`), { code: 'ERR_FOLDER_EXISTS' })
           }
         }
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -259,7 +259,6 @@ export default (opts = {}) => {
         throw Object.assign(new Error(`API returned a partial response.`), { code: 'ERR_API_RESPONSE' })
       }
 
-      // FIXME: if all filepaths are absolute this fails to handle individual files.
       for (const { path, hash } of res) {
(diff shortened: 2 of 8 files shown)


fsdiogo:
Oh damn, sorry for the confusion, I edited the first post...

**Option (1)** was to disable drag-and-drop altogether. The problem here is that we can't differentiate files from folders when dragging and dropping. After adding the dropped resources to IPFS we compare the response and stream length due to [this issue](https://github.com/ipfs/js-ipfs-api/issues/797) here:

https://github.com/ipfs-shipyard/ipfs-webui/blob/11710148d1aba2f91106ca1d7722e896bcbe30ad/src/bundles/files.js#L244

But as we don't know if it's files or folders it will fail sometimes.

This doesn't happen while adding through the file input because:

Oh damn, sorry for the confusion, I edited the first post...

**Option (1)** was to disable drag-and-drop altogether. The problem here is that we can't differentiate files from folders when dragging and dropping. After adding the dropped resources to IPFS we compare the response and stream length due to [this issue](https://github.com/ipfs/js-ipfs-api/issues/797) here:

https://github.com/ipfs-shipyard/ipfs-webui/blob/11710148d1aba2f91106ca1d7722e896bcbe30ad/src/bundles/files.js#L244

But as we don't know if it's files or folders it will fail sometimes.

This doesn't happen while adding through the file input because each button has a different input associated to it and we flag it from the get go:

https://github.com/ipfs-shipyard/ipfs-webui/blob/11710148d1aba2f91106ca1d7722e896bcbe30ad/src/files/file-input/FileInput.js#L126-L139

fsdiogo:
When converting the files to to streams in the bundle:

https://github.com/ipfs-shipyard/ipfs-webui/blob/11710148d1aba2f91106ca1d7722e896bcbe30ad/src/bundles/files.js#L230

This check happens to see if it's a dir or not:

https://github.com/ipfs-shipyard/ipfs-webui/blob/11710148d1aba2f91106ca1d7722e896bcbe30ad/src/lib/files.js#L27-L29

But has you can see it basically checks if the html element had the `webkitRelativePath` that is only present in the input, so when we drag and drop this is always false, even though it's a folder.

As you can see, this is very flimsy.

lidel:
@fsdiogo ok, lets keep drag & drop but add a simple best-effort heuristic for detecting folders that makes things work most of the time:

- if `file.type` is not empty, then we know 100% it is a file
- if `file.type` is empty and `file.size % 4096 == 0` then assume it is a directory
  - this is a quick hack to take advantage of the fact directory sizes are aligned to 4KB and this is best we can do right now – we can refine this heuristic later, in a separate PR

olizilla:
Ok, the problem here is react-dnd. it's not giving us direct access to the DataTransfer object: https://github.com/react-dnd/react-dnd/issues/712

There was a PR to add support for that in 2 months ago, but it got reverted.

With direct access to the DataTransfer object from drop event, we can handle directory drops in at least chrome and firefox: 

https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry

olizilla:
> if file.type is empty and file.size % 4096 == 0 then assume it is a directory
@lidel that is a neat hack, but is linux only. On macOS on FF I'm seeing file size as `0` (which we still might be able to use a clue, but it gets more complicated.)
screenshot 2019-01-10 at 13 16 12

fsdiogo:
> @lidel that is a neat hack, but is linux only. On macOS on FF I'm seeing file size as 0 (which we still might be able to use a clue, but it gets more complicated.)

On Chrome and Safari the size appears as 128.

olizilla:
There are a couple of related issues in here when dropping directories.

- Where we unpack a directory into a list of files, we incorrectly derive the filenames on Firefox as the `webkitRelativePath` contains the `root`, so when we join them as below, the root dir is duplicated in the file path so `/a/foo.txt` becomes `/a/a/foo.txt`. This doesn't happen on Chrome as `webkitRelativePath` is empty, so the `root` and `name` properties are used.
https://github.com/ipfs-shipyard/ipfs-webui/blob/261f4f46d5e260e1143996fa0f9115f64812ed17/src/lib/dnd-backend.js#L27-L31

- Where we calculate how many streams we expect on the response from adding to ipfs, we don't take into account the extra response entries for **every** sub directory that might exist when adding a deep tree of directories... we only add an extra one for the root directory. This fails in all browsers.
https://github.com/ipfs-shipyard/ipfs-webui/blob/11710148d1aba2f91106ca1d7722e896bcbe30ad/src/bundles/files.js#L244-L247

I am working on a fix.

fsdiogo:
You aced it @olizilla ‼️👌

> You can't drop files or dirs into a subfolder, they always land at the MFS root. I'm not sure that ever worked.

I'm confused about this, as it seems to work for me 🤔

olizilla:
@fsdiogo I may well have broken it. I wanted to share this fix early to get feedback, and I didn't dig into it. I'll take a look now!

fsdiogo:
Yes, that's what I meant 😄

fsdiogo:
@hacdias did you got the chance to test this on Windows?

hacdias:
@fsdiogo drag'n'drop of files & directories seem to be working here!

olizilla:
Ok, this has been a good nudge to refresh my "test on windows" flow, and I used browstack to do a real test of file and directory drag-and-drop and add via file input on an Win 11 + Edge VM remote desktop session in Firefox! (obviously this is no replacement for a real cross-browser test set up, but it's great for ad-hoc checks as we go)

ipfs-webui-add-files-edge

...this also shows the odd UX of our file browser for drag-and-drop... I can see that I'm deliberately avoiding dropping on the `..` folder, but that is the most enticing thing to drop on. We need to make the whole screen the drop target and add some visual indications of what will happen when dragging over.

ClementeSerrano:
Any solution to this issue?

Issue #813, opened 2018-09-25
Show file operations and meta info on file preview page


olizilla:
When I want to share a file, it feel natural to click on it. This opens up the file preview page, which shows the path and the file, but the actions aren't available.

**file actions**
<img width="1552" alt="screenshot 2018-09-25 15 29 51" src="https://user-images.githubusercontent.com/58871/46021177-efe40580-c0d7-11e8-82c7-66469be2ad09.png">

**file preview**
<img width="1552" alt="screenshot 2018-09-25 15 29 56" src="https://user-images.githubusercontent.com/58871/46021175-efe40580-c0d7-11e8-8bed-f04bb7498398.png">

It'd be great to still have access to:

- Share
- Download
- Delete
- inspect
- Rename

along with info like

- file CID
- file size
- assumed file type / file icon

fsdiogo:
Dropbox does it like this:

<img width="1069" alt="dropbox" src="https://user-images.githubusercontent.com/33324750/51045820-7146a180-15bc-11e9-9199-8e12ff5b8898.png">


I sketched out two proposals:

| 1. I like the cleanliness of this one | 2. Shameless Dropbox copy |
| --- | --- |
| <img width="1257" alt="proposal-1" src="https://user-images.githubusercontent.com/33324750/51050383-7c073380-15c8-11e9-8233-98f544a229f5.png"> | <img width="1257" alt="proposal-2" src="https://user-images.githubusercontent.com/33324750/51050387-80335100-15c8-11e9-8771-78cf3fd823a3.png"> |

What do you guys think? @ipfs-shipyard/gui

PR #936, opened 2019-01-15
feat: add file operations in preview

fsdiogo:
preview

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/813.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6219,29 +6219,10 @@
         "datastore-core": "~0.5.0",
         "encoding-down": "^5.0.4",
         "interface-datastore": "~0.5.0",
+        "level-js": "github:timkuijsten/level.js#18e03adab34c49523be7d3d58fafb0c632f61303",
         "leveldown": "^3.0.2",
         "levelup": "^2.0.2",
         "pull-stream": "^3.6.9"
-      },
-      "dependencies": {
-        "abstract-leveldown": {
-          "version": "2.4.1",
-          "resolved": "https://registry.npmjs.org/abstract-leveldown/-/abstract-leveldown-2.4.1.tgz",
-          "integrity": "sha1-s7/tuITraToSd18MVenwpCDM7mQ=",
-          "requires": {
-            "xtend": "~4.0.0"
-          }
-        },
-        "level-js": {
-          "version": "github:timkuijsten/level.js#18e03adab34c49523be7d3d58fafb0c632f61303",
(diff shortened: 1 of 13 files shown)


olizilla:
@fsdiogo I'm not sure why we're seeing so many changes to the package-lock.json... can you talk me through your workflow? This is probably all super obvious, but I want to share steps we go through when devloping to see if there are differences that might explain it.

When trying out a branch i tend to just run:

```console
git checkout <branch>
npm install
npm start
```

and that doesn't typically lead to a change in the package-lock.json (though it does on occasion, would be great to figure out why)

When adding a dep while working on a branch, i run

```console
npm install <dep>
```

and that updates both the `package.json` and the `package-lock.json`. Typically it just updates the section of the `package-lock.json` that is relevant to the added dep, Somtimes it changes more of the `package-lock.json` than expected as the dep tree of the new lib causes some other deps to get hoisted / deduped in the tree.

In this PR, there is no change to the `package.json` so I wouldn't expect any change to the `package-lock.json` to be included in the PR changeset. Our locked dep tree should only be chaged when we decide to add things or delibreately want to update things. 

At this moment, there are bunch of updates to ipld that we can't pull in until we do a bunch of work to update ipld-explorer-components. There are some other transitve deps in our tree that seem to regularly causes issues, something around level.js and leveldown. changes to them in the package-lock regularly cause build failures, so I'm being extra cautious about changes to the `package-lock`.

fsdiogo:
> ... and that doesn't typically lead to a change in the package-lock.json (though it does on occasion, would be great to figure out why)

That is my workflow. Although when I check out to a new branch and install the deps the `package-lock.json` seems to be updated more often that not.

What happened in the last two commits was that I removed the `package-lock.json` and generated a new one to try to pass the CI, but some tests were failing, so I reverted it.

olizilla:
@fsdiogo I hear you. When the package-lock does change without me editing the package.json, i've either been resetting it with a `git checkout -- package-lock.json` if it's just in my local changeset, or finding the commit sha for where i branched off master and resetting the package-lock so it is removed from the changeset altogether with `git checkout <sha from current master> package-lock.json`.

The general pattern I'm following is, if the `package.lock` hasn't changed in the PR, then the `packge-lock.json` shouldn't be in the PR.

fsdiogo:
I'll get the `package-lock.json` from master and this should be fixed!

fsdiogo:
Now rendering correctly in Firefox:

ff

Will merge when the CI is ✅

Issue #916, opened 2018-12-20
Files: drag&drop target is to small


lidel:
### Problem

Right now, only the file list seems to be the drop target, which makes it not only a chore to aim, but impossible to drag-and-drop into an empty directory, as the file dropped on `..` (the only item in empty directory) gets added to parent directory:

![peek 2018-12-20 11-42](https://user-images.githubusercontent.com/157609/50280400-abed6c00-044c-11e9-9d3b-8a9485235307.gif)

### Solution Ideas

- the whole viewport should be the drop target 
  - Ideally, even if user is not on Files screen
- unless user drops on a specific directory on Files screen, then file is added to currently opened directory or `/` 
- there should be a visual cue "Drop here to add to <MFS path>" to ensure user it is safe to drop at the current target

olizilla:
Strongly agree. I'm not clear on why we didn't make the whole viewport the drop target in the first place. @hacdias I know you had a reason, can you explain? We really need a solution to this. I'd say being able to drop anywhere is more important  than being able to drop on a dir in the filelist if we have to choose, though I think we can have both.

PR #942, opened 2019-01-18
feat: extend dropzone to page

fsdiogo:
- The dropzone was extended to the whole page
- We can now drop a file in any page: it is added to the root folder and the user redirected to the files page
- In the files page we can still drop the files in specific folders as before

dropit

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/916.


fsdiogo:
This is a branch of https://github.com/ipfs-shipyard/ipfs-webui/pull/936, so that should be merged first.

PR #948, opened 2019-01-28
chore: lazy load ipld formats

fsdiogo:
This needs to merged first https://github.com/ipfs-shipyard/ipld-explorer-components/pull/14 and released.


package.json
```diff
@@ -41,8 +41,8 @@
     "ipfs-geoip": "^3.0.0",
     "ipfs-redux-bundle": "^5.0.0",
     "ipfs-unixfs": "^0.1.15",
-    "ipld": "^0.18.0",
-    "ipld-explorer-components": "^1.0.1",
+    "ipld": "^0.21.1",
+    "ipld-explorer-components": "^1.1.0",
```
fsdiogo: This needs to be updated to the latest version when released.

package.json
```diff
@@ -19,6 +19,7 @@
     "test:e2e:ci": "cross-env URL=http://localhost:3001 run-p --race test:serve test:e2e",
     "analyze": "source-map-explorer build/static/js/main.*",
     "bundlesize": "bundlesize",
+    "bundle-report": "webpack-bundle-analyzer build/stats.json",
```
olizilla: is this a better bundle analyser than source-map-explorer? do we need both or would you recommend switching over to jus this one?

fsdiogo: I find it easier to look at `webpack-bundle-analyzer`, as it shows more info such as the gzipped size and such (it's possible that `source-map-explorer` does that too, I just don't know about it).

We only need one, I'd switch to `webpack-bundle-analyzer` if that's ok with you!

olizilla: @fsdiogo got for it 👍

Issue #926, opened 2019-01-05
Add loading indicator while waiting for files responses


olizilla:
If the user copies the root a data-set into their MFS, it is not automatically recursively fetched.  (it probably should be)

If a user drills down into a data-set that hasn't yet been fetched, the UI freezes with no information as to what is going on.

A quick test

```console
$ ipfs dns arewedistributedyet.com
/ipfs/QmeANzfbKFunvrenGDVbAttucJvNF6wrWH96QeSWdmgGy6

$ ipfs files cp /ipfs/QmeANzfbKFunvrenGDVbAttucJvNF6wrWH96QeSWdmgGy6 /awdy
```

Now navigate to the files page in the web ui and click on the `/awdy`  entry. It taks a  few seconds to load, while the data is fetched. There is no indication to the user. A larger data set can  hang things for minutes.

PR #949, opened 2019-02-01
feat: add loading animation for files resp

fsdiogo:
Adds a loading animation when changing folders because some requests can hang for a bit and there was no clear indication that something was going on.

When you click on a folder, the files list gets an opacity and I went for something like the youtube cards animation, but keeping the content visible: 

anim

Fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/926.

### DIFF
--- a/src/files/files-list/FilesList.js
+++ b/src/files/files-list/FilesList.js
@@ -5,7 +5,6 @@ import PropTypes from 'prop-types'
 import { connect } from 'redux-bundler-react'
 import { Trans, translate } from 'react-i18next'
 import { join } from 'path'
-import classnames from 'classnames'
 import { sorts } from '../../bundles/files'
 // Reac DnD
 import { NativeTypes } from 'react-dnd-html5-backend'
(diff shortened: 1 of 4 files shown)


src/files/files-list/FilesList.js
```diff
@@ -305,22 +310,24 @@ export class FilesList extends React.Component {
...
-            setIsDragging={this.isDragging}
-            translucent={isDragging || (isOver && canDrop)}
-            name='..'
-            focused={this.state.focused === '..'}
-            cantDrag
-            cantSelect
-            {...upperDir} />
-        }
-        {this.files}
+        <div className={wrapperAnimClass}>
```
olizilla: Let's extract this to a component! Could we have something like

```jsx
<LoadingAnimation loading={filesIsFetchin}>
  <File .../>
</LoadingAnimation>
```

fsdiogo: We can, but it won't just work out of the box by itself, because the parent container must have some logic and specific css classes for it to work.

fsdiogo: @olizilla can you check commit https://github.com/ipfs-shipyard/ipfs-webui/pull/949/commits/42a2304b6f0bf3e7bac9dff76359c9c2a6a04077 and see what you think?

Do you know a way of decoupling and keeping all the logic needed inside the `<LoadingAnimation>`, without having the `<FilesList>` component to have to deal with something?

olizilla: How about https://github.com/ipfs-shipyard/ipfs-webui/pull/949/commits/d6722ac2a1ab23e6ea33f8799a85cacfa0a4a21f

fsdiogo: Love it ❤️

PR #950, opened 2019-02-02
fixed lots of typos and missing spaces

alexandreleroux:


olizilla:
Thank you for taking the time to do this! We use Transifex to manage our translation process, the locale json files are generated from there, so these changes would be lost next time we run the build process. Would you port these fixes to https://www.transifex.com/ipfs/ipfs-webui/dashboard/ ?

This is a good indicator that we need  to do more to make that clearer. We are storing the translations in the repo, to make it easier for developers to try out other locales. But we could remove them and add instructions for devs on how to fetch them on demand.

alexandreleroux:
Thanks @olizilla for the info. I have created a Transifex account and team membership for the IPFS Web UI translations. Once I get approved, I'll try to find the time to contribute back those proposed changes through Transifex.

And yes, it would be pertinent to ensure others don't so the same mistake I did by editing the copy on this repos. Thanks! :-)

alexandreleroux:
Done, fixes to translation applied on Transifex

Issue #915, opened 2018-12-20
Files: make it easier to create a new folder


lidel:
A lot of "file management" starts with creating a new folder and moving things to it.
Right now "New folder" action is hidden under "Add to IPFS" button and hard to discover and requires 2 clicks instead of one. 

What if we make it a top-level button (in faded out, much less  prominent colors than "Add to IPFS")?

> ![2018-12-20--11-21-43](https://user-images.githubusercontent.com/157609/50279389-d7bb2280-0449-11e9-92e9-368768ead8f5.png)

Thoughts?

olizilla:
Yep, it's confusing to have "new folder" in the same list. I'd like to do some design exploration on what we can do to make the "add to IPFS" process much better. "Add by path" is confusing for new users and it sound like "file system path", so I think we need to re-think this menu entirely.

lidel:
Indeed, "Add by path" suggests something is not on IPFS and needs to be added, while all it does it import of existing IPFS data to MFS.

I guess it could also be moved to top-level, and/or renamed to something like "Import IPFS path".

fsdiogo:
**Dropbox** doesn't use dropdowns, all options are visible:

<img width="625" alt="dropbox" src="https://user-images.githubusercontent.com/33324750/50845297-6abded00-1364-11e9-9bc2-41ab23853de9.png">

I agree we should think about how to improve this menu and the overall experience of dragging and dropping files on the Web UI.

PR #951, opened 2019-02-04
chore: update add dropdown

fsdiogo:
As per https://github.com/ipfs-shipyard/ipfs-webui/issues/915#issuecomment-452768683, I've refactored the add dropdown. Heavily inspired by Drive and Dropbox.

- Separate the `create new folder` from the `add to ipfs` dropdown. This way, everything inside that dropdow is to add file/folder/cid to our repo.
- Renamed the dropdown to just `+ Add`. Previously we had `+ Add to IPFS`, which is not 100% correct. This let me update the `Add by path` which was super confusing, to `Add from IPFS`, which I think is more clear because it tells you you're adding something that is already on IPFS. 

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/915.

### DIFF
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -12,10 +12,10 @@
   "downloadPDF": "Download PDF.",
   "downloadInstead": "Try <1>downloading</1> it instead.",
   "cantBePreviewed": "Sorry, this file can't be previewed",
-  "addFile": "Add file",
-  "addFolder": "Add folder",
-  "addByPath": "Add by path",
-  "addToIPFS": "Add to IPFS",
+  "addFile": "File",
+  "addFolder": "Folder",
+  "addByPath": "From IPFS",
+  "addToIPFS": "Add",
   "newFolder": "New folder",
   "actions": {
     "add": "Add",
@@ -52,7 +52,7 @@
     "descriptionFile": "{count, plural, one {Are you sure you want to delete this file? This action is permanent and cannot be reversed.} other {Are you sure you want to delete {count} files? This action is permanent and cannot be reversed.}}"
   },
   "addByPathModal": {
-    "title": "Add by path",
+    "title": "Add from IPFS",
     "description": "Insert the path to add."
   },
   "newFolderModal": {
--- a/src/files/FilesPage.js
+++ b/src/files/FilesPage.js
@@ -5,7 +5,6 @@ import { connect } from 'redux-bundler-react'
 import downloadFile from './download-file'
 import { join } from 'path'
 import { translate, Trans } from 'react-i18next'
-
 // Components
 import Breadcrumbs from './breadcrumbs/Breadcrumbs'
 import FilesList from './files-list/FilesList'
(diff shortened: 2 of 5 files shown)


src/files/file-input/FileInput.js
```diff
@@ -20,7 +19,7 @@ const AddButton = translate('files')(({ progress = null, t, tReady, ...props })
   }, ['Button f6 relative transition-all sans-serif dib v-mid fw5 nowrap lh-copy bn br1 pa2 focus-outline'])
 
   return (
-    <button disabled={sending} className={cls} style={{ width: '144px' }} {...props}>
+    <button disabled={sending} className={cls} style={{ width: '120px' }} {...props}>
```
olizilla: huh!? this never used our <Button> component. Not related to your PR @fsdiogo but could you take a look at  swapping this out for <Button> , to make things more consistent.

src/files/FilesPage.js
```diff
@@ -214,12 +230,17 @@ class FilesPage extends React.Component {
...
 
               { files.type === 'directory'
-                ? <FileInput
-                  className='ml-auto'
-                  onMakeDir={this.makeDir}
-                  onAddFiles={this.add}
-                  onAddByPath={this.addByPath}
-                  addProgress={writeFilesProgress} />
+                ? <div className='ml-auto flex items-center'>
+                  <span
```
olizilla: This would be better as a `<Button>` with a `bg='bg-transparent'`, so that we get hover styles and focus. 

We need to improve our a11y reporting too, as this'd be difficult for assistive technologies to spot as clickable.

fsdiogo: Done. Although we'll need to refactor the `<Button>` component, as some classes end up not being overridden, e.g. the `fw` one!

src/files/FilesPage.js
```diff
@@ -231,11 +232,13 @@ class FilesPage extends React.Component {
...
-                  <span
-                    className='mr3 flex items-center f6 gray pointer'
+                  <Button
+                    className='mr1 flex items-center f6 fw3 pointer'
+                    color='charcoal-muted'
+                    bg='bg-transparent'
                     onClick={() => this.showNewFolderModal()}>
-                    <FolderIcon className='mr1 fill-gray w2 o-80 glow' />{t('newFolder')}
-                  </span>
+                    <FolderIcon viewBox='10 15 80 80' height='20px' className='fill-charcoal-muted w2' />{t('newFolder')}
```
olizilla: @fsdiogo is there a problem with the positioning of this icon that we should fix in the svg?

olizilla:
I tweaked the AddButton to remove the i18n and lng props as they were being passed through to the dom element. I tweaked the new file button to use v-mid on the svg rather than flex, and wrapped the text in its own span with a `fw3`

src/files/FilesPage.js
fsdiogo: I don't think so, I just tweaked it a bit so it didn't make its parent container grow!

public/locales/en/files.json
```diff
@@ -12,10 +12,10 @@
   "downloadPDF": "Download PDF.",
   "downloadInstead": "Try <1>downloading</1> it instead.",
   "cantBePreviewed": "Sorry, this file can't be previewed",
-  "addFile": "Add file",
-  "addFolder": "Add folder",
-  "addByPath": "Add by path",
-  "addToIPFS": "Add to IPFS",
+  "addFile": "File",
+  "addFolder": "Folder",
```
lidel: Perhaps 

> Folder with files

?

public/locales/en/files.json
```diff
@@ -12,10 +12,10 @@
...
+  "addByPath": "From IPFS",
```
lidel: Perhaps more explicit styling of "IPFS" as a path would make it more clear:

> From `/ipfs/…`

terichadbourne:
A couple of suggestions/observations made on our call today: 

- Under the option to add from IPFS, perhaps we could  reference options  more specific than "path," like "CID" or "IPNS path," and provide examples of acceptable formats.
- For me personally, "Add" doesn't imply that the thing you're adding is pre-existing, whereas "Upload" or "Import" do (I know you have technical objections to those terms), so leaving it verbless doesn't help me know that you're referring to existing files/folders instead of blank ones

fsdiogo:
@lidel 	@terichadbourne what do you think:

from-ipfs

terichadbourne:
@fsdiogo Yes, that’s much easier for me to understand!

Issue #889, opened 2018-11-26
Spacebar selection does not work in Firefox


lidel:
In `Files` screen we can use keyboard arrow keys to move between files.
Pressing <kbd>Spacebar</kbd> adds file to selection (checkbox), but it only works in Chrome.
In Firefox checkbox gets selected and then instantly unchecked.

(keyboard navigation is important for Accessibility https://github.com/ipfs/ipfs-gui/issues/28)

PR #957, opened 2019-02-12
feat: add react-virtualized to the files list

fsdiogo:
- Using `react-virtualized` to only render the files in view
- Loading animation when listing takes more than 1s to load
- Dir sizes only show if current dir has < 100 files (fix in an incoming PR)


olizilla:
@fsdiogo I'm seeing the firefox rendering issue again on this branch

screenshot 2019-02-13 at 09 00 24

fsdiogo:
Oh noes! With the whole `<ContextMenu>` refactor I forgot to apply the Firefox fix again.

I'll add that fix to master and then sync this branch.

src/files/files-list/FilesList.js
```diff
@@ -33,6 +36,7 @@ export class FilesList extends React.Component {
     updateSorting: PropTypes.func.isRequired,
     root: PropTypes.string.isRequired,
     downloadProgress: PropTypes.number,
+    filesIsFetching: PropTypes.boolean,
```
olizilla:
```suggestion
    filesIsFetching: PropTypes.bool,
```
My bad, i think i added that.

olizilla:
The "select and it immediately deselect" problem has re-appeared on firefox.
- Use keyboard nav to highlight an item
- hit spacebar to select
- checkbox is checked and then unchecked again
- _expected_ checkbox remains checked

see: https://github.com/ipfs-shipyard/ipfs-webui/issues/889

src/bundles/files.js
```diff
@@ -390,6 +378,11 @@ export default (opts = {}) => {
 
     selectFilesIsFetching: (state) => state.files.pending.some(a => a.type === actions.FETCH),
 
+    selectShowLoadingAnimation: (state) => {
+      const pending = state.files.pending.find(a => a.type === actions.FETCH)
+      return pending ? (Date.now() - pending.start) > 1000 : false
```
olizilla: of note, this selector will only be re-run if something happens, like an `APP_IDLE` event, or the user interacts with the app. That's why it will sometimes take a while to see the loading animaton... We're going to revisit this in a future PR so no need to change it now.

src/files/file-icon/FileIcon.js
```diff
@@ -11,23 +11,23 @@ import DocText from '../../icons/GlyphDocText'
 
 const style = { width: 36 }
 
-export default function FileIcon ({ name, type }) {
+export default function FileIcon ({ name, type, cls }) {
```
olizilla: give `cls` a default value, ot `undefined` appears in the class list.

```suggestion
export default function FileIcon ({ name, type, cls = '' }) {
```

src/files/files-list/FilesList.js
```diff
@@ -212,31 +196,39 @@ export class FilesList extends React.Component {
...
           name = '..'
         } else {
-          name = this.props.files[index].name
+          name = files[index].name
+        }
+
+        // If the file we are going to focus is out of view (removed
+        // from the DOM by react-virtualized), focus the first visible file
+        if (!this.filesRefs[name]) {
+          name = files[firstVisibleRow].name
```
olizilla: Perhaps one for a future PR, but I think we can

- use `scrollToCell` in react virtulized 
  - https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md#scrolltocell--columnindex-number-rowindex-number-
- store `focused` as its array index not its file name.
- remove `this.fileRefs` entirely. It slowly fills up with a dom node for all items, which defeats some of the value of using react-virtualized.

PR #965, opened 2019-02-21
chore: add circleci based deploy

olizilla:
ci is dead. long live ci.

License: MIT


.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+            hash=$(ipfs-cluster-ctl \
+              --host /dnsaddr/cluster.ipfs.io \
+              --basic-auth $CLUSTER_USER:$CLUSTER_PASSWORD \
+              add --rmin 3 --rmax 3 --name "$DOMAIN" \
+              --recursive /tmp/workspace/$BUILD_DIR | tail -n1 | cut -d " " -f 2 )
+
+            preview_url=https://ipfs.io/ipfs/$hash
+
+            # Preload ipfs hash on the gateway
+            curl --silent --output /dev/null --show-error $preview_url
```
lidel: Was the intention to preload just the HTML of main page?
If not, entire thing could be preloaded asynchronously in best-effort fashion without breaking the build:

```bash
nohup curl --max-time 900 -s "https://ipfs.io/api/v0/refs?r=true&arg=${$preview_url}" >/dev/null 2>&1 &
```

.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+      - persist_to_workspace:
+          root: .
+          paths:
+            - build
+
+  deploy:
+    docker:
+      - image: olizilla/ipfs-dns-deploy
+        environment:
+          DOMAIN: explore.ipld.io
```
fsdiogo: `webui.ipfs.io`?

olizilla: derp

.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
...
+      - attach_workspace:
+          at: /tmp/workspace
+      - run:
+          name: Pin website, post notification for PRs or update DNS on master
+          command: |
+            # Upload build dir to cluster.ipfs.io
+            hash=$(ipfs-cluster-ctl \
+              --host /dnsaddr/cluster.ipfs.io \
+              --basic-auth $CLUSTER_USER:$CLUSTER_PASSWORD \
+              add --rmin 3 --rmax 3 --name "$DOMAIN" \
```
hsanjuan: Do not set --rmin/--rmax, that way cluster's defaults will be used.. Also, if by chance you have the [short]-commit ID available, you could add it to the `--name` tag so there is more info.

.circleci/config.yml
```diff
@@ -0,0 +1,68 @@
...
+              --recursive /tmp/workspace/$BUILD_DIR | tail -n1 | cut -d " " -f 2 )
```
hsanjuan: You want to use the `--quietest` or `--Q` option which only outputs the final hash, rather than `tail ...`

olizilla:
Thanks @hsanjuan that's good to know. Will update.

PR #967, opened 2019-02-22
fix: tweak cluster params for ci build

olizilla:
see: https://github.com/ipfs-shipyard/ipfs-webui/pull/965#pullrequestreview-206605757

License: MIT

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,10 +28,11 @@ jobs:
           command: |
             # Upload build dir to cluster.ipfs.io
             hash=$(ipfs-cluster-ctl \
+              --quietest \
               --host /dnsaddr/cluster.ipfs.io \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -28,10 +28,11 @@ jobs:
           command: |
             # Upload build dir to cluster.ipfs.io
             hash=$(ipfs-cluster-ctl \
+              --quietest \
```
olizilla: @hsanjuan using `--quietest` on ipfs-cluster-ctl here causes

```console
Incorrect Usage. flag provided but not defined: -quietest
```

did i use it wrong?

hsanjuan: Sorry ` --quieter, -Q                          Write only final hash to output`

PR #969, opened 2019-02-22
feat: add connection

### LINKED ISSUE #891: Peers: UI  for swarm connect
Here's an idea:  "Add Connection" button

- could be a button similar to "Add to IPFS" from  Files screen
  - benefit from reusing established visual pattern
- opens form that accepts multiaddr, when confirmed executes `ipfs swarm connect`
- open questions:
  - how to indicate success/failure?
    - somehow store and "pin" peerid of manually added connection(s) to the top of Peer list?
    - display error message instead of peerid line if failed?
  - support "pet names" ? could be first step towards [ipfs-friends](https://github.com/tableflip/ipfs-friends)-like feature

hacdias:
Closes #891.

As a first iteration this is what I've done:

image

image

It works, but we don't provide feedback if it worked or not. What do you think?

/cc @lidel @olizilla @fsdiogo

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "C_Cpp.dimInactiveRegions": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "C_Cpp.dimInactiveRegions": false
-}
--- a/public/locales/en/notify.json
+++ b/public/locales/en/notify.json
@@ -5,5 +5,7 @@
   "windowIpfsRequestFailed": "IPFS request failed. Please check your IPFS Companion settings.",
   "ipfsIsBack": "Normal IPFS service has resumed. Enjoy!",
   "filesEventFailed": "Failed to add files to IPFS. Please try again.",
-  "folderExists": "A folder with that name already exists. Please choose another."
+  "folderExists": "A folder with that name already exists. Please choose another.",
+  "couldntConnectToSwarm": "Could not connect to the swarm.",
+  "connectedToSwarm": "Successfully connected to the swarm."
 }
--- a/public/locales/en/notify.json
+++ b/public/locales/en/notify.json
@@ -6,6 +6,6 @@
   "ipfsIsBack": "Normal IPFS service has resumed. Enjoy!",
   "filesEventFailed": "Failed to add files to IPFS. Please try again.",
   "folderExists": "A folder with that name already exists. Please choose another.",
-  "couldntConnectToSwarm": "Could not connect to the swarm.",
-  "connectedToSwarm": "Successfully connected to the swarm."
+  "couldntConnectToPeer": "Could not connect to the provided peer.",
+  "connectedToPeer": "Successfully connected to the provided peer."
 }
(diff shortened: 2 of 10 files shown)


lidel:
Cool! Quick feedback: we should provide a hint to the user that address should be in multiaddr format, similar to what we already do on Files → Add from Path:

>

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "C_Cpp.dimInactiveRegions": false
```
fsdiogo: Remove this file. We should add the `.vscode` folder to the `.gitignore`.

fsdiogo:
> It works, but we don't provide feedback if it worked or not.

We should provide feedback! Why not use the notify bundle as:


notify

A message saying `Successfully connected to the swarm` or something like that is sufficient IMO!

hacdias:
image

image

image

hacdias:
I think it's ready for review!

public/locales/en/notify.json
```diff
@@ -5,5 +5,7 @@
   "windowIpfsRequestFailed": "IPFS request failed. Please check your IPFS Companion settings.",
   "ipfsIsBack": "Normal IPFS service has resumed. Enjoy!",
   "filesEventFailed": "Failed to add files to IPFS. Please try again.",
-  "folderExists": "A folder with that name already exists. Please choose another."
+  "folderExists": "A folder with that name already exists. Please choose another.",
+  "couldntConnectToSwarm": "Could not connect to the swarm.",
```
lidel: Probably need to change this language a bit:
User is probably connected to the swarm already and is just trying to connect to one of members of the swarm, not THE swarm (that happens during bootstrap).

hacdias: Done!

fsdiogo:
The multiaddr validation in the modal input isn't totally safe, as you can pass an incorrect CID and it will still be accepted, although it will fail to connect.

E.g.: `/ip4/192.168.2.92/tcp/4001/ipfs/foobar`

hacdias:
@fsdiogo yeah we can improve that later with https://github.com/ipfs/is-ipfs/issues/26 perhaps.

hacdias:
@olizilla can we merge?

src/peers/AddConnection/AddConnection.js
```diff
@@ -0,0 +1,73 @@
...
+      open: !this.state.open
+    })
+  }
+
+  addConnection = (maddr) => {
+    this.props.doConnectSwarm(maddr)
+    this.toggleModal()
+  }
+
+  validate = (maddr) => {
```
lidel: @hacdias can we switch this to `peerMultiaddr` check from  [is-ipfs v0.6.0](https://github.com/ipfs/is-ipfs/releases/tag/v0.6.0)? 
It gives better validation (confirms the multiaddr represents IPFS peer). 

> ### `isIPFS.peerMultiaddr(addr)`
> 
> Returns `true` if the provided `string`, [`Multiaddr`](https://github.com/multiformats/js-multiaddr) or `Buffer` represents a valid "IPFS Peer" multiaddr (matching [`IPFS` format from `mafmt`](https://github.com/multiformats/js-mafmt#api)) or `false` otherwise.

hacdias: Done!!! 😄

olizilla:
I'd like to keep this as a feature for the 2.5 release, while we are still ironing the kinks out of the analytics stuff for 2.4, that's why i've not reviewed it yet.

PR #972, opened 2019-02-23
fix: don't print out expected error

### LINKED ISSUE #966: Failed to get url from Addresses.API TypeError: "t is null"
**Describe the bug**
This is showing up in the logs:

```console
Failed to get url from Addresses.API TypeError: "t is null"
```

It's not fatal, but it shouldnt be there.

**To Reproduce**

1. https://ipfs.io/ipfs/QmZTMSPS7Wk1N6yUFgMRdeXxhQc2uAeLPpQZbn7n2B7VBk/#/welcome
2. Check the dev console

screenshot 2019-02-21 at 17 50 55

hacdias:
Fixes #966 by not printing the error. That's an expected error in that function. We could also make the function receive an `expectsFailure` boolean variable which would tell whether or not to print the error.

License: MIT

### DIFF
--- a/src/bundles/config.js
+++ b/src/bundles/config.js
@@ -57,13 +57,13 @@ bundle.reactConfigFetch = createSelector(
 )
 
 function getURLFromAddress (name, config) {
+  if (!config) return null
+
   try {
     const address = config.Addresses[name]
     return toUri(address).replace('tcp://', 'http://')
   } catch (error) {
-    if (config) {
-      console.log(`Failed to get url from Addresses.${name}`, error)
-    }
+    console.log(`Failed to get url from Addresses.${name}`, error)
     return null
   }
 }
--- a/src/bundles/config.js
+++ b/src/bundles/config.js
@@ -60,7 +60,10 @@ function getURLFromAddress (name, config) {
   try {
     const address = config.Addresses[name]
     return toUri(address).replace('tcp://', 'http://')
-  } catch (_) {
+  } catch (error) {
+    if (config) {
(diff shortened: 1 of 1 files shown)


src/bundles/config.js
```diff
@@ -60,8 +60,7 @@ function getURLFromAddress (name, config) {
   try {
     const address = config.Addresses[name]
     return toUri(address).replace('tcp://', 'http://')
-  } catch (error) {
-    console.log(`Failed to get url from Addresses.${name}`, error)
+  } catch (_) {
```
olizilla: This smothers all errors, which we don't want to do. The error we can ignore is where `config` is undefined or null.

hacdias:
Fixed @olizilla!

olizilla:
I don't *want* to be obsessive (_clearly i am _) about such a tiny change, but this feels like a suprising way to fix this error... As a general pattern that I've got ingrained in my head, always handle the edge cases up front, before letting things flow to the rest of the logic.

```js
function getURLFromAddress (name, config) {
  if (!config) return null
  try {
    const address = config.Addresses[name]
    return toUri(address).replace('tcp://', 'http://')
  } catch (error) {
      console.log(`Failed to get url from Addresses.${name}`, error)
      return null
  }
}
```

hacdias:
@olizilla yeah, that makes sense! Changed. Should've thought about that first

olizilla:
@hacdias pls think about the what you'd  like to read in a change log when you write your PR titles.

`fix: don't print out expected error` won't mean anything to a reader looking at the release notes.

PR #973, opened 2019-02-24
chore: move ci to circle

olizilla:
We don't need to run our unit and e2e tests on travis _and_ circle. We do have to do our deploys on circle. So move all the CI to circle.

Switching the base image for our build step to the circle node image "+browsers", ment getting puppeteer running was pretty painless, so I'm happy to move it all over to circle.

License: MIT


olizilla:
For context... we use circleci for deployment to simplify secrets management. travis and github actions need you to configure secrets per project. circle lets you configure them per org. infra would rather we didn't scatter our secrets across all the projects so we can roll our tokens regularly.

olizilla:
@lidel @fsdiogo I'm gonna merge this, as I don't see any reason not to. we can always revive the travis build, but I don't see the need for it right now.

PR #974, opened 2019-02-25
fix: ensure we have a full api available

olizilla:
- fixes issue where a readonly api would fool us into thinking we have a full api. If you look at https://webui.ipfs.io right now, you'll find it can't find your local ipfs daemon, as it finds the one provided by the domain first.

License: MIT


olizilla:
of note, it might be worth checking for a local ipfs before checking for a remote one, but I can imagine there is no "right answer" for that. But it seems ideal to use a local one where available.

Issue #975, opened 2019-02-25
Natual sorting for file names


olizilla:
**Is your feature request related to a problem? Please describe.**

ascii sorting is terrible
![screenshot 2019-02-25 at 12 29 19](https://user-images.githubusercontent.com/58871/53337483-773cdb80-38f9-11e9-9851-659550525d8b.png)

**Describe the solution you'd like**
natuaral sorting using localeCompare as per https://github.com/ipfs-shipyard/ipfs-webui/pull/909

PR #977, opened 2019-02-25
fix: use localeCompare to sort files

fsdiogo:
Fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/975.

### DIFF
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -21,18 +21,6 @@ export const sorts = {
   BY_SIZE: 'size'
 }
 
-function compare (a, b, asc) {
-  const strings = typeof a === 'string' && typeof b === 'string'
-
-  if (strings ? a.toLowerCase() > b.toLowerCase() : a > b) {
-    return asc ? 1 : -1
-  } else if (strings ? a.toLowerCase() < b.toLowerCase() : a < b) {
-    return asc ? -1 : 1
-  } else {
-    return 0
-  }
-}
-
 const make = (basename, action) => (...args) => async (args2) => {
   const id = Symbol(basename)
   const { dispatch, getIpfs, store } = args2
@@ -361,9 +349,23 @@ export default (opts = {}) => {
         content: pageContent.content.sort((a, b) => {
           if (a.type === b.type || isMac) {
             if (sorting.by === sorts.BY_NAME) {
-              return compare(a.name, b.name, sorting.asc)
+              a = a.name
+              b = b.name
+
+              return sorting.asc
+                ? a.localeCompare(b, undefined, { numeric: true, sensitivity: 'base' })
+                : b.localeCompare(a, undefined, { numeric: true, sensitivity: 'base' })
             } else {
-              return compare(a.cumulativeSize || a.size, b.cumulativeSize || b.size, sorting.asc)
+              a = a.cumulativeSize || a.size
+              b = b.cumulativeSize || b.size
+
+              if (a > b) {
+                return sorting.asc ? 1 : -1
+              } else if (a < b) {
+                return sorting.asc ? -1 : 1
+              } else {
+                return 0
+              }
             }
           }
 
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -340,6 +340,7 @@ export default (opts = {}) => {
 
     selectFiles: (state) => {
       const { pageContent, sorting } = state.files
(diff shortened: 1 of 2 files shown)


src/bundles/files.js
```diff
@@ -361,9 +349,23 @@ export default (opts = {}) => {
         content: pageContent.content.sort((a, b) => {
           if (a.type === b.type || isMac) {
             if (sorting.by === sorts.BY_NAME) {
-              return compare(a.name, b.name, sorting.asc)
+              a = a.name
+              b = b.name
+
+              return sorting.asc
+                ? a.localeCompare(b, undefined, { numeric: true, sensitivity: 'base' })
```
olizilla: Sorting can be simpler if we give oursleves the helper functions to hide the details... The "one weird trick" to for comparator functions is to pass the sort direction as 1 or -1 so you can just mulitply the result of the comparing a and b by dir. If (a - b) is a postive but dir is -1, you invert the result, which is what you want when changing the sort order.

```js
// some where else in a sort.js lib file

//  returns a natural sort comparator for strings. `dir` is 1 for ascending or -1 for descending
export function sortByName (dir = 1, opts  = { numeric: true, sensitivity: 'base' }) {
   return (a, b) => a.localeCompare(b, undefined, opts) * dir
}

//  returns a numerical sort comparator. `dir` is 1 for ascending or -1 for descending
export function sortBySize (dir  = 1) {
  return (a, b) => (a - b) * dir
}
  
// here at use site
if (sorting.by === sorts.BY_NAME) {
  return sortByName(sorting.asc ? 1 : -1)(a.name, b.name)
} else {
  return sortBySize(sorting.asc ? 1 : -1)(a.cumulativeSize || a.size, b.cumulativeSize || b.size)
}
```

Lets extract this to a helper and reuse it in both locations where we want name based sorting, so future us can see the right way to do it rather than defaulting to ascii-sorting.

olizilla: cc @hacdias do you  know the "one weird trick" when writing comparator functions?

fsdiogo: Thanks, this is so much better! ✨

hacdias: @olizilla wow. Never thought about that!

src/bundles/files.js
```diff
@@ -361,9 +350,9 @@ export default (opts = {}) => {
         content: pageContent.content.sort((a, b) => {
           if (a.type === b.type || isMac) {
             if (sorting.by === sorts.BY_NAME) {
-              return compare(a.name, b.name, sorting.asc)
+              return sortByName(sorting.asc ? 1 : -1)(a.name, b.name)
```
olizilla: the outer function is a comparartor  as well, so it gets called once for every comparison. It'd be reasonanble to construct these comparators ouside the function.

fsdiogo: Is [this](https://github.com/ipfs-shipyard/ipfs-webui/pull/977/commits/87d775027cefa4bc937e552709f92248f2137b41) what you mean?

If you were referring to the `sorting.by`,

```js
const sortDir = sorting.asc ? 1 : -1
const sortByFn = sorting.by === sorts.BY_NAME ? sortByName : sortBySize

return {
	...pageContent,
	content: pageContent.content.sort((a, b) => {
		if (a.type === b.type || isMac) {
			// the sorting fns have diff params
			// sortByName(sortDir)(a.name, b.name)
			// sortBySize(sortDir)(a.cumulativeSize || a.size, b.cumulativeSize || b.size)
			sortByFn(sortDir)(...)
		}
	})
}
```

olizilla: not quite. we know the direction before we get to `pageContent.content.sort` so we can build the `sortBySize(sortDir)` and `sortByName(sortDir)` comparators before hand.

fsdiogo: Sorry, but I'm not sure I follow 🤔

olizilla: calling `sortByName(dir)` creates and returns a function, We can create that function outside of the comparator, so we dont re-create it once for every pair of items. we just create it once before we start the sort.

olizilla:
```js
const nameSort = sortByName(dir)

//...

return {
	...pageContent,
	content: pageContent.content.sort((a, b) => {
		if (a.type === b.type || isMac) {
		     nameSort(a.name, b.name)
  // ...
		}
	})
}

Issue #980, opened 2019-02-26
Analytics in Go-IPFS


Stebalien:
Go-IPFS analytics requirements:

1. Opt-in. We can prompt the user on first setup, possibly even modify their config, maybe even skip the dialog if the user has do-not-track set. However, it needs to be opt-in.
2. Minimal. We don't need things like City/Country. We should only collect information we _need_ to improve our software and wouldn't in any way identify our users if it were to leak.
3. We need to clearly define what we collect, ideally displaying a sample.

If you'd like a good example of how to do this, please check out how syncthing handles analytics.

magik6k:
> through IPFS/libp2p.

https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md#p2p-http-proxy?

olizilla:
For context for people not following along: the current situation is described in #930

> Opt-in. We can prompt the user on first setup, possibly even modify their config, maybe even skip the dialog if the user has do-not-track set. However, it needs to be opt-in.

Agreed. It is simplest from where we are to just disable by default and ask the user.

>    Minimal. We don't need things like City/Country. We should only collect information we _need_ to improve our software and wouldn't in any way identify our users if it were to leak.

Agreed.

> We need to clearly define what we collect, ideally displaying a sample.

This what we have in place so far

![](https://user-images.githubusercontent.com/58871/52051030-43fd6b80-254a-11e9-8789-5e1a2aea40b6.png)
https://github.com/ipfs-shipyard/ipfs-webui/pull/930#issuecomment-459309214

Stebalien:
Ideally, we'd also do as much event aggregation locally instead of sending events to the server (e.g., "the user uses X/Y/Z pages instead of "the user visited X/Y/Z pages at time T"). We don't need (or want) to know when a user is using the WebUI, we just want to know which features are useful, etc.

For context, my syncthing instance reports:

```
{
  "alwaysLocalNets": false,
  "announce": {
    "defaultServersDNS": 1,
    "defaultServersIP": 0,
    "globalEnabled": true,
    "localEnabled": true,
    "otherServers": 0
  },
  "blockStats": {},
  "cacheIgnoredFiles": false,
  "customDefaultFolderPath": false,
  "customReleaseURL": true,
  "customTempIndexMinBlocks": false,
  "customTrafficClass": false,
  "deviceUses": {
    "compressAlways": 0,
    "compressMetadata": 3,
    "compressNever": 0,
[...]

PR #981, opened 2019-02-26
fix: make analytics opt-in

fsdiogo:
I think this deserves a proper refactor instead of a hotfix like this one.


olizilla:
Agreed. I think we an just remove the doNotTrack logic. It simplifies things.

fsdiogo:
@olizilla this is ready for review.

TL;DR: I removed all the logic related to the `doNotTrack` setting, so analytics are disabled by default. I kept the banner asking to enable analytics when no choice has been made, after that it disappears.

olizilla:
Ah, hold on, there is more here https://github.com/ipfs-shipyard/ipfs-webui/issues/980#issuecomment-467806732

I'll take this PR on @fsdiogo

olizilla:
I'm gonna merge this now, so we get opt-in deployed to webui.ipfs.io now, and open an new PR for the rest.

Issue #979, opened 2019-02-26
Unobtrusive help system to onboard new users


olizilla:
> **unobtrusive** | ʌnəbˈtruːsɪv |
> _adjective_
> not conspicuous or attracting attention: _the service was unobtrusive and efficient._

### Is being inconspicuous an important goal?

Perhaps not directly. I think the main point was that help should be easily available for those that want it and shouldn't distract those who dont want it.

Other valuable goals would be
- Maintainble, and always in sync with the current UI
- Localizable, via transifex and https://github.com/ipfs/i18n
- Legible and accessible
- Linked to proto.school work and IPFS documentation endevours
- Crowdsource /  Allow folks to add their own annotations (discussify?)

### What does an unobtrusive help system look like? 

Your thoughts here!

olizilla:
Maybe https://introjs.com/ ?

lidel:
I'm for it, given the educational mission of UIs  maintained by IPFS project, we should aim to provide contextual help using the same visual language across all UIs.

I like the idea of something like [introjs/hint](https://introjs.com/example/hint/index.html) for one-time onboarding coupled with persistent icon next to various values in UI that may need explaining. Both should be using the same visual indicator, perhaps something like: <img src="https://user-images.githubusercontent.com/157609/53423098-04a72b00-39e1-11e9-856e-c326e38ca5e1.png" width="16px" height="16px"/> <img src="https://user-images.githubusercontent.com/157609/53423262-56e84c00-39e1-11e9-9b14-1ca0196942e7.png" width="16px" height="16px"/> <img src="https://user-images.githubusercontent.com/157609/53423263-56e84c00-39e1-11e9-807f-fe362a30ca84.png" width="16px" height="16px"/> 


If we want it to be unobtrusive, the first step would be to come up with convention for linking to relevant docs from UIs. Right now we can link to docs but it will look like any other link. To make it legible and accessible  `cursor: help`  will not be enough, we need a consistent hint on top of that, eg. mentioned indicator icon:

> _Some Vague Technical Term_<sup><img src="https://user-images.githubusercontent.com/157609/53423263-56e84c00-39e1-11e9-807f-fe362a30ca84.png" width="16px" height="16px"/> </sup>

olizilla:
A few folks have expressed a dislike of introjs style "guided" intros as a help system.

[...]

PR #983, opened 2019-03-06
feat: help system

fsdiogo:
Puts a help system in place with guided tours using [`react-joyride`](https://react-joyride.com/).

Preview (don't mind the button colors, some got degraded by converting to gif):

joyride

TODOs:

- [x] Merge https://github.com/ipfs-shipyard/ipld-explorer-components/pull/17

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/979.


fsdiogo:
I'm thinking about how we should enable/disable the guided tours.

We could have a checkbox in `Settings` like we have for the analytics:

> settings

Another option is to have something in the header to enable the tour on that page:

> header

Any one got more suggestions?

I was thinking of enabling the tours the first time a user goes to a certain page, after finishing the tour they wouldn't appear anymore unless the user commands it.


cc @ipfs-shipyard/gui

ericronne:
That's slick, @fsdiogo!

Since the tool has a number of pages, i wonder if it might be too much for users if the modal appeared for all of them on first visit.

Let me know what you think about this as an unobtrusive-yet-findable motif for launching the tour? It means adding a help button, which could later hold expanded content.
image

Something to test with users?

fsdiogo:
> Since the tool has a number of pages, i wonder if it might be too much for users if the modal appeared for all of them on first visit.

I have to agree with you 👍

> Let me know what you think about this as an unobtrusive-yet-findable motif for launching the tour? It means adding a help button, which could later hold expanded content.

Yeah, in my previous comment I thought about that, my only concern is that it won't be that much visible to new users. That tooltip could help, would it be activated on hover or could it be visible the first time the user enters that page maybe?

ericronne:
My gut says to show it only once by default, on whatever page they initially visit. Once they click the `?` button or the close `x`, never show it again. Even if they close it without clicking the `?`, they now know where to find help at any time. Also the button is in a somewhat standard "utility region" of the screen, so it should be easy to find.

**But** my gut is often wrong! In fact, i just armchair tested my design with one of my designated "all-purpose users" here at my home office, and they didn't notice the dialog at all. Too subtle, suffers from banner blindness, plus it could be taken for a search box if seen out of the corner of your eye.

Here's an alternate design, which is harder to miss. LMK what you think! …

image

And here it is in the context of the whole page, in case you want to do some gorilla testing of your own  😁 

image

src/components/tour/TourHelper.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import { connect } from 'redux-bundler-react'
+
+export const TourHelper = ({ doEnableTours, size = 23 }) => {
+  const handleClick = () => {
+    doEnableTours()
+  }
+
+  return (
+    <span className='dib mr1 ml4-m pointer' onClick={handleClick}>
```
olizilla: any reason not to pass `doEnabledTours` in directly to the `onClick`

fsdiogo: Nope. I personally always prefer to have a separate function so it's easier to add more logic in the future.

fsdiogo:
Thanks @ericronne, I'll try something with that design, I think it will do the job!

olizilla:
Can merge master in here pls @fsdiogo

fsdiogo:
Added a tooltip to let users know where to find the guided tours:  

tooltip

This only shows once, when the user closes the tooltip a flag is saved on localStorage so it doesn't show again.

ericronne:
Could the tooltip be moved closer to the button, for clarity? The copy i supplied could've been more explicit, too, to ensure users understand. How about: "Click the "?" button any time for a guided tour of the page you're viewing."

fsdiogo:
The tooltip is fluid and automatically positions itself, I don't think I can tweak it that extent.

Hmm, as the tooltip is already pointing to the `?` I think it's redundant to have that in the copy, but I'm not against it!

ericronne:
No worries re positioning. I'm inclined to err on the side of obvious for the text. It could also help those who use screen readers. Thanksir! 🙏

fsdiogo:
@olizilla this is ready for review! 

This https://github.com/ipfs-shipyard/ipld-explorer-components/pull/17 needs to be merged first, if you have any suggestions about the copy of the tours please comment in the PR review 🙏

public/locales/en/explore.json
```diff
@@ -31,5 +31,35 @@
...
   "codec": "codec",
-  "multihash": "multihash"
+  "multihash": "multihash",
+  "tour": {
+    "projects": {
+      "title": "Featured datasets",
+      "paragraph1": "Explore the featured datasets or paste a CID to see how that data is structured and linked across protocols."
+    },
+    "explorer": {
+      "step1": {
```
lidel: It may be useful to include more context in key names, eg `step1_ipldExplorer`. 

This small tweak makes translator's life bit easier and removes any confusion if we decide to change the order in the future.

fsdiogo: I followed the following structure for (almost) every page:

```
// We only have one single tour per page,
// so the steps are directly inside the tour
"tour": {
	"stepX: { ... }
	"stepY: { ... }
}
```

The odd case here is the IPLD Explorer because we have two tours, so I split them:

```
// As we have two tours for I've created another hierarchy 
"tour": {
	// Explore projects page
	"pageX": {
		"stepX: { // steps }
	},
	// Inside the explore page
	"pageY": {
		"stepY: { //steps }
	}
}
```

The reasoning for not being explicit on the steps is because they can change often, so it's easier to just have `step1`, `step2`, `step3` to not have to think about the steps title if we need to move them around or add/remove steps.

lidel: I've looked at how Transifex handles situation when we change `step1` to something totally different than `IPLD Explorer` and it will remove existing translations, but thanks to Shared Translation Memory we have enabled there, if the same source string appears under different key, all translations will automatically be restored there.

**tl;dr** you are right, ignore my comment, it will be fine :-)

olizilla:
Is there a way to tell joyride which side of a component to appear on, the settings info is hidden off the bottom:

Screenshot 2019-05-20 at 14 52 33

fsdiogo:
Yes @olizilla, we can choose the position between: `top`, `right`, `bottom`, `left` and `center`, being the latter positioned to the body of the page, so it won't highlight the component.

hacdias:
@fsdiogo @olizilla @ericronne what's the state of this PR? Can this be rebased and merged?

fsdiogo:
@hacdias from what I remember (this was a long time ago) this was almost ready to go.

What's left to do is merge https://github.com/ipfs-shipyard/ipld-explorer-components/pull/17 first (as the Explore page lives in another repo we have to add the guided tour there), update the dependency here and 🚢 it.

fsdiogo:
I'll rebase this ASAP.

fsdiogo:
This is now synced with `master`. I've also updated the Files tour to match the new features. 

The CI is failing, but I think that with rerun it will pass (I don't have permission to trigger it). 

@hacdias can you take a look at the Files page to see if everything is alright? It was a bit of a pain to merge this 😅

Issue #888, opened 2018-11-20
Upgrade to create-react-app v2


olizilla:
CRA v2 is out https://reactjs.org/blog/2018/10/01/create-react-app-v2.html

It has CSS Modules, postcss, import svgs as component, and new babel, webpack and jest, which'll simplify a few things for us.

fsdiogo:
Just finishing this so I can then solve https://github.com/ipfs-shipyard/ipfs-webui/issues/986.

Issue #986, opened 2019-03-11
Update to Storybook 5


fsdiogo:
Update to the shiny Storybook 5 🌟 

https://medium.com/storybookjs/storybook-5-0-db1d0f9c83b8

olizilla:
Let's make sure we build and publish the storybook somewhere. I could be as simple as `/storybook`

cwaring:
Just a thought; do we want to expose/upgrade storybook in this project if
the priority is to migrate to a consolidated component library?

fsdiogo:
@cwaring I think we do, as we have some whole pages in Storybook to check their state.

The individual components can be extracted to be component library, but I guess that we'll still have some to show in the Web UI storybook.

olizilla:
Let's have both!

PR #987, opened 2019-03-11
feat: update to CRA 2 and Storybook 5

fsdiogo:
Updates to [Create React App 2](https://reactjs.org/blog/2018/10/01/create-react-app-v2.html) and [Storybook 5](https://medium.com/storybookjs/storybook-5-0-db1d0f9c83b8).

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/888 and https://github.com/ipfs-shipyard/ipfs-webui/issues/986.


fsdiogo:
> We should probably still have a lint command, but that could run eslint as it's already available.

💯 agree.

> We should document that this project uses prettier, and encourage people to use an editor plugin for it. Having it format on save is pretty helpful.

I'd prefer to unburden the developers and make prettier optional.

> Should we run prettier before commit? I find any pre-commit hook quite annoying, so maybe not. But then we should double down on having folks run prettier in their editors. Same goes for eslint warning.

True, maybe we should remove prettier altogether and just use eslint.

Quick and easy, what do you think?

olizilla:
@fsdiogo can ya update from master pls

fsdiogo:
Synced with master @olizilla.

olizilla:
package-lock got funky.

I can't start the app in local dev mode from an `npm ci`. am investigate.

```
./node_modules/create-hmac/browser.js
Module not found: Can't resolve 'create-hash/md5' in '/Users/oli/Code/ipfs-shipyard/ipfs-webui/node_modules/create-hmac'
```

fsdiogo:
I used [`npm-merge-driver`](https://github.com/npm/npm-merge-driver) to automatically solve the `package-lock.json` conflicts.

You are right, doing `npm ci` is breaking the dev mode.

olizilla:
@fsdiogo you did the right thing, but it did not do the right thing! Is working now.

fsdiogo:
Nice! Want to shed some light of how you solved it please?

olizilla:
@fsdiogo The error showed that `create-hmac` depends on `create-hash/md5` but looking at the files in my local `node_modules/create-hash` showed no `md5.js` which explained the error... looking at the other files in there it turned out that the entire dep was the **wrong thing**. `create-hash` was being resolved to a totally different dep:`es-to-primitive`... I did this:

```console
$ cat node_modules/create-hash/package.json 
{
    "name": "es-to-primitive",
    "version": "1.2.0",
    "author": "Jordan Harband",
    "description": "ECMAScript "ToPrimitive" algorithm. Provides ES5 and ES2015 versions.",
```

so then i was all like _whuuuuuuut!?_ so i checked the package-lock.json and found this weird nugget

```json
"create-hash": {
     "version": "1.2.0",
     "resolved": "https://registry.npmjs.org/es-to-primitive/-/es-to-primitive-1.2.0.tgz",
```

sooo, that's weird, but it at least explains why npm was dropping `es-to-primitve` in where `create-hash` should be.

the solution, well, no magic here... 

```console
$ rm -rf node_module package-lock.json
$ npm i
```
🎉 

...previously we had to be careful to preserve our package-lock.json when we were depending on an older version of ipld and a specific arrangement of ipld deps, but since we upgraded that, we're safe to re-build the package-lock when things get weird.

fsdiogo:
betterthan

Thanks @olizilla!

Issue #1010, opened 2019-04-11
Uncaught Error: stream.push() after EOF


olizilla:
**Describe the bug**
Uploading a file > 1MB fails, with the error:

```
Uncaught Error: stream.push() after EOF
    at new NodeError (errors-browser.js:30)
    at readableAddChunk (_stream_readable.js:269)
    at PullDuplexStream.push../node_modules/readable-stream/lib/_stream_readable.js.Readable.push (_stream_readable.js:240)
    at next (index.js:55)
    at FileReader.loaded (index.js:22)
```

I think the size is only relevent in so far as it is large enoug that it needs to be split into chunks. I can't recreate the issue with smaller files.

**of note, I do not get this error uploading the same ~1MB flie to the 2.4.4 release that is in the current stable IPFS Dektop 0.7 release.**

**To Reproduce**
1. Using the lastest build on webui.ipfs.io or https://github.com/ipfs-shipyard/ipfs-webui/commit/ee3764d87ea2262007651b733588a0968150bcfc
2. Upload an ~80MB file
3. See an error in the logs. the file is not added to the MFS dir.

**Expected behavior**
File should be added to the dir

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser FF and Chrome

olizilla:
@hugomrdias do you know what changed in `ipfs-http-client` recently that might have triggered this?

olizilla:
Moar stack

  | fn | @ | src |
  |-- | -- | -- | 
  | NodeError | @ | errors-browser.js:30
  | readableAddChunk | @ | _stream_readable.js:269
  | push../node_modules/readable-stream/lib/_stream_readable.js.Readable.push | @ | _stream_readable.js:240
  | next | @ | index.js:55
  | loaded | @ | index.js:22
  | FileReader (async) |   |  
  | (anonymous) | @ | index.js:31
  | drainPull | @ | index.js:46
  | _read | @ | index.js:64
  | push../node_modules/readable-stream/lib/_stream_readable.js.Readable.read | @ | _stream_readable.js:440
  | maybeReadMore_ | @ | _stream_readable.js:551
  | push../node_modules/process/browser.js.Item.run | @ | browser.js:167
  | drainQueue | @ | browser.js:131
  | setTimeout (async) |   |  
  | runTimeout | @ | browser.js:43
  | push../node_modules/process/browser.js.process.nextTick | @ | browser.js:156
[...]

PR #1011, opened 2019-04-11
fix: get drop dir working again

olizilla:
- replace our directory walking code with https://github.com/grabantot/datatransfer-files-promise#readme
- normalise the logic so adding via fileinput and drop both go
through the filesToStreams logic.

License: MIT

### DIFF
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -54,6 +54,7 @@ const make = (basename, action) => (...args) => async (args2) => {
       await store.doUpdateHash(`/files${path}`)
     }
   } catch (error) {
+    console.log(error)
     dispatch({ type: `FILES_${basename}_FAILED`, payload: { id, error } })
   } finally {
     if (basename !== actions.FETCH) {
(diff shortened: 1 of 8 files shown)


hacdias:
Hmmm... I can't seem to be able to drop / 'Add Folder' some specific folders. I get this error:

```
Uncaught Error: stream.push() after EOF
    at new NodeError (errors-browser.js:30)
    at readableAddChunk (_stream_readable.js:269)
    at PullDuplexStream.push../node_modules/readable-stream/lib/_stream_readable.js.Readable.push (_stream_readable.js:240)
    at next (index.js:55)
    at FileReader.loaded (index.js:22)
```

Although, it doesn't seem to be related to the folder adding logic. It's the file the problem. I tried adding individual files and... well... found some 'faulty'. Here's one:

https://www.dropbox.com/s/wph4h3ysv739gm9/test%20file.ai?dl=0

olizilla:
I should have mentioned in the issue, it's hard to test this right now as there is a seperate issue over here https://github.com/ipfs-shipyard/ipfs-webui/issues/1010

olizilla:
Multiple files smaller than 1MB should work

hacdias:
@olizilla oh sorry, didn't see that issue. Other than that, it looks good to me.

src/bundles/files.js
```diff
@@ -245,8 +245,9 @@ export default (opts = {}) => {
       }
     },
 
-    doFilesWrite: make(actions.WRITE, async (ipfs, root, rawFiles, id, { dispatch }) => {
-      const { streams, totalSize } = await filesToStreams(rawFiles)
+    doFilesWrite: make(actions.WRITE, async (ipfs, root, filesOrPromise, id, { dispatch }) => {
+      let files = filesOrPromise.then ? await filesOrPromise : filesOrPromise
```
fsdiogo:
```suggestion
      let files = await filesOrPromise
```

If it's a promise it will wait for its resolution, if it's something else it will return immediately, so no need for this check.

olizilla: ❤️  thanks @fsdiogo, that's neat ❤️

PR #1020, opened 2019-04-24
fix: handle file-drop inside empty dir

olizilla:
if you drop files inside dirs that have zero or fewer items than
fills the vertical height, the drop would land on the app rather
than the file list, which would cause files dropped inside the dir
to be added at the root rather than the dir.

This PR is a quick fix to ensure the drops handled at the app level
add the files to the current dir where we are on the files page.

| Before | After |
|--------|------|
| webui-drop-on-empty-dir-before | webui-drop-on-empty-dir-after

License: MIT

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -35,12 +35,11 @@ export class App extends Component {
 
   addFiles = async (files) => {
     const { doFilesWrite, doUpdateHash, routeInfo } = this.props
-
-    // Add the dropped files to the root
-    doFilesWrite('/', files)
-
+    const isFilesPage = routeInfo.pattern === '/files*'
+    const addAtPath = isFilesPage ? routeInfo.params.path : '/'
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -35,12 +35,11 @@ export class App extends Component {
 
   addFiles = async (files) => {
     const { doFilesWrite, doUpdateHash, routeInfo } = this.props
-
-    // Add the dropped files to the root
-    doFilesWrite('/', files)
-
+    const isFilesPage = routeInfo.pattern === '/files*'
```
fsdiogo: Just out of curiosity, what's the difference between `routeInfo.pattern === '/files*'` and `routeInfo.url.startsWith('/files')`?

olizilla: nothing much, just felt more explicit and fewer characters.

PR #1023, opened 2019-04-28
feat: allow settings even when the daemon is offline

### LINKED ISSUE #1019: IPFS Desktop, Language and Analytics settings available when the node is offline
Right now, IPFS Desktop settings are embedded into the Web UI, on the Settings page. Although, that page is unavailable when the node is offline.

The only settings that should not be available when the node is offline is the IPFS configuration file (json). Both language, analytics and IPFS Desktop settings must be available.

image

What if a user gets to the settings page to change something - such as the language - but can't because he doesn't understand the current language _and_ consequently doesn't understand what's going on and the node is offline? Perhaps an unlikely scenario but it might happen.

Despite of that, those options (desktop, analytics and language) are completely independent from the activeness of the node itself, so they should be available.

hacdias:
This closes #1019 by allowing the Settings page even when IPFS config is not available, or the daemon is offline. It might be useful because the settings page contains much more than IPFS own settings.

Screenshot 2019-04-28 at 09 04 35

License: MIT

### DIFF
--- a/public/locales/en/settings.json
+++ b/public/locales/en/settings.json
@@ -16,6 +16,7 @@
   },
   "fetchingSettings": "Fetching settings...",
   "configApiNotAvailable": "The IPFS config API is not available. Please disable the \"IPFS Companion\" Web Extension and try again.",
+  "ipfsDaemonOffline": "The IPFS daemon is offline. Please turn it on and try again.",
   "settingsUnavailable": "Settings not available. Please check your IPFS daemon is running.",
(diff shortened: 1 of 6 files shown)


hacdias:
I know I shouldn't force push, but I did :O Just trying to trigger the CI with the latest updates. For some reason the tests pass locally but weren't passing in the CI.

olizilla:
@hacdias you should be able to re-trigger the build via the circle-ci web interface. let me know if that's not the case

hacdias:
@olizilla it says I don't have write permissions, which is weird. I believe I could trigger builds before. It passes now though 😄

PR #1024, opened 2019-04-28
fix: work even when bandwidth metrics are disabled

### LINKED ISSUE #1009: /api/v0/stats/bw returns 500 when bandwidth reporter is disabled
**Describe the bug**
Similar to https://github.com/ipfs-shipyard/ipfs-webui/issues/853
API endpoint `bw` returns:
`{"Message":"bandwidth reporter disabled in config","Code":0,"Type":"error"}` and the webui never considers itself connected.
my profile

Calling `ipfs config --json Swarm.DisableBandwidthMetrics false` resolved this.

It would be cool if we could switch on this message and advise the user somehow.

hacdias:
Fixes #1009 and https://github.com/ipfs-shipyard/ipfs-desktop/issues/893 by showing a message to the user when they have the bandwidth metrics disabled. 

Also, removes some duplicated code. We had two bundles doing basically the same thing (fetching bandwidth stats).

Screenshot 2019-04-28 at 12 28 46

License: MIT


public/locales/en/status.json
```diff
@@ -43,5 +43,7 @@
     "yesLabel": "OK",
     "noLabel": "No thanks",
     "detailsLabel": "More info"
-  }
+  },
+  "bandwidthStats": "Bandwidth Stats",
+  "bandwidthStatsDisabled": "You have the bandwidth metrics disabled. You can enable them by typing the command bellow or changing the key <1>Swarm.DisableBandwidthMetrics</1> to <3>false</3> on <5>Settings</5>."
```
olizilla: Needs a note about "then restart the IPFS daemon to apply the change."

Issue #1025, opened 2019-04-30
Show pins, add pins and browse the whole universe


hacdias:
**Is your feature request related to a problem? Please describe.**

No

**Describe the solution you'd like**

It would be interesting if we could show the user pinned pins (yeah!) on the Files section inside the Web UI. I imagine it as special folder with different icon and some explanation.

It'd be a read-only folder and would only be shown to users that have one or more recursive (and direct?) pins. The folder would allow to navigate the pinned directories/files.

It wouldn't, of course, show indirect pins.

**Describe alternatives you've considered**

Use the CLI. I'm thinking about something visual.

**Additional context**

N/A

---

Closes https://github.com/ipfs-shipyard/ipfs-desktop/issues/922.

olizilla:
Yes, I want to show a page above the MFS root, that lists out yout pins, and some interesting CIDs, so you can see things in the blockstore that aren't in the MFS, and parhaps a way to `ipfs.get` an arbitrary CID.

lidel:
> It'd be a read-only folder and would only be shown to users that have one or more recursive (and direct?) pins.

It could support context actions such as "Copy to MFS/Files"  and "Unpin" :)

hacdias:
@olizilla so what would the structure of the root be? I did thought about something like this:

- `MFS`/`Home` (we'd need to give it a good name)
- `Pins` ? or should we list the pins alongside MFS? Or should we list the `pins` + interesting stuff, and the pins would have a tag telling it's a pin and a way to unpin?
- Also add some way to navigate to an arbitrary CID/IPFS/IPNS/path

hacdias:
Some questions for things to decide:  

1. On the root `/`, what should we should besides the MFS?
	1. Should we show our PINS?
	2. Or should we show something else such as `ipfs` and `ipns` as they are the two main namespaces. Then, inside `ipfs`, we'd have the pins.

hacdias:
This might be out of scope and perhaps should be moved to another issue: I noticed the files page is being re-rendered like crazy for no reason. I've also noticed that we could really simplify our code if we kept most of the action code **and** state on the redux bundle. That way, we could connect each component to the specific bits of data they need, simplifying the code. And since it would be a "shared" state, it could help us with functionalities that involve the selected files, for instance.

olizilla:
In general it's better to keep new features and refactors as seperate tasks. Let's make that a seperate issue.

For the "what should we show where" the goals are:

[...]

PR #1027, opened 2019-05-01
feat: navigate files in the wild 🦌

hacdias:
### Requires:

- [x] https://github.com/ipfs-shipyard/ipfs-css/pull/34


hacdias:
So, this PR is kind of working and here are some screenshots of what is is right now. For more background on this PR, please read #1025 first.

So, right now, we can already navigate `/ipfs`, `/ipns` and MFS namespaces.

- On `/ipfs`, the user can see their direct and recursive pins listed.
- On `/ipns`, we can navigate to paths such as `/ipns/ipfs.io`, but we can't go to the root.
- On `/mfs`, we get access to user's "home". We should really change `/mfs` to something like Home or an Home icon.

We also have a top bar, like GitHub's (on the repository) with some statistics: right now we have the number of blocks (clicking leads to nowhere), the number of pins (clicking leads to `/ipfs`) and a "MFS" button. We were thinking about adding the space occupied by the repository _and_ a comparison between the current occupied space and the limit the user has set in the configuration.

Screenshot 2019-06-04 at 08 36 05
Screenshot 2019-06-04 at 08 35 55
Screenshot 2019-06-04 at 08 36 12

I'm pinging you here, @ericronne, so you could give me feedback on this. What do you think we could do to make this look better? Would you change the navigation? If so, why and how? How'd you improve the top/stats bar?

hacdias:
Current status:

image

fsdiogo:
Looking good 🔥 

Needs vertical alignment on the gray box to center the content 📏

ericronne:
@hacdias "home" here feels ambiguous to me. Is it meant to be the label for the overall box (ie there are 44535 blocks and 0 pins in "home")?

image

hacdias:
@ericronne I think the concept is a bit harder than it should be. Basically, what I'm trying to accomplish here is making "home" the same as "mfs"/Mutable File System. That's where you can place your files and edit them like it was a regular file system.

You might have more blocks and pins than those files that are on MFS, because those are not necessarily connected.

While usually you can't modify files in IPFS (such as deleting, renaming, moving, copying...), you can do so in MFS... or Home in this case.

hacdias:
Current state of art:

image

hacdias:
Also, here's a better history than 'Files - IPFS' thousands of times:

image

olizilla:
This is great! I feel uncomfortable using `/  home`  as the breadcrumb for the MFS, as it suggests that MFS paths are prefixed with `/home`. A simple solution would for now would be to drop the initial slash, and use `home` and `pins` as a label or section header to tell the user _"you're in the MFS" section_, or _"you're in the pins section"_. We can show the current ipfs address or mfs path for the directory we're in below that section header.

let's replace "home" with "files" as the MFS is accessed via the `ipfs files` subcommand right  now, so the UI should match that. So we'd have a section header like "Files" when you are looking at the MFS, and a section header of Pins when looking at your pinned files, and potentially a section header of "IPFS" or "IPNS" when looking at a raw `/ipfs/*` or `/ipns/*` address.

---

E.g. Looking at pins list

---

#### Pins

- /ipfs/QmHash1
- /ipfs/QmHash2
- /ipfs/QmHash3

---

E.g. Drilling into a pinned directory

---
#### Pins

` /ipfs/QmHash1`

- List
- of
- Files

---


E.g. Looking at the MFS root

---
#### Files
`/`
`QmHashFor/`

- List
- of
- Files in your MFS root

---

E.g. Looking at an MFS path:

---
#### Files
`/foo/bar`
`QmHashForFooBar`

- List
- of
- Files in foo/bar

---

hacdias:
@olizilla I agree and understand and I can make the necessary changes. Although I wonder how'd someone use `/foo/bar` breadcrumb to go to the root? Would they click on the first slash?

hacdias:
Ping @olizilla

hacdias:
Or we can use the label as a link :o

hacdias:
So just talked with @olizilla and this is what I'm gonna do for now:

- Add a `Files` pill instead of `/home` on the left of the breadcrumbs.
- Use `/files`, `/ipns`, `/ipns` and `/pins` root paths for each thing to be easier for the user to distinguish. It will be better for us in code terms too.
- How do we explain "what are blocks?"

fsdiogo:
This is awesome @hacdias 🙌

Some thoughts about the UI:



We should do some minor tweaks here, as it feels weird the pins and files being clickable, but not the blocks. Do you think it would help adding the "Pin" logo here to go along with the home? And then find a way to add the number of peers there too.



Why the "...More"? Will it be a new feature? If not, we should just keep the `+ Add` but disabled as it is.



Should we add the Pin/Unpin feature in the bottom bar?



I feel the breadcrumbs are way to large, we should try to lower the font-size and/or the font-weight. Why not hide the CID below the breadcrumbs, and only show it on hover?

hacdias:
Hey @fsdiogo! Thanks for the feedback!

> Do you think it would help adding the "Pin" logo here to go along with the home? And then find a way to add the number of peers there too.

I presume you meant pins. Yeah, I personally wouldn't mind. We could even drop the number of pins. But I believe there are a lot of people that like numbers. If we find a good way to have both, it'd perhaps be nice. /cc @olizilla @ericronne 

> Why the "...More"? Will it be a new feature? If not, we should just keep the + Add but disabled as it is.

The `more` is just the current `...`. It works on single files only and some folders. Just try it and you'll see. Basically it shows the context menu for the current page.

> Should we add the Pin/Unpin feature in the bottom bar?

Perhaps... yes!

> I feel the breadcrumbs are way to large, we should try to lower the font-size and/or the font-weight. Why not hide the CID below the breadcrumbs, and only show it on hover?

I can try tomorrow. I agree with the font size reduction. I didn't want to remove to hide the CID, but if we reduce the font size I think we should so we're not giving it the same importance as the breadcrumbs themselves. 😃 /cc @ericronne @olizilla

hacdias:
Just tweaked the breadcrumbs:

localhost_3000_

hacdias:
@olizilla CI is failing due to commit 7b176a95c96ba60a374bf3ca62fd614ea069b9f6. I took a look but I can't figure out which titles are wrong and causing this. They all look good!

olizilla:
- What happens when the breadcrumb text is long, and it takes up more horizontal space than is available?
- Let's remve the label `CID:`, so it's more similar to the files list.
- The breadcrumbs are just a hint, they are not the most important thing, so they should be presented more softly... it's context for the user, but in general, you are clicking around in your files, so you mainly know where you are. The breadcrumbs are a reminder if you forgot, or are coming back to this page later. As a starting point it the breadcrumbs here should match the size, color and position of those in the ipld explore section.

hacdias:
From now on, this happens when the breadcrumbs overflow:

image

(Fades out on the left)

Current state:

image

hacdias:
Ping @olizilla

hacdias:
@olizilla @fsdiogo I'm merging this now (it's been exactly 3 months since this PR was opened!). The tests are passing and everything seems to be working well.

I ask you all to try anything you think that should be tested to make sure if it's working correctly. If so, please let me know and I'll fix it!

PR #1028, opened 2019-05-01
fix: useless list re-renders

hacdias:
Fixes the useless _tons_ of re-renders on file pages. It happened because every time we returned from `selectFiles`, we were sorting at the time and using the spread operator, hence creating a new _different_  object, causing a re-render. Now it just changes when the files are fetched _and_ when the sorting is updated.

Also, fixed the issue where you open the Web UI on files page and it just wouldn't load.

License: MIT

### DIFF
--- a/src/bundles/files.js
+++ b/src/bundles/files.js
@@ -65,6 +65,28 @@ const make = (basename, action) => (...args) => async (args2) => {
   return data
 }
 
+const sortFiles = (files, sorting) => {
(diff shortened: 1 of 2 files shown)


olizilla:
Thanks for spotting this and taking the time to fix it. Improvements like this help make the app more robust, which in turn will make people trust it more, and _want_ to use it.

I wonder how we might create tests for this kind of thing issue...

PR #1029, opened 2019-05-01
refactor: cleanup files page

hacdias:
This declutters the `FilesPage` file a lot.

- Moves the modals to their own directory and the info boxes to their own directory too 5409447
- Cleans up files.extra on preview c2055cc
- Quickly fixes a message f8a3059
- Cleans up context menu 63b4164
- Moves header outside of files page ea6b761

License: MIT

### DIFF
--- a/src/files/context-menu/ContextMenu.js
+++ b/src/files/context-menu/ContextMenu.js
@@ -61,7 +61,7 @@ class ContextMenu extends React.Component {
         <DropdownMenu
           top={-8}
           arrowMarginRight='11px'
-          left='calc(100% - 200px + 0.5rem)'
+          left='calc(100% - 200px)'
           translateX={-translateX}
           translateY={-translateY}
           open={this.props.isOpen}
--- a/package-lock.json
+++ b/package-lock.json
@@ -3486,12 +3486,9 @@
       }
     },
     "@tableflip/react-dropdown": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/@tableflip/react-dropdown/-/react-dropdown-1.2.0.tgz",
-      "integrity": "sha512-yn9vcXvgITQceGHE4pAkuwEMos8H3cyaDfVYQVwP0raAMML5U0DDMOj4CANus78HwWTFgM9vNY8TIUVxkhuBPA==",
-      "requires": {
-        "tachyons": "^4.10.0"
-      }
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@tableflip/react-dropdown/-/react-dropdown-1.3.0.tgz",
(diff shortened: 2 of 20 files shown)


hacdias:
@olizilla @fsdiogo for the best reviewing experience, I think it'd be better to go commit by commit as their subject-focused 😄

olizilla:
@hacdias the context menu modal doesn't close if you click off it, but does on master, can you take a look pls?

hacdias:
@olizilla could you take a look at https://github.com/tableflip/react-dropdown/pull/5 please?

src/files/context-menu/ContextMenu.js
```diff
@@ -61,7 +61,7 @@ class ContextMenu extends React.Component {
         <DropdownMenu
           top={-8}
           arrowMarginRight='11px'
-          left='calc(100% - 200px + 0.5rem)'
```
fsdiogo: Without this, the placement of the ContextMenu in preview mode is wrong.

hacdias: It works here :O

fsdiogo: One minor nitpick, with this PR when you right-click a file to open the context menu, if you then left click on a file/folder it will enter its preview mode or navigate to that folder, keeping the ContextMenu open.

In the live version this doesn't happen, if the ContextMenu is opened and you left-click anywhere, the ContextMenu gets closed, as I think that's the expected behaviour.

hacdias: image

Notice that I changed the place where ContextMenu is, so it probably won't need it now. That's why I removed it.

fsdiogo: > It works here :O

As you can see it's not centered to the middle dot:

> ctx

hacdias: Perhaps I should add like some pixels there to center it, but not that much.

fsdiogo: But if you do it will ruin the alignment when not in preview mode I think (in the files list).

hacdias: @fsdiogo could you take a look now, please? 😃

fsdiogo: Niiice, working good!

hacdias:
Thanks for the review @fsdiogo. Just waiting for the updated react-dropdown. /cc @olizilla  😄

olizilla:
thanks @hacdias ... @tableflip/react-dropdown@1.3.0 is now a thing

hacdias:
@olizilla updated 😄

hacdias:
Ping @olizilla

PR #1030, opened 2019-05-01
refactor: cleans up files.extra on preview

hacdias:
Removes the "pending" TODO on files bundle, by adding three properties to the main object and using it instead.

License: MIT


hacdias:
Decided to put this on #1029 only (each commit is well scoped) ;)

PR #1034, opened 2019-05-13
fix: use electron dir picker on IPFS Desktop

hacdias:
Required for https://github.com/ipfs-shipyard/ipfs-desktop/pull/927.

Simply adds a new parameter for `doFilesWrite` and uses IPFS Desktop handling directories.

License: MIT


src/lib/files.js
```diff
@@ -8,6 +8,10 @@ const ignore = [
 ]
 
 export async function filesToStreams (files) {
+  if (!Array.isArray(files)) {
```
olizilla: The last bug i fixed in file uploading was becuase the logic for dropping files didn't go through this function. 

We need to simplify the file upload path, so there is one common point at which we filter out files that should be ignored, and transform the input into the format expected by ipfs.add.

Is it possible to have the `selectDirectory` do less work, and just have it offer up the data as it would have been if the user had used the browser file input for adding a directory?

olizilla: Specifically, this PR https://github.com/ipfs-shipyard/ipfs-webui/pull/1011/files where the drag-n-drop code was duplicating some of the work that filesToStreams does, but was doing it slightly differently.

hacdias: > Is it possible to have the selectDirectory do less work, and just have it offer up the data as it would have been if the user had used the browser file input for adding a directory?

This function (`filesToStreams`) treats HTML5 File Streams, not Node.js Streams. I can't make `selectDirectory` return HTML5 File objects via a node.js fs function. At least as far as I know.

If you look at the code (https://github.com/ipfs-shipyard/ipfs-desktop/pull/927/files#diff-8c58d7bb18cd5b6faa7ff801849e00acR53), it basically just does native calls.

olizilla: So we need to figure out what the contract of `doFilesWrite` is, as it's dealing with too many edge cases.

We want one place in the code where the input is normalised and made ready for passing it to ipfs, and filtered for any files that shouldn't be there.

in the abstract, we want several specific, well named functions that handle the details of preparing the input from a specific scenario, and and they all convert the inputs into one shape, that is then passed on to one common function which always does the same thing.

We dont want to push many different types of input into one function and use detections at various points to adjust its  behaviour as that gets super fragile...

hacdias: @olizilla made some updates. Please take a look. Now we call filestostreams early on

src/bundles/files.js
```diff
@@ -264,12 +264,11 @@ export default (opts = {}) => {
       }
     },
 
-    doFilesWrite: make(actions.WRITE, async (ipfs, root, filesOrPromise, id, { dispatch }) => {
+    doFilesWrite: make(actions.WRITE, async (ipfs, root, files, id, { dispatch }) => {
       // NOTE: the simpler form `let files = await filesOrPromise`, leaves the
```
olizilla: These comments  refer to the old code

hacdias:
Ping @olizilla

olizilla:
To dig ourselves out of this hole we need to write tests for

- [ ] file-drop at root
  - 1 file
  - multiple files
  - multiple files and system files are filtered out
  - 1 directory
  - 1 directory and system files are filtered out
  - muliple directories (pretty sure dropping multiple dirs has never worked)
- [ ] file-drop on folder
  - 1 file
  - multiple files
  - multiple files and system files are filtered out
  - 1 directory
  - 1 directory and system files are filtered out
  - muliple directories (pretty sure dropping multiple dirs has never worked)
- [ ] file-input
   - 1 file
   - multiple files
   - multiple files and system files are filtered out
  - 1 directory
  - 1 directory and system files are filtered out

and have a way to run them in:

- [ ] firefox
- [ ] chrome
- [ ] ipfs-desktop

I  think in these PR as they are, filtering of system files wont happen if you add files via a file input in ipfs-desktop. But there are so many routes through that it's hard to say from just staring at the code.

olizilla:
@hacdias can you update this PR now the other files fixes are in pls.

hacdias:
@olizilla done!

hacdias:
@olizilla can we get this merged?

PR #1035, opened 2019-05-13
Fix manifest.json content

hazae41:
Fixed app names and colors

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -1,6 +1,6 @@
 {
-  "short_name": "React App",
-  "name": "Create React App Sample",
+  "short_name": "IPFS WebUI",
+  "name": "IPFS Web UI",
   "icons": [
     {
       "src": "favicon.ico",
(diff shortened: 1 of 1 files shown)


public/manifest.json
```diff
@@ -10,6 +10,6 @@
   ],
   "start_url": "./index.html",
   "display": "standalone",
-  "theme_color": "#000000",
-  "background_color": "#ffffff"
+  "theme_color": "#0b3a53",
```
olizilla: I'm not sure where these color values get used, but the defaults set them to black and white, so I'd guess we don't want to set both of these colors to be the same.

hacdias: More info:

- https://developers.google.com/web/fundamentals/web-app-manifest/#background-color
- https://developers.google.com/web/fundamentals/web-app-manifest/#theme-color

I think it's OK to keep the same.

olizilla: Ah, thanks. Good docs!

PR #1039, opened 2019-05-19
English locale – miscellaneous capitalization, grammatical and typographical fixes

jc86035:
A double check would be useful, since I haven't actually seen all of the strings in use in the interface, but in summary, this PR makes some punctuation more consistent, fixes the incorrect "it's" on the home page, introduces Unicode equivalents of ASCII punctuation characters, fixes some proper nouns, and reduces unnecessary random capitalization.


olizilla:
Thank you for this @jc86035 It's super thorough and I agree with all the suggestions. The only concern I have is what will be the impact on folks who have offered up translations... 

@lidel when we change keys in the source file, do translators get notifications? do existing translations get invalidated or are the existing ones still used? Most of these changes should go in anyway, but we could consider not upgrading quote marks or ellipses it it's gonna generate admin for every translation.

lidel:
> when we change keys in the source file, do translators get notifications?

AFAIK they should get "new strings to translate" notification within 24h of merging this PR.

> do existing translations get invalidated or are the existing ones still used? 

Old translations are invalidated, but are not lost, we keep them in shared "Translation Memory", so restoring them is one click:
 
>

public/locales/en/files.json
```diff
@@ -9,48 +9,48 @@
...
   "downloadInstead": "Try <1>downloading</1> it instead.",
-  "cantBePreviewed": "Sorry, this file can't be previewed",
+  "cantBePreviewed": "Sorry, this file can’t be previewed",
   "addFile": "File",
   "addFolder": "Folder",
   "addByPath": "From IPFS",
   "addToIPFS": "Add",
   "newFolder": "New folder",
-  "generating": "Generating...",
+  "generating": "Generating…",
```
lidel: :heart:  ellipsis!

olizilla:
Thank you very much @jc86035 for such thorough improvement.

Issue #814, opened 2018-09-25
Discover and try out experimental IPFS features


olizilla:
The IPFS experimental features that are available should be surfaced clearly in Web UI. As a first pass, add them as a list of toggle-able feature switches in the settings tab.

~The ux of this won't be great initially, as the user will still have to restart the node, which can't be done in the UI https://github.com/ipfs/interface-ipfs-core/issues/366 We can get the feature in place and then push to make restarting via the API a thing.~

As a first pass, this can be hidden for webui use, but visible when used from ipfs-desktop, as there we can do the work to re-start the daemon when the config is changed.

UI proposal from @ericronne

![](https://user-images.githubusercontent.com/9259185/57809520-3b145c80-772b-11e9-9c43-1cd0776bab26.png)

see: https://github.com/ipfs-shipyard/ipfs-desktop/issues/907#issuecomment-492824825

PR #1048, opened 2019-05-27
feat: IPFS experiments UI

cwaring:
CleanShot 2019-05-27 at 16 17 21

This PR adds new functionality to enable/disable IPFS experiments from the settings UI.

Experiments are controlled from an `EXPERIMENTS` constant in `src/bundles/experiments.js`. In order to add a new experiment simply create a new key to assign a namespace and then add the appropriate values. The `key` will also be used to select the correct language strings from the i18n files. Defaults in English have been added.

```js
npm: { // namespace
  action: async (enabled) => null, // default async callback that is executed on toggle with an `enabled` bool that can be used for switching methods if needed
  enabled: false, // default experiment state
  actionUrls: [ // meta urls that are rendered along the bottom of each experiment
    {
      url: 'https://github.com/ipfs-shipyard/npm-on-ipfs', // url for the a tag
      key: 'readMoreUrl' // translation key
    },
    {
      url: 'https://github.com/ipfs-shipyard/npm-on-ipfs/issues',
      key: 'issueUrl'
    },
    {
      url: 'https://github.com/ipfs-shipyard/npm-on-ipfs',
      key: 'feedbackUrl'
    }
  ],
  desktop: false // bool set to true if you only want the experiment to show in IPFS Desktop
}
```
Preflight checklist:
- [x] remove dummy data
- [x] set desktop to true for `npm`
~- [ ] add notifications for activation and error~
- [x] add examples section description copy
- [x] analytics for enable/disable
- [ ] tests?

fixes: ipfs-shipyard/ipfs-webui#814
fixes: ipfs-shipyard/ipfs-desktop#907

ref: ipfs-shipyard/ipfs-desktop#911


cwaring:
Notes: there are a couple of dummy experiments included and `desktop: false` is setup on `npm` so that it is easier to review/test in your browser. I will amend these once we have the action functionality mapped in and we are ready to merge!

cwaring:
@hacdias I'm not sure how I should call out to enable `npm-on-ipfs` could you let me know when you have .a moment and I'll wire it up in the actions section. Thanks 🙌

src/bundles/experiments.js
```diff
@@ -0,0 +1,97 @@
...
+  doToggleAction: (key, enabled) => async ({ dispatch }) => {
+    if (!key) return
+
+    const experiment = EXPERIMENTS[key]
+    const dispatchAction = createAction(ACTIONS.EXP_TOGGLE, key)
+
+    if (typeof experiment.action === 'function') {
+      await Promise.resolve(experiment.action(enabled))
+        .then(res => dispatch(dispatchAction))
+        .catch(err => console.error(err))
```
cwaring: note: if there is a better way to inform the user about errors via a notification or modal we should do so here.

hacdias:
> @hacdias I'm not sure how I should call out to enable `npm-on-ipfs` could you let me know when you have .a moment and I'll wire it up in the actions section. Thanks 🙌

Heeey! You can use `doDesktopSettingsToggle('npmOnIpfs')` 😃 The feature is not yet merged on IPFS Desktop though.

olizilla:
@cwaring I'll review today, but just wanted to say that this is looking really nice!

src/bundles/experiments.js
fsdiogo: You can use the `notify` bundle to trigger notifications.

Check out [this PR](https://github.com/ipfs-shipyard/ipfs-webui/pull/969) from @hacdias to see how to use it.

cwaring: Thanks, I'll take a look!

cwaring: Is there already a dispatch action I can use to display display a toast notification? It's not clear from the example.

src/settings/SettingsPage.js
```diff
@@ -11,6 +11,7 @@ import LanguageSelector from '../components/language-selector/LanguageSelector'
 import AnalyticsToggle from '../components/analytics-toggle/AnalyticsToggle'
 import JsonEditor from './editor/JsonEditor'
 import DesktopSettings from './DesktopSettings'
+import Experiments from '../components/experiments/experimentsPanel'
```
olizilla: nitpick: our component file names are `UpperCamelCased.js`

cwaring: Oops yep, I'll patch that up

src/bundles/experiments.js
olizilla: the pattern at the moment is to centrally manage all notifications from the notifications bundle, like https://github.com/ipfs-shipyard/ipfs-webui/pull/969/files#diff-7f24a40e51bf52d47e7cf3f255a6a6a8

I don't love that pattern as it's a bit harder to reason about, so we may change it, but that's where they go today.

cwaring: I feel it's ok to release this without notifications and revisit in another PR later on

olizilla: 👍

src/bundles/experiments.js
```diff
@@ -0,0 +1,97 @@
...
+    } else {
+      dispatch(dispatchAction)
+    }
+  },
+  selectIsIpfsDesktop: () => !!window.ipfsDesktop,
```
hacdias: This is already available: https://github.com/ipfs-shipyard/ipfs-webui/blob/feat/experiments/src/bundles/ipfs-desktop.js#L4

src/components/experiments/ExperimentsPanel.js
```diff
@@ -0,0 +1,64 @@
...
+        </div>
+      </Box>
+    )
+  } else {
+    return null
+  }
+}
+
+export default connect(
+  'selectIsIpfsDesktop',
```
hacdias: We ask for the value, but it is never used. In which situations are we going to show the experiments UI?

src/bundles/experiments.js
```diff
@@ -0,0 +1,97 @@
+const ACTIONS = {
+  EXP_TOGGLE: 'EXPERIMENTS_EXP_TOGGLE'
+}
+
+// setup experiments
+const EXPERIMENTS = {
```
hacdias: What if `EXPERIMENTS` obtains IPFS Desktop specific experiments from `window.ipfsDesktop.experiments`?

olizilla: Let's stick with the simplicity of configuring all the experiments in one place for now. We can split them later, if we need to.

cwaring: Agree. This is a nice idea but probably unnecessary overhead at this stage. I prefer to have the logic self-contained if possible 👌

src/components/experiments/ExperimentsPanel.js
cwaring: I'll remove this as it's no longer used

src/bundles/experiments.js
cwaring: Good catch-up, this should be removed too as it's leftover from a previous commit.

src/bundles/experiments.js
hacdias: Then, we need to _know_ when to show them, it means, if we're inside IPFS Desktop or not.

cwaring: Curious: can you also extend translations from IPFS desktop? Because that would be key for this to work.

hacdias: I see we have a `desktop` variable. Although, the user might not support that experiment due to other reasons, such as the ones I've mentioned above

cwaring: See my PR description above, you can use the `desktop: bool` flag for this and the reducer filters out options based on the context :)

hacdias: Actually the `desktop` variable from the example on the first post is **not** used.

hacdias: @cwaring gotcha!

hacdias:
@cwaring I think 'Read more', 'Open an issue' and 'Leave feedback' buttons should open in a different tab

src/bundles/experiments.js
```diff
@@ -0,0 +1,97 @@
...
+  npm: {
+    action: async () => null,
```
hacdias: `async () => doDesktopSettingsToggle('npmOnIpfs')` should do the trick

public/locales/en/settings.json
```diff
@@ -63,6 +63,17 @@
...
   },
+  "Experiments": {
+    "description": "Here you can get an early preview into new IPFS features by enabling options below. We are testing these ideas, and they may not be perfect yet so we'd love to hear your feedback.",
+    "issueUrl": "Open an issue",
+    "feedbackUrl": "💌Leave feedback",
+    "readMoreUrl": "Read More",
+    "npm": {
+      "title": "NPM-ON-IPFS",
+      "label": "Enable npm-on-ipfs",
+      "description": "Install npm modules through IPFS!"
```
hacdias: Perhaps something a bit more explanatory: `Install npm modules through IPFS with the 'ipfs-npm' command line tool`. I'm not good at writing this descriptions haha

src/bundles/experiments.js
cwaring: Is that method available as a global in the electron app?

olizilla: Actions change state. We shouldn't need to await for an action, and shouldn't expect a useful value back from dispatching one.  We need to think about how we visualise the "we are enabling that action" state. Desktop will need to dispatch a new action once the experiment is enabled.

hacdias: No, `doDesktopSettingsToggle` comes from `ipfs-desktop` bundle and should only be called if `selectIsIpfsDesktop() === true`.

hacdias: @olizilla it already does. `selectDesktopSettings().npmOnIpfs`. It will always be up to date. Just like any other IPFS Desktop settings.

hacdias: @cwaring perhaps we could send the `store` object here too: https://github.com/ipfs-shipyard/ipfs-webui/pull/1048/files/e819b3ada40314ea3f25850fe1397b45bdf0b6a7#diff-74d797d2ce7950ecff72a587d6cf3f9dR86

So the action has access to all bundle actions

cwaring: Yes that is why I was asking because at the moment the actions defined here are simply callback functions that wait for a result before changing the state. It might be a better pattern to hook into the store and despatch something instead. Leaning on your preferred patterns for this so all hints welcome.

hacdias: @cwaring we can then perhaps expand the idea Oli had. @olizilla said we could pass an array of `ids` from `ipfsDesktop` with the experiments to enable. Perhaps it could be an object like this:

```js
{
	id: {
		enabled: true,
		toggle: function () {}
	},
	anotherId: {
		enabled: false,
		toggle: function () {}
	}
}
```

And that object is provided by this: #1054, which you can merge here.

hacdias: @cwaring have got an idea. We can do the following, when `desktop: true`:

```js
// Check if we should show it
store.doDesktopShouldShowExperiment(experimentId) // ==> true or false

// Get experiment value
store.selectDesktopSettings().experiments.experimentId // ==> true or false

// Toggle experiment value
store.doDesktopToggleExperiment(experimentId) // async action, do not wait for it
```

Sounds good? Take that in account and make all the changes you need, I'll make those functions work on #1054 and then merge it here, sounds good?

cwaring: I think we can do something to connect the two here, my concern is that I wouldn't want  to blindly toggle and then update the enabled state in case there is an error so perhaps we need a result method in the reducer here and the toggle action simply dispatches call to toggle?

cwaring: Also `enabled` in this context means that the experiment is activated not available to be activated, it's persisted in the state

hacdias: @cwaring it seems you read the comment I marked as outdated a while ago. It's not you who needs to update the `enabled` value. IPFS Desktop takes care of that, not Web UI. Please see the latest comment (https://github.com/ipfs-shipyard/ipfs-webui/pull/1048#discussion_r290652415). Perhaps it's a better idea

cwaring: I'm not sure what happened there. I didn't mark it as outdated, possibly because we replied at the same time? Good ideas but I'm still not totally clear, I'll look into this in more detail when I'm back a the desk soon 👌

hacdias: @cwaring cool! I opened a PR: https://github.com/ipfs-shipyard/ipfs-webui/pull/1055. Feel free to merge it here. It is all explained there.

hacdias:
@cwaring made some updates. Now we just use actions. There's no direct contact between `ipfs-desktop` and `experiments` bundle. Also,

- we now block the checkbox while changes are being made so the user doesn't click there multiple times;
- we show an error message when something went wrong.

cwaring:
Great I'll check over it, good move switching the experiments back to an array I was planning to do the same because the conversion overhead was starting to irritate me too 😂

src/bundles/ipfs-desktop.js
```diff
@@ -8,6 +10,10 @@ if (window.ipfsDesktop) {
   bundle = {
     ...bundle,
     reducer: (state = {}, action) => {
+      if (action.type === ACTIONS.EXP_TOGGLE) {
+        window.ipfsDesktop.toggleSetting(`experiments.${action.payload.key}`)
```
cwaring: I would suggest you restrict this toggle action to only execute on known desktop experiments, otherwise if we add webui only ones in the future it will also trigger this method.

src/bundles/notify.js
```diff
@@ -59,6 +59,15 @@ const notify = {
       }
     }
 
+    if (action.type === 'EXPERIMENTS_TOGGLE_FAIL') {
```
cwaring: should we add a success notification too, or do you think that would be too noisy?

src/components/experiments/ExperimentsPanel.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import { connect } from 'redux-bundler-react'
+import Checkbox from '../checkbox/Checkbox'
+import Box from '../box/Box'
+import Title from '../../settings/Title'
+
+const Experiments = ({ doExpToggleAction, experiments, t }) => {
+  // if there are no experiments to show don't render
+  if (experiments && experiments.length > 0) {
+    console.log(experiments)
```
cwaring: rm :)

src/bundles/ipfs-desktop.js
cwaring: we could pass around the `desktop: bool` in the action if that would be the easiest?

src/bundles/notify.js
hacdias: I think it'd be too noisy. The checkbox does the trick already

src/bundles/ipfs-desktop.js
hacdias: It would not be a problem, because there wouldn't be any handler for that event then. But I made a change to fix it: window.ipfsDesktop now has an array with experiments.

src/bundles/experiments.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export default {
+  name: 'experiments',
+  // persist all actions
+  persistActions: Object.values(ACTIONS),
+
+  reducer: (state = {}, action) => {
+    if (action.type === ACTIONS.EXP_TOGGLE) {
+      return block(state, action.payload.key)
```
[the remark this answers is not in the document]
hacdias: Done!

hacdias:
@cwaring just updated!

src/bundles/ipfs-desktop.js
cwaring: In that case perhaps it's better to fail silently rather than having an additional experiments array to update?

src/bundles/notify.js
cwaring: agree, we will likely change these to switches in the future which will improve the UI feedback anyway.

src/bundles/ipfs-desktop.js
hacdias: Yeah, I'll just revert my update then :)

hacdias: Done

cwaring:
thanks @hacdias, one more question: do you think we should unblock after a timeout in case an experiment action didn't trigger for some reason?

hacdias:
@cwaring that's a good question, but what would be the timeout? For example, `ipfs-npm` can take up to 5 minutes to install. It really depends on the machine...

hacdias:
If it was something we knew was immediate, I wouldn't see any problem, but in this case it isn't...

cwaring:
yeah it's a tricky one. I mean if the user reloaded the app or browser window then they could activate again but that may not be immediately obvious. perhaps something to revisit later on.

what might be helpful in determining this is if we wire in analytics to monitor enable/disable and time it took to activate an experiment.

hacdias:
I guess that would be a good idea, to add the analytics. Could you do that, please? If you don't have much time, I can take a look

cwaring:
I'm a bit pushed for time as I need to jump back onto camp but I can take a look early next week if you don't get chance before then :)

hacdias:
I'll do that today then hahaha. It'd be awesome to perhaps merge this today, or this week at most.

hacdias:
@cwaring done on cbbf1c7039b1c54658920fc08ce6ab49e9600bb6. Could you take a look please?

src/bundles/experiments.js
```diff
@@ -0,0 +1,122 @@
...
+      {
+        url: 'https://github.com/ipfs-shipyard/npm-on-ipfs',
+        key: 'readMoreUrl'
+      },
+      {
+        url: 'https://github.com/ipfs-shipyard/npm-on-ipfs/issues',
+        key: 'issueUrl'
+      },
+      {
+        url: 'https://github.com/ipfs-shipyard/npm-on-ipfs',
```
olizilla: We need a plan for how to gather feedback... We could create a "Feedback" issue, and have folks comment on it.  Won't scale if we  get tons of feedback, but I think we can deal with that problem when it happens.

src/bundles/experiments.js
```diff
@@ -0,0 +1,122 @@
...
+    })
+  },
+
+  selectExperimentsState: state => state.experiments,
+
+  selectExperiments: createSelector(
+    'selectIsIpfsDesktop',
+    'selectExperimentsState',
+    (isDesktop, state) => EXPERIMENTS
+      .filter(e => !!e.desktop === isDesktop)
```
olizilla: I like the neat code, but I think we need more flexibility here. My assumption here is that all experiments will be available on desktop and some will be available on desktop and webui. We are deliberately hiding all experiments from webui until we get an api that lets us trigger a restart of the daemon, but once we have that, it's be nice to let webui toggle things like pubsub, mdns etc.

src/bundles/index.js
```diff
@@ -20,13 +20,14 @@ import bundleCache from '../lib/bundle-cache'
...
 import createAnalyticsBundle from './analytics'
+import experimentsBundle from './experiments'
 
 export default composeBundles(
   createCacheBundle(bundleCache.set),
   appIdle({ idleTimeout: 5000 }),
   ipfsBundle({
     tryWindow: false,
-    ipfsConnectionTest: async (ipfs) => {
+    ipfsConnectionTest: async ipfs => {
```
olizilla: We should let prettier do this kind of thing.

src/bundles/experiments.js
hacdias: We're just hiding desktop-specific experiments when we're not inside IPFS Desktop. What's the issue? Didn't understand

src/bundles/index.js
hacdias: Aren't we using [`eslint`](https://github.com/ipfs-shipyard/ipfs-webui/blob/master/package.json#L9)?

src/bundles/experiments.js
hacdias: Should it be there, or on `ipfs-desktop` repository?

src/bundles/experiments.js
cwaring: Yep, this seems to be functioning correctly. It will display everything in desktop and filter out desktop explicit experiments from any other context.

cwaring:
> @cwaring done on [cbbf1c7](https://github.com/ipfs-shipyard/ipfs-webui/commit/cbbf1c7039b1c54658920fc08ce6ab49e9600bb6). Could you take a look please?

does this track which experiment is being enabled/disabled and the time it took to perform the operation? (Sorry on my 📱 ATM so tricky to check)

hacdias:
@cwaring it should only track the time to perform the operation. I'm not sure how to include the experiment here though. I'm not really well aware of how the analytics work

src/bundles/experiments.js
olizilla: You're right. I was thinking about a situtation (but not saying!) where we want to show an experiment on webui but not desktop... and that's probably not going to happen, so this is fine.

Could we change the key on the experiments label from `desktop: true` to `desktopOnly: true`?

src/bundles/index.js
olizilla: My point is, I would like to agree the style guide and automate it rather than see these sorts of changes in PRs

src/bundles/experiments.js
cwaring: Good call, I'll update that 👌

olizilla:
In it's current form, the analytics bundle will just record that an experiment was toggled, but not which one, as it only records the action type. We are deliberately extra cautious in the analytics bundle to be very explicit about what is tracked, so we can give the user a clear answer on what is and isn't tracked.

Either we create an action per experiement, or we add custom logic into the analytics bundle for the experiments action to modify the action name to include the experiment name, here https://github.com/ipfs-shipyard/ipfs-webui/blob/babbb6d2d2e742df50fd169d5c5066cc9e2fc546/src/bundles/analytics.js#L109-L113

cwaring:
> In it's current form, the analytics bundle will just record that an experiment was toggled, but not which one, as it only records the action type. We are deliberately extra cautious in the analytics bundle to be very explicit about what is tracked, so we can give the user a clear answer on what is and isn't tracked.
> 
> Either we create an action per experiement, or we add custom logic into the analytics bundle for the experiments action to modify the action name to include the experiment name, here
> 
> https://github.com/ipfs-shipyard/ipfs-webui/blob/babbb6d2d2e742df50fd169d5c5066cc9e2fc546/src/bundles/analytics.js#L109-L113

@olizilla great, we have a key in the action payload so it should be possible to pick that up here

src/bundles/experiments.js
cwaring: would directing people to a new issue template be a viable option for this?

src/bundles/index.js
cwaring: agree, I have prettier on by default but I don't think it respects the current `eslint` rules without additional config. I'll start a separate PR and we can sling our semicolons around there :)

hacdias:
@cwaring I passed the experiment name on the key. 6b0932e should do the trick!

src/bundles/index.js
hacdias: I'm resolving this conversation right now 😄 I've put the parenthesis back again. Thanks @cwaring! That'd be awesome

src/bundles/experiments.js
olizilla: Neat, what would it say? How would we distinguish it from raise an issue? Are issues a good place for general feedback? They kinda work for it, but nice as it may be, we don't want hundreds of issues that say "i love it" or "I hate it".

Of note, countly has a feedback / ratings plugin https://count.ly/plugins/ratings-and-feedback
though it feels a little gross to popup emojis. Any other options?

daviddias:
The most important question of all (:D), will my little IPFS Cube in my nav bar become the npm on IPFS cube?

image

image

It would feel it gets equipped with a super power!

hacdias:
@daviddias that's definitely an interesting idea. I just wonder how it'd scale - we only have three visible sides of the IPFS cube and we might introduce more experiments or even JS-IPFS support.

I've been wanting to make some changes to IPFS Desktop's icon so perhaps this is a good time to raise an issue there. Done: https://github.com/ipfs-shipyard/ipfs-desktop/issues/956.

daviddias:
> I just wonder how it'd scale - we only have three visible sides of the IPFS cube and we might introduce more experiments or even JS-IPFS support.

I recommend to not overthink it :) You can potentially pick one side to be the "underlying implementation being used" and another for "the last experiment active" and once you have multiple useful experiments, then consider how to represent it.

hacdias:
@olizilla could you take one last look at this please?

PR #1049, opened 2019-05-28
fix: drop multiple files on chrome

### LINKED ISSUE #1045: Dropping multiple files only adds one on Chrome
**Describe the bug**
Using go-ipfs 0.4.19 & Chrome stable 74 drop multiple individual files... only 1 file is added.

**To Reproduce**
1. Go to files
2. Drop two or more files on the page
3. Only one file appears added

Firefox is not affected. The files are all added as expected.

 - OS: macOS
 - Browser: Chrome (and electron)
 - Version: 74

hacdias:
Fixes #1045 by using my [fork](https://github.com/hacdias/datatransfer-files-promise) of `datatransfer-files-promise`. I didn't publish it because I was hoping the original creator would accept my PR, although @grabantot doesn't seem very active on GitHub.

License: MIT

### DIFF
--- a/src/lib/dnd-backend.test.js
+++ b/src/lib/dnd-backend.test.js
@@ -17,7 +17,7 @@ function makeEntry (item) {
       readEntries: function (cb) {
         if (!this.called) {
           this.called = true
-          cb(item.childs.map(makeEntry))
+          cb(item.children.map(makeEntry))
         } else {
           // eslint-disable-next-line
           cb([])
@@ -112,7 +112,7 @@ it('one directory with a file', async () => {
     fullPath: '/dir',
     isDirectory: true,
     isFile: false,
-    childs: [{
+    children: [{
       fullPath: '/dir/test.txt',
       isFile: true,
       isDirectory: false,
@@ -125,10 +125,10 @@ it('one directory with a file', async () => {
   const res = await backend.currentNativeSource.item.filesPromise
 
   expect(res.length).toBe(1)
-  expect(res[0].fullPath).toBe(files[0].childs[0].fullPath)
-  expect(res[0].isDirectory).toBe(files[0].childs[0].isDirectory)
-  expect(res[0].isFile).toBe(files[0].childs[0].isFile)
-  expect(res[0].name).toBe(files[0].childs[0].name)
+  expect(res[0].fullPath).toBe(files[0].children[0].fullPath)
+  expect(res[0].isDirectory).toBe(files[0].children[0].isDirectory)
+  expect(res[0].isFile).toBe(files[0].children[0].isFile)
+  expect(res[0].name).toBe(files[0].children[0].name)
 })
 
 it('multiple directories with some files', async () => {
@@ -136,7 +136,7 @@ it('multiple directories with some files', async () => {
     fullPath: '/dir',
     isDirectory: true,
     isFile: false,
-    childs: [{
+    children: [{
       fullPath: '/dir/test.txt',
       isFile: true,
       isDirectory: false,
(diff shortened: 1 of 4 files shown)


hacdias:
@olizilla why I haven't done the tests for this: https://github.com/ipfs-shipyard/ipfs-webui/pull/1047#issuecomment-496425242

olizilla:
@hacdias 👍 you could publish it under your npm namespace like `@hacdias/datatransfer-files-promise` in the meantime.

olizilla:
So it'd be rad to add tests for the `datatransfer-files-promise`...it's a function that takes an array-ish `DataTransferItemList`, where each entry may be a single file or a tree of files, and returns a flat array of objects with a `filePath` property, that encodes the path in the tree where it came from.

We'd need to do some mocking... each item in the input array needs a `.webkit​GetAsEntry()` function that returns an entry with either `.isFile` or `.isDirectory` set to true and and `entry.name`. Where we set `.isFile` to true, we expect to just see filePath in the output object set to `entry.name`. Where isDirectroy is true, we need to provide an `entry.createReader` function that returns an object with a `.getEntries` function that returns an array of entries... from which we can mock out a tree structure of entries.

It's a bit of work, but it'd be really nice.

olizilla:
For example we should test what happens when a brower use `getAsEntry` instead of `webkitGetAsEntry` as per

> This function is implemented as webkitGetAsEntry() in non-WebKit browsers including Firefox at this time; it may be renamed to simply getAsEntry() in the future, so you should code defensively, looking for both.

https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry

hacdias:
@olizilla yeah, thanks. I'm aware of the structure, I've already started them 😃

hacdias:
@olizilla just released it as a package. The tests will go on #1047

hacdias:
@olizilla I'm hitting an infinite loop on the directory tests and I don't know why. I know the loop is happening on the module (`datatransfer-files-promise`, both with mine and the original version) but it must be something in the structure I'm creating that's causing it.

hacdias:
@olizilla found the issue. Ready for review 😃

src/lib/dnd-backend.test.js
```diff
@@ -0,0 +1,214 @@
...
+  }]
+
+  const backend = getBackend(files)
+  const res = await backend.currentNativeSource.item.filesPromise
+
+  expect(res.length).toBe(files.length)
+  expect(res[0].fullPath).toBe(files[0].fullPath)
+  expect(res[0].isDirectory).toBe(files[0].isDirectory)
+  expect(res[0].isFile).toBe(files[0].isFile)
+  expect(res[0].name).toBe(files[0].name)
```
olizilla: we want to ensure the second file is as expected too.

src/lib/dnd-backend.test.js
```diff
@@ -0,0 +1,214 @@
...
+      isFile: true
+    })
+  }
+
+  if (item.isDirectory) {
+    item.createReader = () => ({
+      readEntries: function (cb) {
+        if (!this.called) {
+          this.called = true
+          cb(item.childs.map(makeEntry))
```
olizilla: nitpick: use `children`, it's the (special case) pluralisation, and is more familiar as it's often used in react.
```suggestion
          cb(item.children.map(makeEntry))
```

hacdias:
@olizilla just updated the tests. One question about https://github.com/ipfs-shipyard/ipfs-webui/blob/master/src/lib/count-dirs.test.js#L32-L42. Why is it testing against 8? Is that the wrong part?

olizilla:
@hacdias it assumes ipfs will make a dir for the root `/`, perhaps that assumption should change.

olizilla:
see: https://github.com/ipfs-shipyard/ipfs-webui/blob/83fbf45156f78c5fd90bb502e600d7f127c32749/src/lib/count-dirs.test.js#L4-L6

olizilla:
@hacdias does that make sense?

hacdias:
@olizilla yes, it does. Could we merge this PR and I'll take a look at count-dirs on another one?

PR #1050, opened 2019-05-31
feat: restart ipfs daemon after config change

hacdias:
See https://github.com/ipfs-shipyard/ipfs-desktop/pull/943.

License: MIT


hacdias:
I've made the required changes and now we just tell Desktop the configuration was changed. Then, it handles the rest.

PR #1055, opened 2019-06-05
feat: experiments API for IPFS Desktop

hacdias:
@cwaring this adds an experiments API for IPFS Desktop redux bundle so we can get the current enabled state, know if an experiment is available and toggle it. The API is:

**To know if `expId` is available:**

```js
store.selectDesktopAvailableExperiments().includes('expId')
```

**To know if `expId` is enabled:**

```js
store.selectDesktopExperiments().expId
```

**To toggle `expId`:**

```js
store. doDesktopToggleExperiment('expId')
```

This should only be used for IPFS Desktop-related experiments such as NPM on IPFS.

### DIFF
--- a/src/bundles/ipfs-desktop.js
+++ b/src/bundles/ipfs-desktop.js
@@ -21,6 +21,8 @@ if (window.ipfsDesktop) {
 
     selectDesktopSettings: state => state.ipfsDesktop,
 
+    selectDesktopExperiments: () => window.ipfsDesktop.experiments,
+
     doDesktopStartListening: () => async ({ dispatch }) => {
       window.ipfsDesktop.onConfigChanged(config => {
         dispatch({
--- a/src/bundles/ipfs-desktop.js
+++ b/src/bundles/ipfs-desktop.js
@@ -7,6 +7,7 @@ let bundle = {
 if (window.ipfsDesktop) {
   bundle = {
     ...bundle,
+
     reducer: (state = {}, action) => {
       if (!action.type.startsWith('DESKTOP_')) {
         return state
@@ -21,7 +22,9 @@ if (window.ipfsDesktop) {
 
     selectDesktopSettings: state => state.ipfsDesktop,
 
-    selectDesktopExperiments: () => window.ipfsDesktop.experiments,
+    selectDesktopAvailableExperiments: () => window.ipfsDesktop.experiments,
+
+    selectDesktopExperiments: state => state.ipfsDesktop.experiments,
 
     doDesktopStartListening: () => async ({ dispatch }) => {
       window.ipfsDesktop.onConfigChanged(config => {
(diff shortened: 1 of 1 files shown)


hacdias:
@cwaring another option would be to mainly dispatch actions. For example, to toggle something, you'd dispatch an event called `TOGGLE_NPM_ON_IPFS`. Then, it is enabled by `ipfs-desktop` bundle.

If it fails, we could dispatch an event `TOGGLE_NPM_ON_IPFS_FAILED` with the error, otherwise we'd change the value of the option.

The main question, IMO, is to decide how to know wether or not the option is enabled:

1. One option could be what I said in the main post above, and you just `selectDesktopExperiments().expId` and it is always up to date.
2. Another option could be to dispatch an event and the experiments bundle would always contain the state itself. Problem with this? It would be harder to get the initial state in a first instance.

cwaring:
Yes, I believe actions (and effects) are the correct approach to manage this communication which would require the desktop bundle to dispatch an appropriate success or failure action on toggle. I would prefer these to be generic actions with a variable payload so that we don't need to add custom hooks for every new experiment.

Currently the webui experiments bundle persists the state of the feature, which also feels like the correct place for this data.

So my suggestion is:
1) I amend the webui to dispatch a generic toggle action that includes the key of the experiment the user is attempting to enable/disable
2) you respond to that action and execute the command to enable the functionality
3) depending on the result of # 2 you dispatch a success or failure action that webui monitors in order to update the enabled state

How does that sound?

hacdias:
@cwaring that sounds great!

But there's still a point missing: initially, how do you know if something is activated or not? Like, for example, I enable npm on ipfs, then refresh the page. How does Web UI know that it is activated? I know the state is persisted to the local storage, although IPFS Desktop settings might be changed manually, so their state would change.

cwaring:
Agree. That's the last missing step but I could use your suggested API here to get the initial values for any desktop experiments on boot.

hacdias:
@cwaring so to simplify, you can use this to get the values:

```js
store.selectDesktopSettings().experiments.expId
```

I won't need to add more stuff.

Once you know the action names for success, toggling and failure, please tell me so I can update this,.

cwaring:
I've refactored and wired up the `doDesktopToggleExperiment` toggle method over in #1048 so now I just need to test in context of the desktop app to see if it all works, also you shouldn't need any additional actions or api surface area 🤞

Let me know your thoughts when you get change.

hacdias:
@cwaring you're assuming `doDesktopToggleExperiment` fails if there's an error. It doesn't! It won't work like that.

I thought we had agreed on dispatching an event saying "Option X needs to be toggled" and then some bundle would toggle it and dispatch a success or fail event.

If you don't mind, I can make some final updates on #1048 to just use events.

cwaring:
Yes you're right, I considered both while making the updates and this seemed like.a simpler option to start with.

The base success/fail actions are in place so if you'd prefer to switch this to a dispatch now please feel free to make the updates. It wasn't 100% clear what pattern I should use to queue up additional actions from a reducer event so I'm keen to have your input here.

PR #1063, opened 2019-07-08
chore: add sorting to peers table

JustMaier:
As discussed in #1062, sorting has been added to the peers table with latency ASC as the default.


olizilla:
@JustMaier the PR for updating the peers tab got changed some, but it's merged now. Are you up for updating your PR so we can get it merged?

src/peers/PeersTable/PeersTable.js
```diff
@@ -2,16 +2,33 @@ import React from 'react'
...
 import { translate, Trans } from 'react-i18next'
-import { Table, Column, AutoSizer } from 'react-virtualized'
+import { Table, Column, AutoSizer, SortDirection } from 'react-virtualized'
 import CountryFlag from 'react-country-flag'
 import Cid from '../../components/cid/Cid'
 
+function sortByProperty (arr, property, direction) {
+  if (!arr) return arr
+  const dirModifier = direction === SortDirection.ASC ? 1 : -1
+  return arr.sort(({ [property]: a }, { [property]: b }) => (a == null) - (b == null) || dirModifier * +(a > b) || dirModifier * -(a < b))
```
olizilla: so fancy. We should probably extract this to a seperate file we can use else where and test.

JustMaier: Happy to make that happen. I'm thinking `/src/lib/sort.js` seems like a good home.

JustMaier:
@olizilla This should be ready to merge now :)

olizilla:
Sorry @JustMaier, your original PR was branched from @fsdiogo's branch, which is now merged into master. This PR will merge cleanly into @fsdiogo's branch, but we want to move the base to be master, so we can get it in there. I could port the changes over myself, but I don't want to lose the commit trail, as this is your work and you should get the git credit! Do you have time to rebase this work from the latest master branch so we can merge it in there? I flipped the base branch in github, but it suggested there were conflicts, so i flipped it back. I think it just needs a rebase from your local branch.

JustMaier:
Going to close and re-open then :)

PR #1068, opened 2019-07-17
chore: add sorting to peers table

JustMaier:
Rebased for master.

Sorting in peer table with default sorting based on latency column


JustMaier:
@fsdiogo oops. Thanks for catching that. Fixed and pushed.

olizilla:
Thank you @JustMaier!

JustMaier:
Yay! Happy to have been able to contribute a little something to the cause :)

PR #1071, opened 2019-07-23
cache geoip lookups

jbenet:
geoip lookups for the same ip4 addr can be cached. this uses a similar
cache to the peerid location cache (arguably peerid loc shouldnt be
cached-- ip addr only?).

---

Requesting a pull to ipfs-shipyard:master from ipfs-shipyard:cache-geoip-lookups

Changes:

a1290cc (jbenet, 20 seconds ago)
   bumped up concurrency for geoip lookups

28af4b6 (jbenet, 6 minutes ago)
   cleaned up geoip cached resolver code

   cleaned up the geoip cached resolver code some. put things into a class.

4f33563 (jbenet, 32 minutes ago)
   cache geoip lookups

   geoip lookups for the same ip4 addr can be cached. this uses a similar
   cache to the peerid location cache
   (arguably peerid loc shouldnt be cached-- ip addr only?) the code is a bit
   messy, and probably should be refactored

   This reduced my load time significantly.

   lots of nodes have the same ip4 addr:
   - nodes under same physical machine
   - nodes under same VPN
   - nodes under same carrier grade nats
   - nodes behind city-wide firewalls

   eg lots of nodes in china show the same ip addr.

e9ece8e (jbenet, 2 hours ago)
   class should be className

   Seems like this is a bug -- caught by the dev server compiler/linter


jbenet:
Note, i couldn't get the map to render properly. it gives this error in the console, and stops rendering dots. Looks like a d3 error w/ the paths:
- https://stackoverflow.com/questions/43811434/error-path-attribute-d-expected-number-mnan-nanlnan-nan
- https://stackoverflow.com/questions/37252425/d3-js-d3-min-js1-error-path-attribute-d-expected-number-mnan-nanlnan-na/37535202

I dont think this is due to this changeset -- the same error exists for me in master. 

Note: looks like d3 currently renders one dot per peer. it should probably render one dot per city (massively deduped).

image

jbenet:
I also get these logs in the console -- ~6msgs every second

image

jbenet:
BTW, this is critical when peer count is large :) -- though it starts helping at ~100 peers

image

olizilla:
Wowsez! I need to create some tests with large numbers of peers. I've never seen a 26k before. Thanks for the heads up, will dig into it.

olizilla:
Here is a first draft of all things we need to fix in the peers page: https://github.com/ipfs-shipyard/ipfs-webui/issues/1072

olizilla:
Am digging in to why the e2e test fails on CI. @jbenet I'm gonna adopt this PR. Let me know if you are still working on it.

jbenet:
thanks for adopting! go for it! 👍

olizilla:
@hacdias if you're happy with this it can be merged as is and we can refine it as discussed in #1072

hacdias:
Merged with master, looks good and seems to work well.

PR #1079, opened 2019-08-01
feat: update translations

hacdias:
`tx pull -a`


hacdias:
Some sentences seem to have regressed for some reason. Any ideas @lidel?

lidel:
@hacdias it is due to changes in English source: https://github.com/ipfs-shipyard/ipfs-webui/pull/1039

It is ok to merge this, and repeat `tx pull -a` again before release, as translations for fixes strings may land later.

But don't stress too much about it, it should not block a release.
Some regressions will happen, and it is okay (it will get translated in the future).

PR #1114, opened 2019-08-16
feat: add inspect near browse

### LINKED ISSUE #1109: Can't ipld-explore anymore
We replaced the IPLD Explore bar by the Files Explore bar. Now we can't ipld-explore anymore.

Suggestion:

1. Change the bar on Explore page to the IPLD Explore; **or**
2. Add the bar somewhere in the start page of the explore page.

/cc @olizilla

hacdias:
Close #1109. Closes #1112. #1027 introduced a regression: we replaced the IPLD explore bar by the files Explore bar. Although, we didn't put the IPLD explore bar anywhere else.

This PR makes the top bar work for both Files and IPLD sections. How? Like this:

Untitled

License: MIT

### DIFF
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -101,7 +101,7 @@
     }
   },
   "exploreForm": {
-    "explore": "Explore",
+    "browse": "Browse",
     "inspect": "Inspect"
   },
   "hashUnavailable": "hash unavailable",
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -101,7 +101,8 @@
     }
   },
   "exploreForm": {
-    "explore": "Explore"
+    "explore": "Explore",
+    "inspect": "Inspect"
   },
   "hashUnavailable": "hash unavailable",
   "pins": "Pins",
--- a/public/locales/en/status.json
+++ b/public/locales/en/status.json
@@ -65,9 +65,9 @@
       "paragraph1": "This chart shows the network bandwith usage over time.",
       "paragraph2": "The speedometers show the current outgoing and incoming traffic."
     },
-    "step5": {
+    "stepExplore": {
       "title": "Explore",
-      "paragraph1": "Paste a CID and explore how that data is structured and linked across protocols."
+      "paragraph": "Paste a CID and explore the files contained behind the CID or see how that data is structured and linked across protocols."
     }
   },
   "bandwidthStats": "Bandwidth stats",
--- a/src/App.js
+++ b/src/App.js
@@ -58,7 +58,7 @@ export class App extends Component {
   }
 
   render () {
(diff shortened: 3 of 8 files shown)


olizilla:
That is snazzy, but perhaps it could be much simpler.

- if the CID is point to a node that is not a dag-pb unixfs thing, the explore button should open it in the IPLD explore page, otherwise open it in the files page.
- There should be a clear option in the files page to view the current connent in the ipld explorer. I think we already have that in the context menu and the action bar, so perhaps that is already covered.

olizilla:
if you are on the ipld explore page, that bar should open what you search for in the ipld page.

hacdias:
@olizilla 

> if the CID is point to a node that is not a dag-pb unixfs thing, the explore button should open it in the IPLD explore page, otherwise open it in the files page.

I thought about that first, although that would require to resolve possible `/ipns` domains and resolve whole paths such as `/HashToCbor/but/this/is/unixfs`.

> There should be a clear option in the files page to view the current connent in the ipld explorer. I think we already have that in the context menu and the action bar, so perhaps that is already covered.

Yes, already possible.

@lidel yes, using 'Browse'/'Explore' is the best IMO. Will change.

hacdias:
Ping @olizilla @lidel

lidel:
## Handle "Browse" and non-unixfs

Clicking on "Browse" for non-unixfs node (eg. `bafyreiddymapg5zcpma3iu4wingqvois6jirucn5776wdsyg5f3f65v75a`) produces error.

@hacdias  Are you able to rebase this PR on top of master (to include fixes from https://github.com/ipfs-shipyard/ipfs-webui/pull/1115)?

I want to confirm user will see below message instead of error :)

> 

## Proof-read English text

SGTM, but let's get :+1:  from a native speaker (@olizilla / @autonome)

hacdias:
@lidel just merged with master! Remember that now it will only show up if you go to the specific link! If you're just navigating, you'll be automatically taken to the Inspector.

ericronne:
Yep. Actually any interactive element should look clickable by default, otherwise users may not ever know to hover over it. Here are two ways we could telegraph clickability in regards to this feature set …

image

ps wearing my "user hat," i'm still not sure what to make of the house icon. My brain expects that to be a number, in keeping with the pattern set by the blocks and pins counts. The word "files" may be part of the problem, as it suggests "X files" 👽 (number of files) to me.

Would changing the label to "home" make sense?

lidel:
No, "home" feels odd. Actually, do we need the "files" as a navigation item in the section on the right at all?

Clickable "files" is already present as the root of breadcrumbs on the left:


My intuition is to remove it or replace clickable icon with non-clickable count. 
It could be total size of files in MFS (`CumulativeSize` from ` ipfs files stat /`

Thoughts?

hacdias:
@ericronne @lidel moving that conversation to https://github.com/ipfs-shipyard/ipfs-webui/issues/1122.

PR #1115, opened 2019-08-16
fix: non unixfs on files

### LINKED ISSUE #1095: Files/Explore: unrecognized node type when DAG is not unixfs
##  Problem

https://github.com/ipfs-shipyard/ipfs-webui/pull/1027 introduced a clickable "Pins" counter on the "Files" screen. 

It works fine with empty/new node, but I found a bug that manifests against my pretty old repo which has over 800 pins.

@hacdias let me know if the error is not something  obvious, I can look at this closer and provide  more info if needed.

## Screenshots

> 

>  

> ss-2019-08-14--13-04-28

**Additional context**

go-ipfs v0.4.22, ipfs-webui 2.5.0

hacdias:
Closes #1095.

Clicking on the item goes to to the explore section:

image

It has less ctx menu opts:

image

If the user tries to navigate to it anyway (though the URL or sth):

image

### DIFF
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -108,5 +108,6 @@
   "pinned": "Pinned",
   "blocks": "Blocks",
   "more": "More",
-  "files": "Files"
+  "files": "Files",
+  "cidNotFileNorDir": "The current link isn't a file, nor a directory. Try to <0>fuck</0> it instead."
 }
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -109,5 +109,5 @@
   "blocks": "Blocks",
   "more": "More",
   "files": "Files",
-  "cidNotFileNorDir": "The current link isn't a file, nor a directory. Try to <0>fuck</0> it instead."
+  "cidNotFileNorDir": "The current link isn't a file, nor a directory. Try to <0>inspect</0> it instead."
 }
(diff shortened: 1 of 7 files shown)


public/locales/en/files.json
```diff
@@ -108,5 +108,6 @@
   "pinned": "Pinned",
   "blocks": "Blocks",
   "more": "More",
-  "files": "Files"
+  "files": "Files",
+  "cidNotFileNorDir": "The current link isn't a file, nor a directory. Try to <0>fuck</0> it instead."
```
fsdiogo:
```suggestion
  "cidNotFileNorDir": "The current link isn't a file, nor a directory. Try to <0>explore</0> it instead."
```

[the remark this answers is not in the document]
hacdias: Who knows?

src/bundles/files/actions.js
```diff
@@ -22,6 +22,27 @@ const realMfsPath = (path) => {
...
+const stat = async (ipfs, hashOrPath) => {
+  const path = hashOrPath.startsWith('/')
+    ? hashOrPath
+    : `/ipfs/${hashOrPath}`
+
+  try {
+    const stats = await ipfs.files.stat(path)
+    return stats
+  } catch (e) {
+    if (e.toString().toLowerCase().includes('unixfs')) {
```
lidel: This turned out to not be robust enough. 

I have old repo with various weird stuff in it and it breaks the pin screen due to error messages without `unixfs`:

unhandled-errors-2019-08-20--16-32-16

I proposed a robust fix in https://github.com/ipfs-shipyard/ipfs-webui/pull/1121

Issue #1072, opened 2019-07-23
Issues with the peers page


olizilla:
### Sorting 

- #909 introduced lexographic sorting of the peers by multiaddr
- #1068 allows the user to sort per column

We don't need both, or at least, we don't need to always sort by multiaddr in the `selectPeers` function, if we're going to sort by something else later on.

- [x] Remove sorting by multiaddr, it is replaced by sorting by connection type. 
- [x] Fix location sorting on peers page.
- [ ] Check the performance of sorting with large lists of peers.
	- known issues with >4000 peers (some notes in https://github.com/ipfs-shipyard/ipfs-webui/pull/1262#pullrequestreview-306242986)
- [x] Update geoip dataset (#1249)

### Caching

- [x] #1071 introduces caching the location for an IP. The geoip look up is where most of the work happens. We should remove caching by PeerID, as it's totally reasonable for PeerIDs to move IP addresses.

### Rendering

- [x] As pointed out in #1071 we get d3 to render a dot per peer. Many peers get geolocated to the same city, and the map is at global scale, so we can optimise this to only render a dot per city / unique location.
- [ ] The peers table should fill the available vertical space on screens with larger vertical height.

### Complexity

- [ ] We do a lot of work in peer-locations.js to expose the internal logic of the queuing of geoip look-ups as state in redux. We are not visualising the queue, nor do we plan to, so i suspect the code could be written more simply. We could extend the work in #1071 to encapsulate the geoip look up queue and caching in a class, and simply have it periodically update the redux state with batches of locations as they are found, rather than have all the queuing logic managed in redux.

jbenet:
This issue is great to have! 👏 

Some more thoughts, take them or leave them.
- The structure of the cache & loading for geoip locations seems off to me -- cache lookups are fast, and right now they're gating on the concurrency queue. We could switch things up so that we can check the cache first for all items, and only if not in the cache do we add them to the (bounded) geoip lookup queue. (this should have the effect of loading all the peers & ip addresses we've looked up before first, and then 
- i think geoip db doesnt change very often (years?) the geoip cache could stay for longer. (may be better to bound it by size too, so it doesnt get too big? not sure)
- Rendering seems really expensive at the moment. i think there's probably inefficiency because the actual work (rendering a list with <100K items, rendering a map w/ <100K dots) doesn't seem that intensive.
[...]

PR #1116, opened 2019-08-16
chore: Peer page sorting adjustments

JustMaier:
In response to #1072 I've done the following:
- [x] Removed sorting by multiaddr
- [x] Fixed location sorting

I also made the following minor changes:
- Remove sorting notes column
- Simplified latency sorting (removed `rawLatency`)

---

The last item mentioned in #1072 is checking performance with a large number of peers. Any suggestions about the best way to test that?

### DIFF
--- a/src/bundles/peer-locations.js
+++ b/src/bundles/peer-locations.js
@@ -152,8 +152,7 @@ export default function (opts) {
           locationObj.latitude
         ]
         const connection = parseConnection(peer.addr)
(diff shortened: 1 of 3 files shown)


hacdias:
> The last item mentioned in #1072 is checking performance with a large number of peers. Any suggestions about the best way to test that?

That's a good question. Perhaps through testing or even mocking the peers bundle and just inject thousands of peers. I wonder if that'd work! Or even using the Storybook with tons of peers

Issue #1122, opened 2019-08-20
Files page: buttons and inconsistencies


hacdias:
From https://github.com/ipfs-shipyard/ipfs-webui/pull/1114

> @ericronne 
> Yep. Actually any interactive element should look clickable by default, otherwise users may not ever know to hover over it. Here are two ways we could telegraph clickability in regards to this feature set …
>
> ![image](https://user-images.githubusercontent.com/9259185/63365611-96af4100-c33d-11e9-8a9d-718178deb1a1.png)
>
> ps wearing my "user hat," i'm still not sure what to make of the house icon. My brain expects that to be a number, in keeping with the pattern set by the blocks and pins counts. The word "files" may be part of the problem, as it suggests "X files" 👽 (number of files) to me.
>
> Would changing the label to "home" make sense?

> @lidel 
> No, "home" feels out of place. Now that I think about it, do we need the "files" as a navigation item in the section on the right at all?
>
> Clickable "files" is already present as the root of breadcrumbs on the left:
![2019-08-20--20-00-41](https://user-images.githubusercontent.com/157609/63371951-5e335580-c385-11e9-8985-ba28bcf8b5b6.png)
>
> My intuition is to remove it or replace clickable icon with non-clickable count. 
> It could be total size of files in MFS (`CumulativeSize` from ` ipfs files stat /`
> 
> Thoughts?

hacdias:
@lidel then, if you went to pins section, how'd you reach 'Files'? You could certainly click on the navigation bar on the left, but...

ericronne:
Ah, yes, i thought i was forgetting some [conversations](https://github.com/ipfs-shipyard/ipfs-webui/pull/1027). 

I think @lidel's intuition sounds good: perhaps we make the `blocks pins files` box merely a non-clickable counter. To @fsdiogo's point from a while back, it's an odd mix of clickable / non-clickable items.

_Ignore the following if it represents an utter lack of proper grokking of things! …_
Might a pull-down selector it be a good solution for the breadcrumb region of the interface? Maybe we even include a `blocks` option summarizing anything we know of their blocks (if anything beyond the count), for completeness. (And so as to avoid a pull-down with only two options.)

![Artboard](https://user-images.githubusercontent.com/9259185/63387345-ca549000-c36a-11e9-867b-522348c74f37.png)

hacdias:
[...]

PR #1128, opened 2019-08-25
feat: improve nav bar

hacdias:
This applies #1122!

image

Is it better with one or zero decimals?

image

I believe that using 0 decimals can hide a bit the difference between repo and the files sizing... 

License: MIT

### DIFF
--- a/src/files/header/Header.js
+++ b/src/files/header/Header.js
@@ -24,7 +24,11 @@ function BarOption ({ children, title, isLink = false, className = '', ...etc })
 }
 
 function humanSize (size) {
-  return filesize(size || 0, { round: 1, spacer: '' })
+  if (!size) return 'N/A'
+
+  return filesize(size || 0, {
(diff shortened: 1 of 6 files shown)


src/files/header/Header.js
```diff
@@ -48,16 +54,20 @@ class Header extends React.Component {
         </div>
 
         <div className='mb3 flex justify-between items-center bg-snow-muted joyride-files-add'>
-          <BarOption title={t('blocks')}>
-            { repoNumObjects ? SimplifyNumber(repoNumObjects, { decimal: 0 }) : 'N/A' }
+          <BarOption title={t('files')} isLink onClick={() => { onNavigate('/files') }}>
+            { filesSize ? humanSize(filesSize || 0) : 'N/A' }
```
lidel: Let's move checks responsible for `'N/A'` and `|| 0`  to `humanSize`, so its not repeated  here and in line 70

hacdias:
@lidel done:

image

PR #1133, opened 2019-08-28
chore: add dev.webui.ipfs.io

hacdias:
@olizilla as far as I understood, all subdomains from `ipfs.io` are pinned to the gateway. This adds a `dev.webui.ipfs.io` subdomain for the latest commit and changes the current `webui.ipfs.io` to the latest **released version** and not commit.

On tags, we update **both** domains DNSLinks, otherwise we just update the development website.

Please let me know if there is anything else that should be done to enable the domain `dev.webui.ipfs.io`. I believe someone should set some CNAME or something like that.

Ref.: This blocks further progress on fixing IPFS Desktop CI issues: https://github.com/ipfs-shipyard/ipfs-desktop/issues/1055.

License: MIT

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,6 @@ jobs:
       - image: olizilla/ipfs-dns-deploy:latest
         environment:
           DOMAIN: webui.ipfs.io
-          DEV_DOMAIN: dev.webui.ipfs.io
           BUILD_DIR: build
     steps:
       - attach_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
             echo "Website added to IPFS: https://ipfs.io/ipfs/$hash"
 
             if [ "$CIRCLE_BRANCH" == "master" ] ; then
-              dnslink-dnsimple -d $DEV_DOMAIN -r _dnslink -l /ipfs/$hash
+              dnslink-dnsimple -d $DOMAIN -r _dnslink.dev -l /ipfs/$hash
             fi
 
             if [ ! -z "$CIRCLE_TAG" ] ; then
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
               dnslink-dnsimple -d $DOMAIN -r _dnslink.dev -l /ipfs/$hash
             fi
 
-            if [ ! -z "$CIRCLE_TAG" ] ; then
+            if [ npx semver "$CIRCLE_TAG" ] ; then
               dnslink-dnsimple -d $DOMAIN -r _dnslink -l /ipfs/$hash
             fi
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -29,6 +29,7 @@ jobs:
       - image: olizilla/ipfs-dns-deploy:latest
         environment:
           DOMAIN: webui.ipfs.io
+          DEV_DOMAIN: dev.webui.ipfs.io
```
olizilla:
```suggestion
```

Not needed... The domain here is actually the "zone". The zone is a logical grouping of dns records that can include subdomains of the root domain. The zone here is `webui.ipfs.io`, and we'll use that as the root for configuring `_dnslink.webui.ipfs.io` and `_dnslink.dev.webui.ipfs.io`

.circleci/config.yml
```diff
@@ -43,6 +44,10 @@ jobs:
             echo "Website added to IPFS: https://ipfs.io/ipfs/$hash"
 
             if [ "$CIRCLE_BRANCH" == "master" ] ; then
+              dnslink-dnsimple -d $DEV_DOMAIN -r _dnslink -l /ipfs/$hash
```
olizilla:
```suggestion
              dnslink-dnsimple -d $DOMAIN -r _dnslink.dev -l /ipfs/$hash
```
The record name includes the .dev rather than the zone.

.circleci/config.yml
```diff
@@ -43,6 +44,10 @@ jobs:
...
+            fi
+
+            if [ ! -z "$CIRCLE_TAG" ] ; then
```
olizilla: Nice! I prefer this to using a deploy branch, as is done in peer pad. Would you be up for updating the example in the ipfs-dns-deploy readme with this?

hacdias: Done! https://github.com/ipfs-shipyard/ipfs-dns-deploy/pull/4/files

olizilla: hmm we dont want any old tag to be interpreted as a release... the solution here uses a regex filter to ensure that only tags that look like semver trigger builds... https://discuss.circleci.com/t/migrating-tag-based-deployment-to-2-0/23967/3

What do you think.

hacdias: Updated to use [`semver`](https://www.npmjs.com/package/semver), which is a small package that checks if the input is a sem version! It's easier and more readable than regexing in the shell

hacdias: Perhaps it could be installed globally on the Docker image, just like `dnslink-cloudflare`.

hacdias: Ping @olizilla.

olizilla: Very cool! Let's use npx for now.

olizilla:
Setting up the nginx config for the domain is blocked on porting our deploy process to ansible. Should land this week.

PR #1136, opened 2019-08-30
feat: share countly data with ipfs desktop

hacdias:
This allows Web UI to receive the Countly Key and Device ID from IPFS Desktop itself. With that, the data can be better analysed. Also, this shares the consents we make on Web UI so Desktop knows what to send and what not to.

Please see https://github.com/ipfs-shipyard/ipfs-desktop/issues/1063.

License: MIT


hacdias:
Made some changes to track IPFS Desktop Settings toggle through Web UI instead so we could even simplify the code on IPFS Desktop side. We are tracking if they are enabled or disabled or if toggling failed.

Also, we now show to the users the tracked actions by IPFS Desktop through a custom selector in IPFS Desktop bundle.

PR #1138, opened 2019-09-02
fix: draggable area on desktop

hacdias:
Increases the draggable area of IPFS Desktop. The new draggable area is (anywhere in the light blue area can be clicked to drag and move the window around):

image

- Please see https://github.com/ipfs-shipyard/ipfs-desktop/pull/1071


hacdias:
@olizilla I decided to merge this since we already talked about it on a call and it's quite trivial. Anything you think that might be worth changing, let me know.

PR #1167, opened 2019-09-15
feat: show all protocols in connectivity

### LINKED ISSUE #1141: Peers: WebSockets transport displayed as TCP
**Describe the bug**

WebSockets peers are indistinguishable from TCP ones 

**To Reproduce**
1. Disable TCP transport (to ensure its not used if peer listens on multiple transports)
2. Connect only to WebSocket peer (eg. `/dns4/node1.preload.ipfs.io/tcp/443/wss/ipfs/Qmbut9Ywz9YEDrz8ySBSgWyJk41Uvm2QJPhwDJzJyGFsD6`)

**Current behavior**

`/tcp/443/wss/` is listed as regular  `tcp`:

wss-as-tcp-2019-09-03--19-18-36

**Expected behavior**

WebSocket connections should be shown as `ipv6・tcp・wss` and  `ipv6・tcp・ws`

hacdias:
Fixes #1141. Shows all protocols in the connectivity column. **This is an alternative for #1166. Only one must be merged.**

image

License: MIT

### DIFF
--- a/src/bundles/peer-locations.js
+++ b/src/bundles/peer-locations.js
@@ -253,9 +253,7 @@ const toLocationString = loc => {
 }
 
 const parseConnection = (multiaddr) => {
-  const opts = multiaddr.toOptions()
-
-  return `${opts.family}・${opts.transport}`
+  return multiaddr.protoNames().join('・')
 }
 
 const parseLatency = (latency) => {


lidel:
FYI noticed interesting edge case with js-ipfs in Brave:
when ws-star multiaddrs are present, they do not fit and get trimmed:

> 

Personally, I think it is actually a "good enough" way to display this type of edge case. Ellipsis prompts uses to hover and when https://github.com/ipfs-shipyard/ipfs-webui/pull/1169 lands the full multiaddr is displayed, which is what I would expect from this type of UI.

hacdias:
Thanks for the review @lidel. I also prefer this one. Hence, I am merging it!

PR #1168, opened 2019-09-15
fix: correctly highlight p2p-circuit addresses

### LINKED ISSUE #1160: /p2p-circuit/ multiaddrs highlight invalid segment
Noticed inconsistency in how multiaddr segments are highlighted.

Example from Status page:

I believe IP and PORT should be highlighted in both cases. 
Quick fix would be to detect `/p2p-circuit/ ` and shift hihilght by one :)

hacdias:
Fixes #1160. Shifts the highlight of `/p2p-circuit` addresses to the right, highlighting the IP address and the port.

I noticed there are some addresses that don't contain IP address, nor port. Should we perhaps disable the highlight on those? If so, that should be a different PR though.

image

License: MIT

### DIFF
--- a/src/components/address/Address.js
+++ b/src/components/address/Address.js
@@ -3,11 +3,13 @@ import React from 'react'
 const Address = ({ value }) => {
   if (!value) return null
   const parts = value.split('/')
+  // used to shift the highlight one right on p2p-circuit addresses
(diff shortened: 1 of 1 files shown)


hacdias:
Yes, will do that in another PR. Sounds perfectly good.

PR #1188, opened 2019-09-19
feat: highlight everything but protocols

hacdias:
I also added `ipfs` to the list of protocols to make sure we don't highlight it.

image

License: MIT

### DIFF
--- a/src/components/address/Address.js
+++ b/src/components/address/Address.js
@@ -1,15 +1,18 @@
 import React from 'react'
+import Multiaddr from 'multiaddr'
 
 const Address = ({ value }) => {
   if (!value) return null
+
+  const ma = Multiaddr(value)
+  const protos = ma.protoNames().concat('ipfs')
   const parts = value.split('/')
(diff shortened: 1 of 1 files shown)


src/components/address/Address.js
```diff
@@ -1,15 +1,18 @@
 import React from 'react'
+import Multiaddr from 'multiaddr'
 
 const Address = ({ value }) => {
   if (!value) return null
+
+  const ma = Multiaddr(value)
+  const protos = ma.protoNames().concat('ipfs')
```
lidel: Heads up: we are switching string representation of peer  multiaddrs from `/ipfs/` to `/p2p/` (https://github.com/libp2p/libp2p/issues/79).

In JS land, the default was switched in [js-multiaddr/releases/tag/v7.0.0](https://github.com/multiformats/js-multiaddr/releases/tag/v7.0.0).

To make this future proof, we should support both:

```suggestion
  const protos = ma.protoNames().concat(['ipfs', 'p2p'])
```

PR #1189, opened 2019-09-19
feat: show full location on hover

### LINKED ISSUE #1187: Peers screen: show full location on hover
Long Locations are truncated, but hovering on them does nothing:
> 

I think we should display full location text on hover, to match behavior from other columns.

hacdias:
Fixes #1187. Shows full location when hovering over the location cell.

image

License: MIT

### DIFF
--- a/src/peers/PeersTable/PeersTable.js
+++ b/src/peers/PeersTable/PeersTable.js
@@ -37,7 +37,7 @@ export class PeersTable extends React.Component {
   }
 
   locationCellRenderer = ({ rowData }) => (
-    <span>
+    <span title={ rowData.location ? rowData.location : this.props.t('unknownLocation')}>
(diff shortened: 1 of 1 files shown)


src/peers/PeersTable/PeersTable.js
```diff
@@ -37,7 +37,7 @@ export class PeersTable extends React.Component {
   }
 
   locationCellRenderer = ({ rowData }) => (
-    <span>
+    <span title={ rowData.location ? rowData.location : this.props.t('unknownLocation')}>
```
lidel: I think a shorter notation should also work? (but check first):
```suggestion
    <span title={ rowData.location || this.props.t('unknownLocation')}>
```

hacdias: It does!

PR #1262, opened 2019-10-20
refactor: improve peers performance

### LINKED ISSUE #1273: Peers: prioritize geoip lookup of N peers with lowest Latency
Right now, when user has thousands of peers it takes serious time before Location in the top of peer  list (sorted by latency) gets resolved on the initial load (with empty geoip cache).

This means user is looking at list of "Unknown" nodes:

> 

Possible optimization is to grab N (10, 15) peers with the lowest latency and execute lookup on that short priority list before doing lookup for all of them. OR it could be additional async thing.

hacdias:
This refactors the peers locations fetching/updating new locations every ~three~ one second~s~ so any changes will only be seen after ~three~ one second~s~. This reduced dramatically the CPU usage on my end and still keeps the functionality.

As it was stated by @olizilla on #1072:

> We do a lot of work in peer-locations.js to expose the internal logic of the queuing of geoip look-ups as state in redux. We are not visualising the queue, nor do we plan to, so i suspect the code could be written more simply. We could extend the work in #1071 to encapsulate the geoip look up queue and caching in a class, and simply have it periodically update the redux state with batches of locations as they are found, rather than have all the queuing logic managed in redux.

@lidel @autonome, can you take a look at this one, please?

----

closes #1273


hacdias:
I still need to fix the tests. However, I don't think that'll be an easy task taking in account I've changed this a lot. Perhaps I'll focus my tests on the `PeerLocationResolver` class.

src/bundles/peer-locations.js
```diff
@@ -300,44 +148,65 @@ class PeerLocationResolver {
...
-          this.geoipCache.set(ipv4Addr, data)
-          delete this.geoipLookupPromises[ipv4Addr]
-
-          if (err) return reject(err)
-          resolve(data)
+      this.queue.push(async () => {
+        // maybe we have it cached by ipv4 address already, check that.
+        const location = await this.geoipCache.get(ipv4Addr)
+        if (location) {
+          return location
```
olizilla: What happens to this location? It looks like nothing consumes the result of the queue job, so this ends up just being a little misleading for the reader. I think it's the  peersToAddrs check earlier in the function that would have catches the condition where we'e already looked up this peer.

src/bundles/peer-locations.js
```diff
@@ -12,243 +13,90 @@ export default function (opts) {
...
+    getPromise: async ({ store, getIpfs }) => {
+      const peers = store.selectPeers()
+      const locations = store.selectPeerLocations() || {}
+
+      for (const p of peers) {
+        const peerId = p.peer.toB58String()
+        const addr = p.addr.toString()
+
+        if (!locations[peerId]) {
+          const res = peerLocResolver.checkPeer(peerId, addr, getIpfs)
```
olizilla: Let's see if we can find a more intutive contract for checkPeer... it either returns a location synchronously or queues up some async work. It might be a little more intuitive if we pass the list of peers to PeerLocationsResolver, and await list of locations it can find in its cache... so the contract of the function goes from being sync or null with some async work to always async, with some async taking longer than others...

```js
const locations = await PeerLocationResolver.findLocations(peers, getIpfs)
```

It's basically moving the loop, but I'd have less async to juggle in my head.

src/bundles/peer-locations.js
```diff
@@ -300,44 +148,65 @@ class PeerLocationResolver {
       ...opts.cache
     })
 
+    this.peersToAddrs = {}
```
olizilla: Mapping `peersToAddrs` seems to short circuit the ipv4 lookup cache, and means that a peerId once seen will always be mapped to the same ip address, even if it tells us it's ip address has changed.

src/bundles/peer-locations.js
```diff
@@ -300,44 +148,65 @@ class PeerLocationResolver {
...
+    this.failedAddrs = []
```
olizilla: `failedAddrs` and `peersToAddrs` are both unbounded caches... if we need them consider using an LRU cache with a max size.

hacdias: I'm just storing them to avoid calling the same lookup over and over despite of its failure. I've changed the max size to 500, where we delete the first 100 when it gets full. I think it's not worth it to implement a full LRU cache, but it follows the principle.

src/bundles/peer-locations.js
hacdias: Fixed!

src/bundles/peer-locations.js
hacdias: Thanks for the suggestion! Just implemented!

hacdias:
@olizilla just made a few updates to fix and improve the things you've mentioned. I'll see what I can do about the animations in a different PR.

autonome:
This is awesome, thanks so much digging further into the root cause of the resource consumption.

Is there a specific reason why three seconds is the default?

hacdias:
uh, I said 3? I set it to 1. Not a specific reason, but it's much smoother now...

autonome:
It was originally three and you changed to one? I was thinking the other way... can we go to 5s or higher ;) If we had a list of specific reasons why people need to know, it would be easier to reason about instead of picking arbitrary numbers.

But as long as not burning laptop battery, 1s is fine for now!

src/bundles/peer-locations.js
```diff
@@ -12,243 +13,76 @@ export default function (opts) {
...
-      peers => {
-        if (!peers) return []
-        const allCoord = peers
-          .map(p => p.coordinates)
-          .filter(arr => !!arr)
-
-        const unique = new Set(allCoord.map(JSON.stringify))
-        return Array.from(unique).map(JSON.parse)
+    staleAfter: ms.seconds(1),
+    retryAfter: ms.seconds(1),
```
autonome: Would be good to have the interval as a const at top of file, and documented

hacdias: Just added!

hacdias:
Basically the difference will only make us see the changes later and add more requests to queue later. I don't believe it influences the amount of rendering that much, but _not sure_.

lidel:
IIUC we are fetching info for 10 locations at the same time:

```js
  // Max number of locations to retrieve concurrently
  opts.concurrency = opts.concurrency || 10
```

Did we pick this number for any particular reason?
I'd like to decrease this to 4 to be safely below current XHR/HTTP 1.1 concurrency limits in browsers (iirc its around 6 per same host in Chromium).

hacdias:
@lidel the default is 10 but we are fetching 20! It was changed on #1071

hacdias:
Green light to merge?

package.json
```diff
@@ -52,6 +52,7 @@
     "multiaddr-to-uri": "^5.0.0",
     "prop-types": "^15.7.2",
     "pull-file-reader": "^1.0.2",
+    "queue": "^6.0.1",
```
lidel: I missed this on the first pass: can we switch to [p-queue](https://github.com/sindresorhus/p-queue) instead?
We already use `p-queue` it in various other places across the stack, so it could be reused during the build + one less API to learn for potential contributor :)

src/bundles/peer-locations.js
```diff
@@ -300,44 +138,74 @@ class PeerLocationResolver {
...
+              // mark this one as failed so we don't retry again
+              this.failedAddrs.push(ipv4Addr)
 
-          if (err) return reject(err)
-          resolve(data)
+              // if there's 500 or more failed addresses, remove the
+              // oldest 100 failures.
+              if (this.failedAddrs.length >= 500) {
+                this.failedAddrs.splice(0, 100)
+              }
```
lidel: Small nit: drop-ins like this, even with good comments like this one,  make code harder to reason about :P
Let's remove this custom code by using [hashlru](https://www.npmjs.com/package/hashlru) for `failedAddrs`.
It comes with known characteristics, removes cognitive overhead (new contributor sees LRU and knows what is happening), and we already use it in js-ipfs.

hacdias:
@lidel done!

src/bundles/peer-locations.js
```diff
@@ -300,44 +140,67 @@ class PeerLocationResolver {
...
-          this.geoipCache.set(ipv4Addr, data)
-          delete this.geoipLookupPromises[ipv4Addr]
+          geoip.lookup(ipfs, ipv4Addr, (err, data) => {
+            delete this.geoipLookupPromises[ipv4Addr]
 
-          if (err) return reject(err)
-          resolve(data)
+            if (err) {
+              // mark this one as failed so we don't retry again
+              this.failedAddrs.set(ipv4Addr, 1)
```
lidel: `true` looks better than a magic number :)
```suggestion
              this.failedAddrs.set(ipv4Addr, true)
```

hacdias: Thanks!

hacdias:
Both of those are now fixed @lidel!

lidel:
Hm, for some reason map and location column [are not updated now](https://user-images.githubusercontent.com/157609/67411437-eddfc500-f5bd-11e9-90f4-8d5ec1fbb800.png). 
@hacdias are you able to reproduce? (I got the same result in Firefox and Brave)

hacdias:
Can't reproduce on Firefox, nor Chrome.

hacdias:
@lidel any update on this?

src/bundles/peer-locations.js
```diff
@@ -300,44 +141,67 @@ class PeerLocationResolver {
...
 
-    if (!locPromise) {
-      locPromise = new Promise((resolve, reject) => {
-        const ipfs = getIpfs()
-        geoip.lookup(ipfs, ipv4Addr, (err, data) => {
-          // save it in the caches, and remove the promise.
-          this.geoipCache.set(ipv4Addr, data)
-          delete this.geoipLookupPromises[ipv4Addr]
+          geoip.lookup(ipfs, ipv4Addr, (err, data) => {
+            delete this.geoipLookupPromises[ipv4Addr]
```
lidel: We remove it here, but can't find a place where we add it :^)

src/bundles/peer-locations.js
```diff
@@ -300,44 +141,67 @@ class PeerLocationResolver {
...
+    for (const p of peers) {
+      const peerId = p.peer.toB58String()
+      const addr = p.addr.toString()
 
-    // maybe we have it cached by ipv4 address already, check that.
-    let location = await this.geoipCache.get(ipv4Addr)
-    if (location) {
-      return location
-    }
+      const ipv4Tuple = Multiaddr(addr).stringTuples().find(isNonHomeIPv4)
```
lidel: I think we can just use multiaddr instance directly?
(Or was this a workaround for some bug?)

```suggestion
      const ipv4Tuple = p.addr.stringTuples().find(isNonHomeIPv4)
```

src/bundles/peer-locations.js
hacdias: It's here: https://github.com/ipfs-shipyard/ipfs-webui/pull/1262/files/0e15b969fbceb9c6a344b246d0f784d8f50db96e#diff-74ea233092a3be56e73ee19c79bc2064R151 :o

lidel:
I broke tests, oops. (will fix shortly)

src/bundles/peer-locations.js
lidel: I meant we are are deleting things but we never set them.
I think we need:

```js
      this.geoipLookupPromises[ipv4Addr] = this.queue.add(async () => {
        return new Promise(resolve => {
           // ...
        })
      })
```

hacdias: oh, yeah, definitely! thanks for noticing that. Can you add that when fixing the testS?

lidel: Yup, added in https://github.com/ipfs-shipyard/ipfs-webui/pull/1262/commits/eeccc8ae651c089343f8fa66aa3cba7d682e2bbe

PR #1265, opened 2019-10-21
fix: disable launch on startup on unsupported platforms

hacdias:
Disables launch on startup on OSes that are not Windows, macOS or Linux.

License: MIT

### DIFF
--- a/src/settings/DesktopSettings.js
+++ b/src/settings/DesktopSettings.js
@@ -40,6 +40,7 @@ export function DesktopSettings ({ t, doDesktopSettingsToggle, desktopSettings }
 
       <CheckboxSetting checked={desktopSettings.autoLaunch || false}
         title={t('launchOnStartup')}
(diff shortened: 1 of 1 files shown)


src/settings/DesktopSettings.js
```diff
@@ -40,6 +40,7 @@ export function DesktopSettings ({ t, doDesktopSettingsToggle, desktopSettings }
 
       <CheckboxSetting checked={desktopSettings.autoLaunch || false}
         title={t('launchOnStartup')}
+        disabled={!['win32', 'darwin', 'linux'].includes(os.platform())}
```
lidel: I think checkbox should be disabled in dev too ([because feature already is](https://github.com/ipfs-shipyard/ipfs-desktop/pull/1200/files#diff-7950c0473bee02498f00f379e35532bcR68-R70))

```suggestion
        disabled={!(['win32', 'darwin', 'linux'].includes(os.platform()) && process.env.NODE_ENV !== 'development')}
```

Issue #223, opened 2016-01-26
Highlight peers in Local Area Network and in close IP space proximity


Kubuxu:
It would be awesome if WebUI highlighted peers that are close to you. Finding peer that is close is something that shows adaption of IPFS what is its aim.

As IP is prefix allocated (apart from cjdns) we can create following rules:
- `peerIP4 ∈ 10.0.0.0/8` - private network (RFC 1918)
- `peerIP4 ∈ 172.16.0.0/12` - private network (RFC 1918)
- `peerIP4 ∈ 192.168.0.0/16` - private network (RFC 1918)
- `peerIP4 ∈ 169.254.0.0/16` - link local (RFC 3927)
- `peerIP6 ∈ fd00::/8` - unique local address (RFC 4193)
- `peerIP6 ∈ fe80::/10` - link local (RFC 4291)
- `peerIP6 ∈ myIP6/64` - smallest end site (RFC 6177)

Also worth considering, maybe with different accent:
- `peerIP4 ∈ myIP4/24` - 255 closest users according to spatial allocation
- `peerIP6 ∈ myIP6/48` - most end sites will be allocated this prefix

where `peerIP6 ∉ fc00::/8` to fix case of cjdns where IPs are not spatial allocated.

olizilla:
This would be very cool. Related work is to add tool-tips to on hover on the peers map in #819

hacdias:
We've done it! it was released a long time ago, but here it is: #1266

PR #1266, opened 2019-10-21
feat: highlight local network peers

hacdias:
**Based on #1262 so that must be merged beforehand.** This PR introduced a special location for local network peers, where show their flag as a handshake with the 'Local Network'. This idea comes from #223 which suggests to highlight local network and close peers geographically.

I wonder if this is the best way to show local network peers. How'd highlight geographically close peers?

image

/cc @lidel @autonome


hacdias:
I thought maybe coloring the row with a different background for local and geographically close peers, but that wouldn't say much by itself. Maybe some badge near the location?

lidel:
@hacdias are you able to create screenshots with 2-3 options? (eg. yellowish background / text color / badge). Will be easier to evaluate that way. 
(including design shop for feedback)

hacdias:
Keeping the emoji for the Local Network ones, we could use this colors (gold for local network, silver for nearby geographically):

image

Same, but with light yellow for local network and light blue for nearby geographically:

image

Keeping the emoji for the Local Network ones, using the color scheme gold for local network and blue for nearby geographically:

image

Last but not the least, with badges! In this case, I don't believe keeping the Local Network emoji is worth it because we have badges!

image

What do you like most? I really love the 'Local Network' with a handshake. I'd personally go to either one of the last two.

ericronne:
Love the handshake, too!

Given that this indicator is all about proximity, i think it makes sense to use a small label, as opposed to stretching it across the entire row. It also seems useful to explicitly mark them as "local" or "nearby," so we don't have to rely on a key, eg.

In the case of local connections, what if we swapped out the globe for the handshake, and swapped out "Unknown" for "Local"? In theory, the user would know the location of a locally connected peer, right? 😸 

x

For nearby nodes, i've simply added "(nearby)" to the end of the nearby node (in gray) below. Not a brilliant solution, admittedly. We could add a small emoji, perhaps …

a

lidel:
Thanks, looking at it changing text/background color makes things too noisy.

I am also worried about confusion between "local" (same IP network class) and "nearby" (geographically).

I'd go with regular colors, handshake icon and probably used full "Local Network". In theory, addresses from "local network" could not be geographically "local" (VPN-bridged LAN networks etc).

hacdias:
image

Just updated to match the suggestions above. 'Local Network' + Handshake for local network nodes and nearby between parenthesis on geographically close peers.

hacdias:
Ping @lidel @autonome @ericronne

src/bundles/peer-locations.js
```diff
@@ -114,6 +124,44 @@ const parseLatency = (latency) => {
...
+    if (myIP) {
+      if (addr.family === 4) {
+        isNearby = ip.cidrSubnet(`${myIP}/24`).contains(addr.address)
+      } else if (addr.family === 6) {
+        isNearby = ip.cidrSubnet(`${myIP}/48`).contains(addr.address) &&
+          !ip.cidrSubnet('fc00::/8').contains(addr.address)
+        // peerIP6 ∉ fc00::/8 to fix case of cjdns where IPs are not spatial allocated.
+      }
+    }
+  } catch (_) {
```
autonome: What exactly might throw here? And are the exceptions related to isPrivate/isNearby?

The try block makes it difficult to see whether or not there's a possibility for a null value to be returned (assuming that's not ok).

I recommend narrowing the try block(s) around specific exceptions and documenting what they are.

Maybe also initialize the values to false to guarantee a safe/usable return value (seems safer, from quick glance), instead of as a side-effect of maybe-related errors.

src/bundles/peer-locations.js
```diff
@@ -48,31 +50,39 @@ export default function (opts) {
...
-        coordinates,
-        connection,
-        address,
-        latency,
-        notes
-      }
-    })
+    'selectIdentity',
+    (peers, locations = {}, bootstrapPeers, identity) => {
+      const myIP = getPublicIP(identity)
```
autonome: `myIP` appears to only be used by `isPrivateAndNearby`, and not in this function (except to pass to `isPrivateAndNearby`).

Maybe better to memoize the return value and localize the usage by passing in `identity` to that function and call `getPublicIP` from there instead?

src/bundles/peer-locations.js
```diff
@@ -114,6 +124,44 @@ const parseLatency = (latency) => {
   return value
 }
 
+function getPublicIP (identity) {
+  if (!identity) return
```
autonome: The callers of this don't explicitly handle a null return value. Is there a reasonable expectation there'll be no identity at this point?

src/bundles/peer-locations.js
hacdias: Done!

src/bundles/peer-locations.js
hacdias: It should, usually. Just updated the function!

src/bundles/peer-locations.js
hacdias: I just updated the function. Please take a look now :D

hacdias:
@autonome just updated the function with memoization (from a package though) and added comments!

src/bundles/peer-locations.js
autonome: Much clearer, thank you!

PR #1312, opened 2019-11-14
fix: remove typo in 'IPFS is a web' description on Files screen

colinfruit:
Small typo that I noticed.


hacdias:
Thanks @colinfruit!

PR #1334, opened 2019-11-28
fix: geoip and peers on map

### LINKED ISSUE #1313: Peers in v2.7.1  no longer get rendered on the map
WebUI on the left: v2.6.0, right: v2.7.1
(tested in empty profile)

I see the same issue in Firefox 70 and Brave 0.70.104
Difference between versions is small ([diff](https://github.com/ipfs-shipyard/ipfs-webui/compare/v2.6.0..v2.7.1)), not sure what could cause this, looks like some async code failing silently.

No bandwidth to look into this, but some ideas:
- perhaps bumping `chart.js` to [v2.9.3](https://github.com/chartjs/Chart.js/releases/tag/v2.9.3) would help?
- code using `p-queue`?

hacdias:
image

Fixes #1313.

License: MIT

### DIFF
--- a/src/bundles/peer-locations.test.js
+++ b/src/bundles/peer-locations.test.js
@@ -126,10 +126,8 @@ const createMockIpfs = (opts) => {
 
   return {
     object: {
-      get: (_, cb) => {
-        setTimeout(() => {
-          cb(null, fakeGeoIpData())
-        }, randomInt(opts.minLatency, opts.maxLatency))
+      get: async () => {
+        return fakeGeoIpData()
       }
     }
   }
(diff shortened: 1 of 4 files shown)


hacdias:
Edit: not understanding why some tests are failing even though it works.

src/bundles/peer-locations.test.js
```diff
@@ -97,7 +97,7 @@ const mockPeersBundle = {
 
 function fakeGeoIpData () {
   return {
-    data: JSON.stringify({
+    Data: JSON.stringify({
```
lidel: FYSA this is why tests always failed (always returned no data)

PR #1351, opened 2019-12-04
Allow vertical scroll in the navbar

jaller94:
This allows to scroll the navbar when your window is not tall enough to display the entire length.

ipfs-scroll-navbar


chandi1988:
> This allows to scroll the navbar when your window is not tall enough to display the entire length.
> 
> ipfs-scroll-navbar

Issue #1164, opened 2019-09-14
Improve Test Suite


hacdias:
We should aim to improve the test suite and make it more throughout. There should be E2E tests with both `go-ipfs` and `js-ipfs` implementations to catch eventual differences in both implementations.

This is a draft issue.

lidel:
I believe at minimum we should run end-to-end tests for adding files and directories via Web UI in specific environments.

The test matrix  would be to test adding files in each runtime (browsers + ipfs-desktop) against each implementation:

|  | go-ipfs |  js-ipfs |
| ---- | ---- | ---- |
| chromium | ? | ? | 
| firefox | ? | ? |
| electron | ? | ? |


Sidenotes:
- In past we had issues in go-ipfs that surfaced with specific file sizes, so we should test 1/10/100/500/1000MB file sizes
- Directory test should use a tree at least three levels deep to ensure there are no regression in custom logic present in webui
- I would also like to have E2E for companion+js-ipfs+chrome.sockets 
  (JS IPFS in Brave: https://github.com/ipfs-shipyard/ipfs-companion/issues/716)

lidel:
I am taking a look at this today.

hacdias:
@lidel cool, let me know of any progress you made. I was also thinking of tackling E2E tests on this

lidel:
I've tried to see how things look like when we add aegir (for running tests in different runtimes). 
It costs about +200MB in `node_modules` (1.1G without vs 1.3G with),
and does not solve the problem of testing against different runtimes. See below.

### Current State

Things get complicated fast, because aegir requires tests written in [Mocha](https://github.com/mochajs/mocha) and [Karma](https://github.com/karma-runner/karma) for browser tests (Firefox, Chrome, Electron), while Web UI uses [Jest](https://jestjs.io) for tests and [Puppeteer](https://github.com/GoogleChrome/puppeteer) for end-to-end tests in Chrome.

End-to-end options these days seem to be limited to `puppeteer`-compatible APIs, which means Chromium, Firefox ([80% complete](https://aslushnikov.github.io/ispuppeteerfirefoxready)) and (maybe) Electron (experimental [puppeteer-electron](https://github.com/nondanee/puppeteer-electron)).



#### (A) Move to Mocha/Karma

IMO does not make much sense, refactoring cost may be high and projects usually go in the opposite direction: Jest is believed a successor to Mocha-based tests. 
[...]

PR #1353, opened 2019-12-05
feat: E2E test suite with puppeteer+ipfsd-ctl

lidel:
> Parent: #1164 
> cc @hugomrdias @hacdias @autonome for visibility/early feedback

This PR adds E2E tests for WebUI that run against HTTP API from real IPFS node

Try with:

```console
$ npm run test:e2e # headless against go-ipfs
```

### Implementation details

The end-to-end tests (e2e) test the full app in a headless Chromium browser. They spawn real IPFS node for HTTP API and a static HTTP server to serve the app.
The purpose of those tests is not being comprehensible, but act as a quick regression and integration suite.

Make sure `npm run build` is run before starting E2E tests:

```console
$ npm run build
$ npm run test:e2e # end-to-end smoke tests (fast, headless, use go-ipfs)
```

### Customizing E2E Tests

Default behavior can be tweaked via env variables below.

#### `E2E_IPFSD_TYPE`

Variable named `E2E_IPFSD_TYPE` defines which IPFS backend should be used for end-to-end tests.

CI setup of ipfs-webui repo runs tests against both JS and GO implementations:

```console
$ E2E_IPFSD_TYPE=go npm run test:e2e # 'go' is the default (if missing)
$ E2E_IPFSD_TYPE=js npm run test:e2e
```

It is possible to test against arbitrary versions by tweaking `ipfs` (js-ipfs)
 and `go-ipfs-dep` (go-ipfs) in `devDependencies` section of `package.json` and applying the change via `npm i`

#### `E2E_API_URL`

Instead of spawning a disposable node and repo for tests, one can point the E2E test suite at arbitrary HTTP API running on localhost:

```console
$ E2E_API_URL=http://127.0.0.1:5001 npm run test:e2e
```

**Caveat 1:** HTTP API used in tests needs to run on the local machine for Peers screen to pass (they test manual swarm connect to ephemeral `/ip4/120.0.0.1/..` multiaddr)

**Caveat 2:** CORS requests from `http://localhost:3001` (static server hosting dev version of webui) need to be added to `Access-Control-Allow-Origin` whitelist array in node's config:

```json
"API": {
  "HTTPHeaders": {
    "Access-Control-Allow-Origin": [
      "http://localhost:3001"
    ]
  }
}
```

Can be done ad-hoc via command line:

```console
$ ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin '["http://localhost:3001"]'
$ ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods '["PUT", "GET", "POST"]'
```

#### Show the browser

By default, the test run headless, so you won't see the browser. To debug test errors, it can be helpful to see the robot clicking around the site.
To disable headless mode and see the browser, set the environment variable `DEBUG=true`: 

```console
$ DEBUG=true npm run test:e2e # e2e in slowed down mode in a browser window
```

#### Breakpoints

It is possible to set a "breakpoint" via `await jestPuppeteer.debug()` to stop tests at a specific line:

```js
jest.setTimeout(600000) // increase test suite timeout
await jestPuppeteer.debug() // puppeteer will pause here
```

In a **continuous integration** environment we lint the code, run the unit tests, build the app, start an http server and run the unit e2e tests:

```sh
> npm run lint
> npm test
> npm run build
> npm run test:e2e
```

### TODO

- [x] `npm run test:e2e` should run headless e2e tests against real go-ipfs
- [x] `DEBUG=true npm run test:e2e` should show browser window
- [x] `ipfs` (API interface for real node) should be in scope of tests
- [x] add mechanism and docs for swapping IPFS node used for tests (`E2E_IPFSD_TYPE`)
- [x] tests
  - [x] Status page (Peer ID)
  - [x] Files page (multiple files at once)
  - [ ] Files page (directory)
	- I was unable to find a way for testing this, but previous test should be enough as it mostly uses the same code paths.
  - [x] Explore page
  - [x] Peers page
  - [x] Settings page
- [x] switch to ipfsd-ctl v1
- [x] ensure it runs js/go from `node_modules` 
- [x] enable linter for `test/e2e`
- [x] add option to point at remote HTTP API (`E2E_API_URL`)


test/e2e/e2e.test.js
```diff
@@ -0,0 +1,172 @@
...
+const waitForTitle = title => page.waitForFunction(`document.title === '${title}'`, { timeout: 8000 })
+
+describe('Status page', () => {
+  beforeAll(async () => {
+    await page.goto(landingPage)
+    // get multiaddr of http api
+    const { apiHost, apiPort } = global.ipfs
+    const apiMultiaddr = `/ip4/${apiHost}/tcp/${apiPort}`
+    // point webui at real api
+    await page.evaluate((apiMultiaddr) => localStorage.setItem('ipfsApi', apiMultiaddr), apiMultiaddr)
```
autonome: Can you split into separate lines so is clear you're storing?

test/e2e/e2e.test.js
```diff
@@ -0,0 +1,172 @@
...
+    // apply settings
+    await page.goto(landingPage)
```
autonome: Are the settings instant-apply and you're just navigating back to start to reset for next test? Or is navigation change triggering the settings being saved?

Can you update comment to make clear which one is happening?

test/e2e/e2e.test.js
```diff
@@ -0,0 +1,172 @@
...
+    await page.click(button)
+    await page.waitForSelector('#add-file', { visible: true })
+
+    const [fileChooser] = await Promise.all([
+      page.waitForFileChooser(),
+      page.click('a[id="add-file"]'), // menu button that triggers file selection
+    ])
+
+    // lets add a static text file from the root of this repo
+    const filename = 'LICENSE'
```
autonome: Seems like this would break if we change the location of the tests, eg into a subdir. Maybe just commit a test file to use explicitly for this purpose?

test/e2e/e2e.test.js
lidel: `ipfsApi` is read from localStorage during webui "boot" done by [ipfs-redux-bundle](https://github.com/ipfs-shipyard/ipfs-redux-bundle).
I'll comment and move it to `test/e2e/setup/test-environment.js` so its done automatically before each test file.

test/e2e/e2e.test.js
```diff
@@ -0,0 +1,172 @@
+const fs = require('fs')
+
+const landingPage = 'http://localhost:3001'
+const waitForTitle = title => page.waitForFunction(`document.title === '${title}'`, { timeout: 8000 })
+
+describe('Status page', () => {
```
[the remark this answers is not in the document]
lidel: yes, splitting into one `.test.js` per screen is on the TODO list

lidel:
Ok, I've split tests into separate files but kept them small to test as much they can  while still easy to maintain.
When merged, I'll PR js-ipfs repo to run external webui E2E with js-ipfs backend. 

@hugomrdias I used new ipfsd-ctl  in `test/e2e/peers.test.js` and ` test/e2e/setup/global-init.js `  – would appreciate sanity check, thanks!

test/e2e/setup/global-init.js
```diff
@@ -0,0 +1,47 @@
...
+      type,
+      test: true, // sets up all CORS headers required for accessing HTTP API port of ipfsd node
+      overrides: { // call findBin here to ensure we use version from devDependencies, and not from ipfsd-ctl
+        js: { ipfsBin: findBin('js') },
+        go: { ipfsBin: findBin('go') }
+      }
+    })
+    ipfsd = await factory.spawn()
+    ipfs = ipfsd.api
+  }
```
hugomrdias: damn this is totally wrong we need to talk so i can improve the ctl docs :) 
- you only need to control one node type per test run ? if yes just use `createController()`
- overrides are the in the second param https://github.com/ipfs/js-ipfsd-ctl#createfactoryoptions-overrides 
- you can do the endpoint === true with ctl also using `createController({remote: true, endpoint: endpoint, test: true, type: process.env.E2E_IPFSD_TYPE || 'go', ipfsBin: findBin(process.env.E2E_IPFSD_TYPE || 'go')})`

lidel: @hugomrdias: thanks for the tips, I fixed first two points in this PR and opened PR to improve ipfsd-ctl docs in https://github.com/ipfs/js-ipfsd-ctl/pull/421 

As for the third point about remote endpoint, I don't think the `createController({remote: true, endpoint: process.env.E2E_API_URL })` will work, because `E2E_API_URL` is the IPFS API, not the one provided by `Ctl.createServer`. 

Running `createController({remote: true, endpoint: 'http://127.0.0.1:5001' })` produces error (probably due to 404 for `/api/v0/..`):

```
HTTPError: Not Found
    at fn (/home/lidel/project/ipfs-webui/node_modules/ky/umd.js:337:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

Let me know if I got this wrong.

lidel:
@hugomrdias applied your suggestions (details in https://github.com/ipfs-shipyard/ipfs-webui/pull/1353#discussion_r363278418). Should be good enough.
I'm going to merge this to unblock integration with js-ipfs and go-ipfs CI processes, we can improve things later in separate PRs.

test/e2e/setup/global-init.js
hugomrdias: if aren't using createServer you shouldn't need ctl only http-client right ?

lidel: Correct. We don't use `createServer` in e2e tests.
If `E2E_API_URL` is defined it means HTTP API got spawned by a third party.

PR #1407, opened 2020-02-25
fix(files): do not double-decode paths in infoFromPath

sinkuu:
`infoFromPath` decodes the given path. If the path is already decoded, it will be re-decoded and breaks MFS paths like `/dapps/http/127.0.0.1%3A8080/ipns` (IPFS companion creates this).

This problem can be reproduced by running:

```
ipfs files mkdir /%3A/foo -p
```

and then visit `#/files/%253A/foo` in Web UI. (It says "Failed to get those files.")

This PR fixes the problem by not decoding the path that is already decoded.


hacdias:
Thanks for this @sinkuu! And sorry for taking so long

PR #1430, opened 2020-04-04
fix: support POST-only HTTP API

### LINKED ISSUE #1429: Support POST-only HTTP API
PR at https://github.com/ipfs/go-ipfs/pull/7097 will block `GET` on `/api/v0` on API port.

Seems that we need to fix ipfs-webui and release a new version so mentioned PR can be merged.

cc @hsanjuan @Stebalien 

### Known problems

- [x] Files screen fails to load - detected by E2E tests https://circleci.com/gh/ipfs/go-ipfs/28451
     `/api/v0/ls` seems to be sent as `GET`:
     > files-2020-04-04--19-28-38

lidel:
### Problem

https://github.com/ipfs/go-ipfs/pull/7097 will block `GET` commands on API port, requiring everything to be a POST request.

This breaks _Files_ screen in ipfs-webui as noted in https://github.com/ipfs-shipyard/ipfs-webui/issues/1429

### Even a bigger problem

ipfs-webui is using older version of js-ipfs-http-client, one before huge refactor into async iterables, which means switching to the latest version won't be a trivial task, as  the amount of code and dependencies that needs to be refactored is unknown.

### Solution

Due to time constraint I made a decision to patch it in npm `postinstall` so we can release this ASAP and ship with go-ipfs 0.5.

`postinstall` script will apply a simple patch on top of ipfs-http-client v39.0.2 to ensure it sends commands as POST

Proper fix will land when ipfs-webui is refactored to work with ipfs-http-client >41.x

Closes #1429

cc @achingbrain  for visibility

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Run the **[cors-config.sh](./cors-config.sh)** script with:
 
 ```sh
 > ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin '["http://localhost:3000", "https://webui.ipfs.io"]'
-> ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods '["PUT", "GET", "POST"]'
+> ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods '["PUT", "POST"]'
 ```
 
 #### Reverting
--- a/cors-config.sh
+++ b/cors-config.sh
@@ -6,7 +6,7 @@ ALLOW_ORIGINS='"http://localhost:3000", "https://webui.ipfs.io"'
 set -e
 
 ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin "[$ALLOW_ORIGINS]"
-ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods '["PUT", "GET", "POST"]'
+ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods '["PUT", "POST"]'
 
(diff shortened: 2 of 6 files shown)


lidel:
@Stebalien @hsanjuan 
I tested this against https://github.com/ipfs/go-ipfs/pull/7097 and Files screen is fixed.
If it is okay with you I'd like to merge this and make a new release of ipfs-webui, so we can update go-ipfs to use it and also fix E2E tests in https://github.com/ipfs/go-ipfs/pull/7097

lidel:
@Stebalien I am not sure if patch release will be possible due to recent move to monorepo.

I'd rather not bother @achingbrain with it. Keeping this fix local to webui decreases noise: everyone should be already on the latest js-ipfs-http-client  >41.x anyway, its just ipfs-webui lagging behind and we have no bandwidth to upgrade it before go-ipfs 0.5.

hsanjuan:
Sounds good, I trust your judgment.

lidel:
I created https://github.com/ipfs-shipyard/ipfs-webui/issues/1431 to track proper upgrade.

I'll merge this and prepare a new release.

Stebalien:
@lidel thinking about this a bit, we should _really_ release a patch release of the HTTP API. Upgrading to async/await is non-trivial so there are likely dapps other than ipfs-webui in the same boat.

lidel:
@Stebalien I agree, I hit into more problems while working on Companion. 

Created https://github.com/ipfs/js-ipfs/issues/2971 and marked it as P0 to make sure POST-only API is properly handled on JS side, including patch release.

PR #1440, opened 2020-04-17
update zh-CN translation, modify the wording of Chinese translation.

bobwxc:
modify the wording of Chinese translation
有些措辞翻译的很奇怪，小修一下


jessicaschilling:
Thank you for contributing translation @bobwxc! 👍

To avoid duplicated work, we don't accept direct PRs with locales.

Did you submit those translations using the Transifex project at https://www.transifex.com/ipfs/ipfs-webui/ ? If not, let us know your Transifex account name, and can add you to the zh-TW translation team so you can contribute there.

FYI, there is no need for submitting PRs via GitHub: updating them on Transifex is enough. We synchronize translations with Transifex before each release of ipfs-webui.

You can find more information at README#translations and ipfs/i18n. Thanks!

PR #1443, opened 2020-04-18
fix: console error when no API available

lidel:
This PR removes error in console  when there is no API available:

error-when-down-2020-04-18--01-56-26

### DIFF
--- a/src/welcome/WelcomePage.js
+++ b/src/welcome/WelcomePage.js
@@ -30,7 +30,7 @@ const WelcomePage = ({ t, doUpdateIpfsApiAddress, apiUrl, ipfsInitFailed, ipfsCo
             </Trans>
             <ApiAddressForm
               t={t}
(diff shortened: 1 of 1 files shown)


src/welcome/WelcomePage.js
```diff
@@ -30,7 +30,7 @@ const WelcomePage = ({ t, doUpdateIpfsApiAddress, apiUrl, ipfsInitFailed, ipfsCo
             </Trans>
             <ApiAddressForm
               t={t}
-              defaultValue={ipfsApiAddress}
+              defaultValue={ipfsApiAddress || ''}
```
rafaelramalho19, line 33: You should probably do
```
WelcomePage.defaultProps = {
   ipfsApiAddress: ''
}
```

Just to become more bulletproof of future uses (and also to avoid the `||` logic on each re-render, since defaultProps is only handled in the first render.

PR #1444, opened 2020-04-18
feat: add hover and clusters to peers map

rafaelramalho19:
## Notes

- I wasn't sure what information to place in the popover, so I just used the country and id for now. If you find any more relevant information I should place, please say so 🙏 

- I couldn't find a way to properly make a reducer work with `createAsyncResourceBundle`, the documentation seemed a bit lacking and I had some really hard time trying to figure out what happened. After reading through the source code I found out that `createAsyncResourceBundle` actually creates a reducer function to handle all the async mechanism mutations, so I extended it. You can see this in src/bundles/peers.js#13
If someone could help me find a better way, I would be very much appreciated 😄 

- Added some transitions to smoothen the interaction with several popovers, as you can see here: 

ezgif com-video-to-gif

## Questions

1. I believe that most of the times the peers will be too close to each other (at least on the map scale) and hovering through them can be weird/glitchy. Should we change the map scale or add buttons to zoom?

2. Should we scroll the list to the hoverable peer when it's out of view?

3. Should we show the popover when we also hover an element from the list?

PS: Sorry for the long merge request description 😅


jessicaschilling:
This is awesome! My thoughts on your questions but @hacdias and @lidel should weigh in too 😊

1. Zoom buttons would be really cool
2. So would scrolling
3. If we have zoom buttons, and we scroll through the list to a region that's out of view on the map, the popover might not be in view. That might be OK, but soliciting others' opinions?

hacdias:
I love this and I agree with @jessicaschilling about scrolling.

lidel:
@rafaelramalho19  I really like the idea of map popover!

May be a bit more useful if popover for a standalone peer includes its full multiaddr (ok to keep the shortened PeerID).. and maybe latency too?

In other words, instead of "`12D3..y8PK`" 
something like: "`/ip4/13.40.98/tcp/4001/p2p/12D3..y8PK` (130ms)" 

FYSA multiaddr prefix is already displayed on hover at [connection column](https://user-images.githubusercontent.com/157609/79755884-17359f80-831a-11ea-9eee-ee49ee2ff68f.png), if you want to reuse the data source.

----

Answering questions:

(1) _Should we change the map scale or add buttons to zoom?_

Adding zoom would not fully solve the problem, because no matter how much you zoom, you will always end up with situation when there is more than one person from a big city, and they all are assigned the same coordinate on the map (geoip resolution we use is limited to the  "city" level). 

I don't think we should add zoom in this PR (it could be something for the future / separate PR). Instead, we should make the map smarter if there is more than one peer in specific area (eg. within the size of the dot itself). 

Loose idea: merge those dots into a single (bigger?) one and show a popover with a peer count for the area. Could still have a list of merged peers below the count.

This would make it easy to hover mouse on parts of the USA, see how many of my peers are from NYC, Tokio. Right now it gets messy when there are many peers (some people tweak settings and run webui with 5000 peers or more!), no way to tell how much bigger US is than EU, especially when viewport is small (could be 10x more peers, but both look the same):

> 


As for (2) and (3):
_Should we scroll the list to the hoverable peer when it's out of view?_
_Should we show the popover when we also hover an element from the list?_

I think those are great improvements for UX, but should be implemented together: if one works, user will expect it working the other way around. 

(I suspect it may be easier to implement without zoom controls, as there is no need to account for edge cases when zoomed in and hovering something outside of map viewport)

rafaelramalho19:
Good suggestion @lidel, I will merge the dots into a bigger one and share the result here.

I'm actually having problems with paning and zooming with d3 + our react setup, so your suggestion actually helps me a lot 😄

jessicaschilling:
+1 to everything @lidel said. Thanks both for thinking through this in such detail 😊

jessicaschilling:
Following up on the bi-weekly call to add a reference to the NY Times coronavirus map illustrating relative size of dots with numbers attached to them. This is overkill for our case, but could be useful inspiration: https://www.nytimes.com/interactive/2020/us/coronavirus-us-cases.html

rafaelramalho19:
New update:
- Grouped the peers in a range of 8 dots together, so they look like this now:
image

- Added labels to indicate how many peers there are per dot type.
image

- The hovering effect now contains a `+X` label indicating that there are more than 5 peers in a circle 
image

- The hover effect updates all peers in a circle:
Screen-Recording-2020-04-21-at-18 35 42

@jessicaschilling @hacdias @lidel what you think of this approach?

jessicaschilling:
This is so cool 😊

A few cosmetic suggestions, assuming @lidel @hacdias are OK with this approach:
- Could the key please go underneath the map as a whole? Otherwise it looks like we have a bunch of peers in Antarctica. We should probably precede the row with the word "Key:", too.
- In the hover, the latency is a little higher up than the peerID. Also, can the latency figure please be in gray, to differentiate it from the rest of the peerID?
- In the hover, if the "+X" label is to indicate the total number of peers "contained in a circle", we should be more explicit that's the total. Assuming that the top 5 we are seeing in the hover are the top 5 in the list as sorted by the user's sort choice (latency in your video), one suggestion might be:
   - In the area where you currently have "+X", instead a header in the same teal uppercase as the table sort headers: "Top X by {sort choice}"
   - At the bottom of the list of 5, justified right, if there are more than 5: "Plus XXX more peers"

Thanks!

src/bundles/peer-locations.js
```diff
@@ -82,16 +82,41 @@ export default function (opts) {
...
     'selectPeerLocationsForSwarm',
     peers => {
       if (!peers) return []
-      const allCoord = peers
-        .map(p => p.coordinates)
-        .filter(arr => !!arr)
 
-      const unique = new Set(allCoord.map(JSON.stringify))
-      return Array.from(unique).map(JSON.parse)
+      return peers.reduce((previous, { peerId, coordinates }) => {
```
rafaelramalho19, line 92: Note to self: Missing tests once UI and UX is reviewed

rafaelramalho19:
> This is so cool 😊
> 
> A few cosmetic suggestions, assuming @lidel @hacdias are OK with this approach:
> 
> * Could the key please go underneath the map as a whole? Otherwise it looks like we have a bunch of peers in Antarctica. We should probably precede the row with the word "Key:", too.
> * In the hover, the latency is a little higher up than the peerID. Also, can the latency figure please be in gray, to differentiate it from the rest of the peerID?
> * In the hover, if the "+X" label is to indicate the total number of peers "contained in a circle", we should be more explicit that's the total. Assuming that the top 5 we are seeing in the hover are the top 5 in the list as sorted by the user's sort choice (latency in your video), one suggestion might be:
>   
>   * In the area where you currently have "+X", instead a header in the same teal uppercase as the table sort headers: "Top X by {sort choice}"
>   * At the bottom of the list of 5, justified right, if there are more than 5: "Plus XXX more peers"
> 
> Thanks!

Thanks for the suggestions Jessica,

Relative to the last point, I didn't sort the elements in the hover with the same logic as in the table because that added an extra layer of overhead and complexion. 
However, If it's something you really see improving the experience, I can do it 😄 .

Now, about the other points you referred, could you please revise them?
* Key under the map:

image

* Latency grayed & Plus XXX more peers: 
image

jessicaschilling:
This is great! I'd like to get @lidel's blessing but I think this is enough to send out into the world. Ideally we should sort the contents of the hover boxes, but I'll let @lidel weigh in on whether that's worth the effort at this stage.

src/bundles/peer-locations.js
```diff
@@ -82,19 +82,40 @@ export default function (opts) {
     })
   )
 
-  bundle.selectPeerCoordinates = createSelector(
+  const COORDINATES_RADIUS = 8
```
lidel, line 85: The map will feel bit empty on desktop, where users have more strict connection limits by default.

Quick fix: I propose a tweak to decrease range from 8 to 4:

```suggestion
  const COORDINATES_RADIUS = 4
```

### radius 8 (before)
empty-Screenshot_2020-04-23 Peers - IPFS(1)

### radius 4 (after)
Screenshot_2020-04-23 Peers - IPFS

----

I believe we also should tweak the Index range to be optimized for ipfs-desktop users, which [will have between 50 and 300 peers](https://github.com/ipfs-shipyard/ipfs-desktop/blob/47e6306db586158a835dcf09cce6fa42279d8c5f/src/daemon/config.js#L46-L47).

Perhaps change the Index ranges to 1-10, 10-100, 100+ ?

rafaelramalho19:
Thanks for the suggestions @lidel, here's how it looks so far:
Screen Recording 2020-04-23 at 18 35 34 (1)

* Also added a transition when new peers appear to smoothen the experience 🚀 

I would love to work again on this later and improve responsiveness (mobile needs a lot of love in the map)

src/bundles/peer-locations.js
```diff
@@ -82,16 +82,41 @@ export default function (opts) {
     })
   )
 
-  bundle.selectPeerCoordinates = createSelector(
+  const COORDINATES_RADIUS = 4
```
lidel, line 85: I think I finally understand why it looked odd to me :sweat_smile: 
Radius was much bigger than the biggest circle, so there was too much whitespace between dots in east coast of US, Europe and China.  I expect those areas to be _denser_, so whitespace between dots should be minimal.

(it may be subjective, so feel free to ignore below suggestion)

```suggestion
  const COORDINATES_RADIUS = 2
```

src/peers/WorldMap/WorldMap.js
```diff
@@ -1,44 +1,112 @@
...
-      <div className='absolute bottom-0 left-0 right-0'>
-        <div className='flex flex-auto flex-column items-center self-end pb5-ns no-select'>
-          <div className='f1 fw5 black'><PeersCount /></div>
-          <div className='f4 b ttu charcoal-muted'>{t('peers')}</div>
+      <div className='relative flex justify-end pt2 pb4'>
+        <div className='f6 p2 no-select flex items-center'>
+          <span className='f6 charcoal-muted pr3'>{t('index')}: </span>
+          <i className='mapDotExplanation mr1' style={{ width: getDotsSize(1) * 2, height: getDotsSize(1) * 2, backgroundColor: getDotsColor(1) }}></i>1-10 {t('peers')}
+          <i className='mapDotExplanation ml3 mr1' style={{ width: getDotsSize(50) * 2, height: getDotsSize(50) * 2, backgroundColor: getDotsColor(50) }}></i> 10-100 {t('peers')}
+          <i className='mapDotExplanation ml3 mr1' style={{ width: getDotsSize(110) * 2, height: getDotsSize(110) * 2, backgroundColor: getDotsColor(110) }}></i>100+ {t('peers')}
```
lidel, line 109: > 

i18n nit: having "Peers" multiple times may look bad in languages where the world is long etc.

I believe it will be safer to simplify this something like:

> 

This will  also enable us to add one more level if we want (eg. 1-10, 10-50, 50-100, 100+)

public/locales/en/peers.json
```diff
@@ -14,6 +15,7 @@
   "insertPeerAddress": "Insert the peer address you want to connect to.",
   "add": "Add",
   "example": "Example:",
+  "plusPeers": "Plus {number} more peers",
```
lidel, line 18: Would this look better or worse? (feel free to ignore)

```suggestion
  "plusPeers": "+ {number} more peers", 
```

src/bundles/peer-locations.js
rafaelramalho19: Intersecting circles (like it's doing in North America) leads to weird hovering glitches, that's why I kept the minimum safe distance of 4. And that can still overlap if a region has more than 100 peers 😕

rafaelramalho19:
> Thanks! Made another pass:
> 
> * Some peers have no latency info, in those cases we should either hide `(ms)` or show `(???ms)`:
>   no-latency-china-2020-04-23--22-54-22
>   
>   * There should be some whitespace between the flag and the address, ideally the same amount as in the beer list below the map.
>   * Would it be possible to reuse multiaddr formatter from the Status screen? It has grayed out protocol sections making it super easy to look at:
>     > 
> * For some time I had this weird bug where menu randomly stopped working. Finally figured it out – svg is getting over the menu, hijacking clicks in Firefox. Perhaps tweaking z-index will be enough to fix this?
>   > overlay-on-menu-2020-04-24--02-56-44
> * Smaller suggestions below

Thanks for the review @lidel!

The flag actually has empty space in MacOS, so that's why I never saw that happen 😢
I've added padding to the flag so it doesn't happen, here are examples:

MacOS flag: 
image

Windows flag:
image

I was able to fix the scrollbar issue, but I needed to get a bit creative and actually get the popover to align down instead of up. If you see this as a UX problem (CC @jessicaschilling), I'll try a more overhead solution (passing the popover as a portal element, and setting the top and left according to the body/html position, while calculating the size of the svg and subtracting offsets)


Nice suggestion on the Status component, didn't see it 👍

public/locales/en/peers.json
rafaelramalho19: Don't really have a preference, just went with @jessicaschilling suggestion at the time 😄

jessicaschilling:
I don't think the popover facing down is a big problem. If there are a ton of peers in Patagonia, it might overlap with the top of the main peer table, but that feels like a bit of an edge case for this release.

public/locales/en/peers.json
jessicaschilling: @lidel - I was waffling between "plus" and "+" in the first place, so if you'd prefer "+", let's go with it 😊

src/bundles/peer-locations.js
jessicaschilling: @lidel I might be inclined to leave as is ... @rafaelramalho19 did a lot of intermediate iterations on this that I got to see and this is kind of the best fit between circle size accuracy and avoiding weird hovering problems. Just my $.02 though 😊

lidel: Leaving 4 SGTM, especially if it was already considered :-)

lidel:
@rafaelramalho19 mind resolving the conflict? (rebase is also fine)

lidel:
@hugomrdias @alanshaw does any of you have bandwidth to take a look at React-specific parts and give @rafaelramalho19 some feedback?

src/bundles/peer-locations.js
```diff
@@ -250,26 +275,23 @@ class PeerLocationResolver {
   }
 
   optimizedPeerSet (peers) {
-    if (this.pass && this.pass < 3) {
+    if (this.pass < 3) {
       // use a copy of peers sorted by latency so we can resolve closest ones first
       // (https://github.com/ipfs-shipyard/ipfs-webui/issues/1273)
       const ms = x => (parseLatency(x.latency) || 9999)
```
hugomrdias, line 281: can we avoid doing the parse inside the sort ?

src/bundles/peer-locations.js
```diff
@@ -250,26 +275,23 @@ class PeerLocationResolver {
...
-      peers = peers.concat().sort((a, b) => ms(a) - ms(b))
+      const sortedPeersByLatency = peers.concat().sort((a, b) => ms(a) - ms(b))
```
hugomrdias, line 282: do we need to copy if we are going to use slice below ?

rafaelramalho19: For the sake of immutability, `sort` does affect the original array, so it would be best to.

Thoughts on this? 🤔

hugomrdias: if we sort it once upon creation and work with it sorted we should be safe.

rafaelramalho19: I guess you're right, I'll change it 😄

rafaelramalho19: After further analysis, I just realized that this is the poller and having this sort in this specific part, actually reduces the number of times the array gets sorted since it only does it for the 3 first times.

src/bundles/peer-locations.js
rafaelramalho19: We could, but that requires another selector dependency just to parseLatency for each peer. And you also just need to parse the latency if it's the first 3 poller requests

rafaelramalho19:
@lidel Everything's ready for merge now, just waiting on your final approval 🎉

PR #1445, opened 2020-04-18
feat: remove ipfs-desktop settings

hacdias:
This is ipfs-webui part of https://github.com/ipfs-shipyard/ipfs-desktop/pull/1425


lidel:
@hacdias I see we keep the code related to Countly: 
do you remember how that works, or can link issues where it was discussed? 

Would be useful for @rafaelramalho19 to be aware how gathering opt-in metrics works.

hacdias:
@lidel @rafaelramalho19 yes, we are keeping the Countly code so we can use the same ID for the Web UI when it is used via IPFS Desktop. Issues and PRs:

- https://github.com/ipfs-shipyard/ipfs-desktop/issues/1063
- https://github.com/ipfs-shipyard/ipfs-desktop/pull/1068
- https://github.com/ipfs-shipyard/ipfs-webui/pull/1136

The latter introduced the sharing between IPFS Desktop and IPFS Web UI where we can use Web UI to enable/disable the analytics.

However, I don't mind removing the Countly-related parts. As far as I remember, we set up this sharing to be able to have better understanding of how IPFS Desktop users were using Web UI, even though we never did that analysis.

hacdias:
Thanks for the approval. I'm merging this now then. @rafaelramalho19 if you need something or some explanation about something I did, just ping me.

PR #1466, opened 2020-04-24
[WIP] Possible approach to #1464

jessicaschilling:
A very rough attempt at resolving #1464, but I'm assuming there's a better way this could be done: this is overriding stock behavior of react-virtualized using `!important`, which isn't a great idea, and is likely breaking other display behaviors along the way.

@rafaelramalho19 -- If you've got a little time between projects, I'd appreciate learning about how you'd approach this. Thanks!


rafaelramalho19:
I'll look into it 😄

jessicaschilling:
Thanks!

PR #1481, opened 2020-04-30
chore: change add to import in files

rafaelramalho19:
## Notes

- I've ran `tx pull -a` just to update translations, so that's why you see some noise in the files changed. If you want I can rollback it :)

## Questions

- Is renaming just the label and value enough? Or should I go through the 3 codebases and rename every "addFile", "addFolder", etc.?


rafaelramalho19:
Also related with https://github.com/ipfs-shipyard/ipfs-companion/issues/817#issuecomment-621929597

Issue #1471, opened 2020-04-27
Invalid language labels in Language picker 


lidel:
For some reason `Settings` → Language → Change  has English multiple times:

> ![](https://user-images.githubusercontent.com/157609/80379455-693c6f00-889e-11ea-8e59-819777ff5dff.png)

Looks like a bug.

jessicaschilling:
If you choose the second instance of "English", it actually switches to Italian.

jessicaschilling:
Please see https://github.com/ipfs-shipyard/ipfs-webui/pull/1483.

PR #1483, opened 2020-04-30
fix: updated languages.json using lol

jessicaschilling:
Looks like language choices were displaying as "English" in the choose-language modal due to their being missing from `languages.json`.

- If it's a matter of having just forgotten to update the contents of `languages.json` after manually running `lol`, this PR addresses that.
- If `lol` is being run automatically and updating `languages.json` as some part of our deploy process, and I missed that, there's a larger problem as of yet unresolved.

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/1471

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ipfs-webui",
-  "version": "2.7.5",
+  "version": "2.8.0",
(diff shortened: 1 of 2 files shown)


rafaelramalho19:
Ups 😅

PR #1486, opened 2020-05-04
chore: change add to import

rafaelramalho19:


jessicaschilling:
LGTM, thanks!

lidel:
Hm.. this PR updates label in dialog shown after clocking "Add"→"From IPFS path".

Should we also rename the "Add" button in the top corner?

rafaelramalho19:
Yes @lidel, sorry about that. I totally forgot about updating that.

I've changed both the Modal title (previously "Add from IPFS") and the Button ID & Text

Please review again 🙏

PR #1487, opened 2020-05-04
fix: improve shell component ux

rafaelramalho19:
So, this is a nifty pick, but it was bugging me out having to remove the `$` sign every time I copied a shell command. So I made it not selectable 😄 

I also made the shell component responsive, which helped in some viewports: 

image

Here's the interaction:
Hnet-image (1)

### DIFF
--- a/src/components/shell/Shell.js
+++ b/src/components/shell/Shell.js
@@ -1,12 +1,14 @@
 import React from 'react'
+import classNames from 'classnames'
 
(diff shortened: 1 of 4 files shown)


rafaelramalho19:
@lidel Couldn't that bother the user if he only wanted to select a certain portion of the command? Maybe a better alternative is to add a copy button? 🤔

Issue #1008, opened 2019-04-10
API shell text does not apply to Windows


djdv:
**Describe the bug**
When failing to connect to the daemon, the webui shows shell prompts that are invalid on Windows.
![shell](https://user-images.githubusercontent.com/13862850/55897834-51a92000-5b8f-11e9-9e2a-ac3c1e3c050a.png)
***
`ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin '["http://127.0.0.1:8080", "http://127.0.0.1:5001", "https://webui.ipfs.io"]'`
returns these messages in: 
 - PowerShell: `Error: failed to unmarshal json. invalid character 'h' looking for beginning of value`
 - CMD: `Error: Unknown Command "API.HTTPHeaders.Access-Control-Allow-Origin"`

Valid forms of this are:
 - PowerShell: `ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin '[\"http://127.0.0.1:8080\", \"http://127.0.0.1:5001\", \"https://webui.ipfs.io\"]'
ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods '[\"PUT\", \"GET\", \"POST\"]'`
 - CMD: `ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin "["""http://127.0.0.1:8080""", """http://127.0.0.1:5001""", """https://webui.ipfs.io"""]"
ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods "["""PUT""", """GET""", """POST"""]"`

It would be nice if the shell element had some way to select what command shell you're using, and formatted the string so the arguments matched the escape type.

abdenoursouidi:
You are a hero ♥

lidel:
cc @jessicaschilling @rafaelramalho19  see the snippets created by @djdv  :-) 

(context: we discussed this very thing today!)

jessicaschilling:
Moving this to the "Quick Fixes" in Zenhub ...

@rafaelramalho19 -- is this something you're still able to take on at some point?

@lidel -- thank you!

rafaelramalho19:
Thank you so much for looking into this in the past @djdv 🙏 

Will fix it ASAP 😄

itm-platform:
Hi there,
I tried both (Windows 11)
- PowerShell: `ipfs config --json API.HTTPHeaders.Access-Control-Allow-Credentials '[\"true\"]'`
[...]

PR #1496, opened 2020-05-20
feat: add windows cmd & powershell init scripts

rafaelramalho19:
Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/1008

Screenshot of the end result:
image

### Notes

Also changed the "Start the IPFS daemon" text to match one common use case:
- User has already started daemon, but didn't setup the headers. After setting up the headers, the user is confused why it's not working. Maybe reading the "restart daemon" will help.


src/welcome/WelcomePage.js
```diff
@@ -72,14 +72,22 @@ const ConnectionStatus = ({ t, connected, sameOrigin }) => {
...
+          </Shell>
+          <Shell title="Windows CMD" className="mt4">
+            <code className='db'><b className='no-select'>$ </b>ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin "[{addOrigin && `"""${origin}""", `}"""{defaultDomains.join('""", """')}"""]"</code>
+            <code className='db'><b className='no-select'>$ </b>ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods "["""PUT""", """POST"""]"</code>
+          </Shell>
         </div>
       )}
       <Trans i18nKey='notConnected.paragraph3'>
-        <p>Start an IPFS daemon in a terminal:</p>
+        <p>Start/Restart the IPFS daemon in a terminal:</p>
```
jessicaschilling, line 90: Maybe “Then, start/restart the...” to make it clearer this is the second step (and lowercase “restart”.

jessicaschilling:
This is great 😊 Just one tiny text suggestion. Thanks!

src/welcome/WelcomePage.js
rafaelramalho19: Done :)

rafaelramalho19:
After discussing with @jessicaschilling, we came with this simple way to further help the user differentiate between which OS bash commands to use: https://i.imgur.com/nJ8JRhL.png

jessicaschilling:
Let's remove the Apple icon per: https://www.apple.com/legal/intellectual-property/guidelinesfor3rdparties.html

rafaelramalho19:
Thanks for looking into that. I reverted the icon changes then 👍

PR #1517, opened 2020-06-05
fix: bootstrapPeers can be null

D4ryl00:
In the `Peers` page, I got this error:
Screen Shot 2020-06-05 at 01 33 18

Sometimes, `bootstrapPeers` can be null so we should test it.
I have this error when I run an Go embedded Ipfs node on mobile.

Signed-off-by: remi <remi@berty.tech>


jessicaschilling:
@lidel and/or @rafaelramalho19, can you please have a look? Thanks!

rafaelramalho19:
LGTM, thanks 😄

PR #1532, opened 2020-06-26
fix: small arrow fill

rafaelramalho19:
Related with https://github.com/ipfs-shipyard/ipfs-css/issues/42

### DIFF
--- a/src/icons/GlyphSmallArrow.js
+++ b/src/icons/GlyphSmallArrow.js
@@ -2,7 +2,7 @@ import React from 'react'
 
 const GlyphSmallArrows = props => (
   <svg viewBox="0 0 32 21" {...props}>
-    <path d="M16 20.2L0.5 4.7L4.4 0.800003L16 12.4L27.6 0.800003L31.5 4.7C26.3 9.8 21.2 15 16 20.2Z" fill="currentColor"/>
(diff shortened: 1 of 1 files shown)


olizilla:
How doo! Sorry for dropping in before we've had a chance to chat. This change is good to go.

I just want to surface some history that I should have documented... The icon react components are generated from the svg icons in ipfs-css by this command: https://github.com/ipfs-shipyard/ipfs-webui/blob/be316e50353d378778a610958af6643ff835946a/package.json#L13

The theory was that we wouldn't want to manually keep the react components in sync with the svgs... the ipfs-css icon files would be the official definition.

In this case it looks like the small arrow icon compnent was added directly to ipfs-webui. I think it's totally fine that we aren't adding every single icon that we use in ipfs-webui to ipfs-css, as some things will be app specific rather than an "IPFS" icon... ipfs-css consuming folks can use whatever chevrons they like from any number of other icon sets. That said if all our apps end up needing a small down arrown, it might be worth reconisdering and pushing it up to ipfs-css.

As we may end up with generated icons and custom app specific icons in the same project, it might be wise to sperate the custom ones from the generated ones. It's not a big issue, but worth considering as it could get harder to unpick later.

jessicaschilling:
`glyph_small_arrow` and `stroke_small_arrow` are indeed in the latest ipfs-css release: see https://gateway.ipfs.io/ipfs/QmWB4HTz8G9RQeNumjFqHeHD7ziVkC3bfxWe37oEE75LZv/

PR #1535, opened 2020-07-01
Feat/pinning settings

rafaelramalho19:
This adds a mocked-up settings page component for pinning settings.

Left some todo's that I want to fix later, when we:
- Have a documentation link
- Have a working pinning API from go-ipfs to fetch current pinned services
- Have a working API to check for available pinning services

Demo: 
W1ax51C

Note: The buttons aren't doing anything on this PR, going to implement it on the next one


src/components/pinning-manager/pinning-manager-modal/PinningManagerModal.js
```diff
@@ -0,0 +1,54 @@
...
+import Button from '../../button/Button'
+
+const LanguageModal = ({ t, tReady, onLeave, link, className, ...props }) => {
+  const handleClick = (lang) => {
+    i18n.changeLanguage(lang)
+    onLeave()
+  }
+
+  return (
+    <Modal {...props} className={className} onCancel={onLeave} style={{ maxWidth: '40em' }}>
```
rafaelramalho19, line 18: This modal is just a placeholder (copied from language modal), will also update it next :)

src/settings/SettingsPage.js
```diff
@@ -31,6 +32,19 @@ export const SettingsPage = ({
       <title>{t('title')} - IPFS</title>
     </Helmet>
 
+    <Box className='mb3 pa4'>
+      <Title>{t('pinningServices.title')}</Title>
+      <Trans i18nKey='pinningServices.description'>
+        <p className='ma0 mr2 lh-copy charcoal f6'>
+          <span>Use local pinning to ensure files on your local node persist and are never garbage-collected.
+          You can also link your accounts with other pinning services to automatically or selectively persist files with those providers, enabling you to keep backup copies of your files and/or make them available to others when your local node is offline. </span>
```
jessicaschilling, line 40:
```suggestion
          You can also link your accounts with other remote pinning services to automatically or selectively persist files with those providers, enabling you to keep backup copies of your files and/or make them available to others when your local node is offline. </span>
```
My bad on this one, the language has evolved somewhat 😊

rafaelramalho19: Done 😄

rafaelramalho19:
image
image

rafaelramalho19:
> Small nits:
> 
> * "Local pinning" does not need "three dot menu" – there is nothing to edit, visit, and AFAIK we can't remove it
> * As noted in my review of #1539, images should be loaded from IPFS.
>   (while I was testing me one of [third party hostnames was down and image failed to load](https://user-images.githubusercontent.com/157609/86485434-fac1d480-bd58-11ea-8584-2e9c54a02793.png))
> 
> @rafaelramalho19 do we really want to start merging these changes `master`?
> 
> * If we need to make a bugfix release before pinning services are a thing, we have a problem and need to manually hide pinning services UI before release.
>   
>   * If you are ok with the risk of that additional work, ok with me.
> * Translations of new strings will get published to Transifex, but as we iterate on this, those may change, and we add unnecessary work for community members in IPFS translation team.
>   
>   * If we are extra diligent around proof-reading new labels and ensure they are pretty much solid when merging to `master`, it should not be a problem, but wanted to mention the concern here as I know how discouraging it is when you translate something and the same paragraph  is changed multiple times withing one or two weeks.
> 
> If you'd like to play it safe, perhaps create `epic/pinning-services` branch and create new PRs against that instead? That way pinning services epic work would be isolated from regular dev and when we want to ship pinning services, we would merge everything to master in a single PR.

Will fix the small issues you pointed out, thanks 👍 

I was thinking about creating a branch called `release/pinning-services`, but `epic/pinning-services` sounds good to me 😄

jessicaschilling:
@lidel Thanks for catching that three-dots menu for local services, feeling a bit silly for not seeing that.

PR #1539, opened 2020-07-02
Feat/pinning settings modal

rafaelramalho19:
Result: image

## TODOS:
- Add link/action to add custom integration
- Handle clicks on modal items and proceed with the "add service" journey

### DIFF
--- a/src/bundles/pinning.js
+++ b/src/bundles/pinning.js
@@ -32,13 +32,13 @@ export default {
   selectAvailablePinningServices: () => ([
     {
       name: 'Pinata',
-      icon: 'https://svgshare.com/i/M7y.svg'
+      icon: 'https://ipfs.io/ipfs/QmVYXV4urQNDzZpddW4zZ9PGvcAbF38BnKWSgch3aNeViW?filename=pinata.svg'
     }, {
       name: 'Infura',
-      icon: 'https://i.pinimg.com/originals/e2/8e/98/e28e98282e6ace744098d2d25ddee7d2.png'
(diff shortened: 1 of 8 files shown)


src/bundles/pinning.js
```diff
@@ -27,5 +27,18 @@ export default {
...
+    {
+      name: 'Pinata',
+      icon: 'https://svgshare.com/i/M7y.svg'
+    }, {
+      name: 'Infura',
+      icon: 'https://i.pinimg.com/originals/e2/8e/98/e28e98282e6ace744098d2d25ddee7d2.png'
+    }, {
+      name: 'Eternum',
+      icon: 'https://www.eternum.io/static/images/icons/favicon-32x32.a2341c8ec160.png'
+    }
```
lidel, line 42: Nit: WebUI should not make requests to remote hostnames, we should include these assets in local build so they are loaded from IPFS.

rafaelramalho19: These icons will be provided by the API that manages available pinning services in the future, but you think it's worth adding them now to IPFS?

lidel: Yes, please add them to IPFS. When showing those screens we want to avoid failures like this:

> Screenshot_2020-07-03 Settings - IPFS

As for image URL, are you suggesting we persist `ApiIcon` under `Pinning.Services` config map?
If so, we should rais that need and discuss that in https://github.com/ipfs/go-ipfs-config/pull/104  (I'd use CID instead of URL).

rafaelramalho19: Done 👍

PR #1552, opened 2020-07-22
chore: update dependencies

rafaelramalho19:
Tested everything in webui and seems to be working fine. But I might of missed some edge-cases so please help 🙏


rafaelramalho19:
> The dependencies update looks good. I'm having some issues but I can't guarantee it's the dependencies fault. I'm running the latest IPFS Desktop version. The issues:
> 
> * When adding a file through "Import" on the root MFS, it works.
> * When adding a file through "Import" on any subdirectory, `API returned a partial response`.
> * When adding a directory through "Import", same error as above.
> * Drag&drop works everywhere.
> * Everything I click gets an outline.
> * When I rename a file and I'm on the same page as the file, it doesn't redirect and opens the "Imported x files" dialog.

Is that replicable in the default branch?

rafaelramalho19:
Thank you for testing that @hacdias, maybe we should open issues with the problems you found? 🤔

Issue #1529, opened 2020-06-25
Improve file add so it can handle gigs of data


Gozala:
This is related to https://github.com/ipfs/js-ipfs/issues/3029, however changes on the webui end would also be required. So this is the placeholder issue for that piece of work.

Gozala:
I don't seem to have right privileges to assign it to myself. But I do intend to drive this.

jessicaschilling:
@Gozala I assigned you! Do you mind adding labels as appropriate?

Gozala:
Some notes (mostly for myself). Here's where dropped files are handled

https://github.com/ipfs-shipyard/ipfs-webui/blob/be316e50353d378778a610958af6643ff835946a/src/App.js#L109-L116

Which invokes `addFiles`

https://github.com/ipfs-shipyard/ipfs-webui/blob/be316e50353d378778a610958af6643ff835946a/src/App.js#L42-L53

Where array of web [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File)s is converted to array of entries with `content` feilds are turned into streams of file contents. (This where we fall of optimal path and will end up having to buffer). 

https://github.com/ipfs-shipyard/ipfs-webui/blob/be316e50353d378778a610958af6643ff835946a/src/lib/files.js#L4-L18

And then `doFilesWrite` uses `ipfs.add` to get those into IPFS via `ipfs-http-client`

https://github.com/ipfs-shipyard/ipfs-webui/blob/be316e50353d378778a610958af6643ff835946a/src/bundles/files/actions.js#L170-L207

Gozala:
> @Gozala I assigned you! Do you mind adding labels as appropriate?

@jessicaschilling I'm afraid I don't have access right to do that either. Also I'd need some guidance on appropriate labeling.

PR #1569, opened 2020-07-29
Migration to the new IPFS JS API with Async Await and Async Iterables

### LINKED ISSUE #1431: Use ipfs-http-client >42 (refactor with Async Await and Async Iterables)
js-ipfs 41 introduced breaking changes to the programmatic JS API:
https://blog.ipfs.io/2020-02-13-js-ipfs-0-41/

The same programmatic API is exposed by the latest `ipfs-http-client` which is a dependency of `ipfs-redux-bundle` used by webui for connecting to the HTTP API of IPFS node.

While switching to the latest API, remove patch/hotfix introduced in #1430

Gozala:
# Overview

We need to upgrade to new IPFS so we can take advantage of a fix to https://github.com/ipfs/js-ipfs/issues/3029 to unblock https://github.com/ipfs-shipyard/ipfs-webui/issues/1529. However upgrade end up been complicated by introduces async iterable https://github.com/ipfs-shipyard/ipfs-webui/issues/1431. And all the involved dependencies that had various conflicts that required updating to get through. Below is high level overview of changes this pull packs:

- Dependency updates
  - Updates ipfs / ipfs-http-client to latest
     - Updates all uses to take into account async iterables.
     - Updates all uses to take into account `ipfs.add` / `ipfs.addAll` API split.
  - Updates `ipfsd-ctl` to lastest (so that latest js-ipfs / go-ipfs could be passed in)
  - Switches from `go-ipfs-dep` to `go-ipfs` (because that is what we do elsewhere and that is what ipfsd-ctl seems to suggest. As far as I understood from talking to @lidel former was used for legacy reasons).
  - Updated bunch of other things because per errors and suggestions from react-scripts
- `filesToStreams` got replaced by `normalizeFiles` function
   - We only pass File objects into it and improvements to `js-ipfs` / `ipfs-http-client` have being improved to no longer require streams.
   - As per #1529 we should not use streams to avoid buffering file content into memory
- Update bunch of code to use `.cid` in place of former `.hash` due to changes in IPFS API.
- Add bunch of TS via JSDoc comments
   - Seems unrelated, however it was instrumental in figuring out all the places that need to be change.
     - In fact I end up spending a lot more time updating tests, because I could not take advantage of type checker to spot all the API incompatibilities in mocks and all.
   - As per discussion on higher bandwidth channel, I've added those annotations in code paths that I had worked on and added `tsconfig.json` / `npm run check` so that files that are annotated get checked.

 

Fixes #1431


src/bundles/files/actions.js
```diff
@@ -22,57 +62,88 @@ const realMfsPath = (path) => {
...
     // No information is lost: if there is an error related
     // to specified hashOrPath user will read it in Inspector.
+    const [, , cid] = path.split('/')
     return {
       path: hashOrPath,
-      hash: hashOrPath,
-      type: 'unknown'
+      cid: new CID(cid),
+      type: 'unknown',
+      size: -1
```
Gozala, line 95: This causes following problem https://github.com/ipfs-shipyard/ipfs-webui/pull/1569/#pullrequestreview-462879642

Gozala: maybe not

package.json
```diff
@@ -21,15 +21,16 @@
     "bundlesize": "bundlesize",
     "eject": "react-scripts eject",
     "storybook": "start-storybook -p 9009 -s public",
-    "build-storybook": "build-storybook -s public"
+    "build-storybook": "build-storybook -s public",
+    "check": "tsc --noEmit"
```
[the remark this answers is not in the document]
Gozala: I have added an action, but honestly current coverage is too small to catch much. It does however makes it easy to understand the code you're working with e.g.

image
image
image
image

Gozala:
> There is a regression on Files screen in functionalities not covered by E2E tests – mind taking a look?

Thanks for catching all these!

> 1. File size column and pin status are wonky for directories and empty files.
>    I suspect API calls used for reading file sizes and pin status changed:

File size was my issue, I have incorrectly entyped `type` field as `'file' | 'directory'` which turns out to be `'file' | 'dir'`.
Fixed by https://github.com/ipfs-shipyard/ipfs-webui/pull/1569/commits/229a66c62c6f63b46ee6e0deb42e65695529d1bf

Pinning status was fixed by https://github.com/ipfs-shipyard/ipfs-webui/pull/1569/commits/fc6d5017874fee54d96f23b8ee6eff0ac84148ca. Overlooked that pin status was checked via `array.includes` which worked for hashes as strings but doesn't for CIDs that are objects.

> Pinning via context menu does not work (yellow error in UI + one in browser console), most likely for the same reason.

Fixed by https://github.com/ipfs-shipyard/ipfs-webui/pull/1569/commits/b1b908c68dca17c6e71f5f0e87fe6220b7ea3b5e. I have no slightest clue what that code supposed to do however. I believe it attempt to flatten array there so I changed check to only do so on arrays, but would be good if someone more knowledgeable could fill out that TODO: to explain what's up.

> ps. I also found a bug around Download action – it was already in master so not related to this PR – filled #1576

Strange. I did tested all the context menu actions, including download (on this branch) and it seems to work and download here. It could be that array flattening code was causing problems for downloads too.

src/bundles/files/utils.js
```diff
@@ -19,7 +26,8 @@ export const make = (basename, action, options = {}) => (...args) => async (args
   try {
     data = await action(getIpfs(), ...args, id, args2)
 
-    const paths = args[0] ? args[0].flat() : []
+    // TODO: Add a comment explaining what is going on here.
```
rafaelramalho19, line 29: Don't forget this TODO 😄

src/files/header/Header.js
```diff
@@ -36,7 +36,8 @@ class Header extends React.Component {
     const pos = this.dotsWrapper.getBoundingClientRect()
     this.props.handleContextMenu(ev, 'TOP', {
       ...this.props.files,
-      pinned: this.props.pins.includes(this.props.files.hash)
+      pinned: this.props.pins.includes(this.props.files.cid.toString()),
+      _: console.log(this.props)
```
rafaelramalho19, line 40: You forgot to delete this?

src/bundles/files/utils.js
Gozala: I was actually hoping you or someone with more knowledge of the code base could do that, because I really have no clue what is going on here.

src/files/header/Header.js
Gozala: Thanks for spotting this!

src/bundles/files/utils.js
Gozala: Submitted a followup https://github.com/ipfs-shipyard/ipfs-webui/issues/1578

PR #1570, opened 2020-07-30
feat: add cube animation as loading indicator

rafaelramalho19:
## Note
- If you want to test this, you should probably throttle your connection to slow or fast 3G 😄 

Screen-Recording-2020-07-30-at-1

- I had to convert this animation to CSS only, so it can work while the JS is being parsed/downloaded.


src/loader/ComponentLoader.stories.js
```diff
@@ -1,18 +1,12 @@
...
-    <div className='sans-serif pa4' style={{ height: 400 }}>
-      <ComponentLoader
-        pastDelay={boolean('pastDelay', true)}
-      />
+storiesOf('ComponentLoader', module)
+  .add('Default', () => (
+    <div className='flex items-center'>
+      <ComponentLoader pastDelay style={{ width: 100, margin: '100px auto' }}/>
+      <ComponentLoader pastDelay style={{ width: 200, margin: '100px auto' }}/>
+      <ComponentLoader pastDelay style={{ width: 300, margin: '100px auto' }}/>
```
lidel, line 10: Why we have three of those?

jessicaschilling: I think this was just to get a sense of the cube's visual impact at a few different sizes, as it may appear at different scale in different settings. Could remove, but it was useful to see in our implementation discussion.

lidel: ah, I was before my coffe and did not realize this is storybook :sweat:  nvm me!

Issue #1421, opened 2020-03-30
Add CLI "tutor mode" to Web UI/Desktop


lanzafame:
## This issue has a bounty!
Successfully closing this issue by making a PR of production-ready code can earn you not only the undying love of the IPFS community — it can net you a financial reward. [See the current list of bounty issues and their values here.](https://github.com/ipfs/devgrants/projects/1)

## How to claim this bounty
Want to work on this bounty project? Please leave a comment on this issue, and we'll be back in touch ASAP to discuss specifics. **Note that because this project requires changes to the existing UI, we'll want to work closely on making sure the solution integrates well with current patterns.**

## The need in brief
Meet the needs expressed in the original proposal below by adding some form of "CLI tutor mode" to IPFS Web UI (which, in turn, will be included in IPFS Desktop, since they share this portion of the codebase). The solution should ...

- Enable the user to generate copy/paste-able CLI snippets for the actions available in Web UI, with an aim to making it easier for newer IPFS users to self-learn the CLI
- Be readily available for such users, but not so unavoidable that users who don't want to take advantage of the feature can't ignore it

### Original proposal from @lanzafame 
I was just creating a VM on GCP and they have a button at the end that takes a UI hand-craft VM creation and turns it into either a curl request or a gcloud cli command and thought it would be cool to have something similar for the webui so if people wanted to know how to pin something like they are in the webui on the cli, they could get the command and the arguments that they had already input. Sort of a self-teaching moment for the curious
<img width="842" alt="Screen Shot 2020-03-27 at 13 37 02" src="https://user-images.githubusercontent.com/5924712/77866126-1490de80-7275-11ea-85f0-2181f47825dd.png">

jessicaschilling:
@lanzafame This is such a cool idea 😊 ... and also sort of a big chunk to bite off. Do you think it's something that would be appropriate to put out for a bounty? Otherwise I fear it might languish just given current resourcing. Any other thoughts from your end?

lanzafame:
@jessicaschilling I think that is a great idea.

jay-dee7:
hey @lanzafame, can i please start working on this over the weekend?

lanzafame:
@jay-dee7 sure thing

jessicaschilling:
[...]

PR #1572, opened 2020-08-01
Feature - CLI Tutor Mode

jay-dee7:
Adds the **CLI Tutor Mode**, which can help with learning the IPFS Command Line Interface from the Webui/Desktop. This feature was requested in #1421 

#### Done
* Option to enable/disable CLI Tutor Mode from the settings page
* If enabled, a `copy-icon` 📋 is shown next to the buttons/menus where those actions can be performed via IPFS' CLI
* This icon is visible at:
  * Files Page -> inside `+ Import` menu (Add File, Folder, Copy from IFPS)
  * Files Page -> File Menu (three dots next to a file) -> (Delete, Rename, Pin/Unpin, Download)
  * Peers Page -> next to `+ Add Connection`
  * Settings Page -> next to `Reset` button in config edit box
  * ~~Setting Page -> next to `Submit` button in API edit box~~

### Pending Work:
* ~~I recently discovered a state management issue with the modals, I couldn't find the root cause so far, it started happening after the refactor today (1st Aug 2020)~~
* ~~After rebasing from master today, there's a small UI glitch (text in the modal actions is left aligned, I'll work it out before making this PR reviewable)~~
* ~~I have added a bunch of TODO's in the code, which can potentially help with the code cleanup~~
* ~~Files component kind of looks like a mess right now, a lot of state is being shared across mulitple components (maybe we can refactor it to make it look neat)~~

Here's a visual for how it looks in action: [Jessica note: these are a bit outdated now since commits for visual tweaks on 12 Aug]

### Settings Page | Option to enable CLI Tutor Mode
enable-ipfs-cli-tutor-mode

### Files Page | CLI Tutor Mode Enabled
cli-tutor-mode-files-page

### Settings Page | CLI Tutor Mode - Open Modal
cli-tutor-mode-open-modal

Signed-off-by: Jasdeep Singh <jasdeepsingh.uppal@gmail.com>


jay-dee7:
@jessicaschilling here's the draft for cli tutor mode, the place which would need more attention is files component, since it has a lot of shared state, please check this out and suggest any refactors/changes for places where i might have broken any acceptable coding patterns. I'll update this PR tomorrow with little of code cleanup and modal actions alignment fix

jessicaschilling:
Thanks, @jay-dee7! I'd like @rafaelramalho19 to look through this in more detail ... but just a heads up that he's on holiday for the next week. Don't want you to think that we are ignoring you. 😊

jay-dee7:
That's okay @jessicaschilling. In-fact apologies for the delay. This should've been up way before than it did. In the meantime I'll keep  on iterating  over this, just to see if anything can be improved upon 😄

src/components/cli-tutor-mode/CliTutorMode.js
```diff
@@ -0,0 +1,109 @@
...
+
+const CliTutorialModal = ({ command, t, onLeave, className, downloadConfig, ...props }) => {
+  const onClickCopyToClipboard = async (command) => {
+    await navigator.clipboard.writeText(command)
+  }
+
+  return (
+    <Modal {...props} className={className} onCancel={onLeave} style={{ maxWidth: '40em' }}>
+      <ModalBody icon={CopyIcon}>
+        <p className='charcoal w-80 center'>{t('copy the command to use it with ipfs command line interface')}</p>
```
jessicaschilling, line 21:
```suggestion
        <p className='charcoal w-80 center'>{t('Copy the following and paste it into your terminal application to do this task in IPFS via the command line.')}</p>
```
Just a text edit.

src/components/cli-tutor-mode/CliTutorMode.js
```diff
@@ -0,0 +1,109 @@
...
+        <div>
+          <Shell title="terminal shell">
```
jessicaschilling, line 23:
```suggestion
          <Shell title="Shell">
```
Just "Shell" is fine here. Use the capital "S" in case we decide not to use CSS to uppercase these lines in future.

jay-dee7:
sure @jessicaschilling i'll make these changes ASAP. There are more things that I didn't notice initially. ContextMenu in files pages needs some work, after rebasing, the copy icon and the text aren't aligned symmetrically. I'll try to get them fixed before the review starts

jay-dee7:
@jessicaschilling I rebased from master just a while ago, one of the incoming changes were in `ApiAddressForm` component and since it was written using hooks (I believe), I was not able to overwrite props easily (I'm not sure what exactly broke the prop sharing, i believe its hooks but i have no idea.). So just for the `ApiAddressForm` component, I've managed the state for `cli-tutor-modal` locally to the component, instead of using the regular redux actions. Rest of the stuff stays almost same. I think this PR is ready for review 🤞

jessicaschilling:
Thanks, @jay-dee7! 
@rafaelramalho19 - are you able to take a look at this PR once you are back in and settled? Thank you!

src/components/cli-tutor-mode/CliTutorMode.js
```diff
@@ -0,0 +1,110 @@
...
+          <Button className='ma2 tc' onClick={() => onClickCopyToClipboard(command)}>
+            {t('actions.copyCommand')}
+          </Button>
+        </div>
+      </ModalActions>
+    </Modal>
+  )
+}
+
+class CliTutorMode extends Component {
```
rafaelramalho19, line 56: You're using a functional component above and here you used a class component. Can you make them equal? (I would advise choosing functional)

src/components/cli-tutor-mode/CliTutorMode.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  return (
+    <Modal {...props} className={className} onCancel={onLeave} style={{ maxWidth: '40em' }}>
+      <ModalBody icon={CopyIcon}>
+        <p className='charcoal w-80 center'>
+          {t('cli.description')}
+        </p>
+        <p className='charcoal-muted w-90 center'>
+          {command.includes('ipfs config replace') ? t('cli.extraNotes') : ''}
```
rafaelramalho19, line 25: Please change this to `command === cliCmdKeys.UPDATE_IPFS_CONFIG` (you need to import cliCmdKeys of course).

We should have a single source of truth for the command, just to avoid changing it on the constant files and breaking on this file.

src/components/cli-tutor-mode/CliTutorMode.js
```diff
@@ -0,0 +1,110 @@
...
+        </div>
+      </ModalBody>
+
+      <ModalActions>
+        <div>
+          <Button className='ma2 tc' bg='bg-gray' onClick={onLeave}>{t('actions.close')}</Button>
+        </div>
+        <div className='flex items-center'>
+          {
+            command && command.includes('ipfs config replace')
```
rafaelramalho19, line 40: Replace here as suggested previously.

Also, why do you check for the existence of `command` here but not on line 25?

src/components/cli-tutor-mode/CliTutorMode.js
```diff
@@ -0,0 +1,110 @@
...
+      return (
+        <Fragment>
+          {
+            showIcon ? <CopyIcon
+              onClick={doOpenCliTutorModal}
+              className='dib fill-current-color ph2 glow o-80 pointer'
+              style={{ height: '28px', transform: 'scale(1.5)', verticalAlign: 'bottom', color: 'dodgerblue' }}
+            /> : null
+          }
+          <Overlay show={isCliTutorModalOpen} onLeave={doOpenCliTutorModal}>
```
rafaelramalho19, line 85: Just to be consistent shouldn't you do `() => doOpenCliTutorModal(false)`? I think that's the issue you had with being an object sometimes (I think that onLeave sends the click event as an argument)

src/components/cli-tutor-mode/CliTutorMode.js
```diff
@@ -0,0 +1,110 @@
...
+        </Fragment>
+      )
+    }
+  }
+}
+
+CliTutorialModal.propTypes = {
+  onLeave: PropTypes.func.isRequired,
+  t: PropTypes.func.isRequired,
+  tReady: PropTypes.bool
```
rafaelramalho19, line 98: You're not using tReady, why declare it?

src/files/FilesPage.js
```diff
@@ -233,14 +236,19 @@ class FilesPage extends React.Component {
           onInspect={() => this.onInspect(contextMenu.file.hash)}
           onDownload={() => this.onDownload([contextMenu.file])}
           onPin={() => this.props.doFilesPin(contextMenu.file.hash)}
-          onUnpin={() => this.props.doFilesUnpin(contextMenu.file.hash)} />
+          onUnpin={() => this.props.doFilesUnpin(contextMenu.file.hash)}
+          isCliTutorModeEnabled={isCliTutorModeEnabled} // @TODO connect ContextMenu to redux
```
rafaelramalho19, line 240: Todos are actually done with `//TODO: `, and isn't it connected already?

src/files/dropdown/Dropdown.js
```diff
@@ -1,10 +1,22 @@
...
-  </button>
+export const Option = ({ children, onClick, className = '', isCliTutorModeEnabled, onCliTutorMode, ...props }) => (
+  isCliTutorModeEnabled
+    ? <div className='flex items-center justify-between'>
+      <button role='menuitem' className={`bg-animate hover-bg-near-white pa2 pointer flex items-center flex-grow-1 ${className}`} onClick={onClick} {...props}>
+        {children}
+      </button>
+      <button {...props} className={`bg-animate hover-bg-near-white pa2 pointer flex items-center  ${className}`}>
+        <CopyIcon {...props} onClick={onCliTutorMode} className='dib fill-current-color ph2 glow o-80 pointer'
+          style={{ height: '28px', transform: 'scale(1.3)', color: 'dodgerblue' }}
```
rafaelramalho19, line 13: Try to avoid the transform (same as commented before)

src/settings/SettingsPage.js
```diff
@@ -56,6 +60,26 @@ export const SettingsPage = ({
 
     <Experiments t={t} />
 
+    <Box className='mb3 pa4'>
+      <div className='mb4'>
+        <Title>{t('CLI TUTOR MODE')}</Title>
+        <Checkbox className='dib' onChange={doToggleCliTutorMode} checked={isCliTutorModeEnabled}
+          label={<span className='fw5 f6'>{t('Enable command-line interface (CLI) tutor mode')}</span>}/>
```
rafaelramalho19, line 67: You need to add a translation key here, not the text

src/settings/SettingsPage.js
```diff
@@ -56,6 +60,26 @@ export const SettingsPage = ({
...
+        <div className='f6 charcoal lh-copy mw7'>
+          {
+            t('Enable this option to display')
```
rafaelramalho19, line 70: Same here.

src/settings/SettingsPage.js
```diff
@@ -56,6 +60,26 @@ export const SettingsPage = ({
...
+          }
+          <CopyIcon
+            className='dib fill-current-color ph2 glow o-80'
+            style={{ height: '28px', transform: 'scale(1.3)', verticalAlign: 'bottom', color: 'dodgerblue' }}/>
+          {
+            t(`next to common IPFS commands. Clicking an icon copies
```
rafaelramalho19, line 76: And here

src/settings/SettingsPage.js
lidel: Namely, add the key to `public/locales/en/settings.json` 
(no need to add it to other locales, that will happen automatically after PR is merged)

src/bundles/cli-tutor-mode.js
```diff
@@ -0,0 +1,54 @@
+export default {
+  name: 'cliTutorMode',
+  actionBaseType: 'CLI_TUTOR_MODE_TOGGLE',
+  init: store => {
```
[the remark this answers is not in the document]
jay-dee7: @rafaelramalho19 this was bad on my part, this piece of code isn't required at all, I cleaned it up in my latest commit

src/bundles/cli-tutor-mode.js
```diff
@@ -0,0 +1,54 @@
...
+    store.cliTutorMode = true
+  },
+  persist: true,
+  checkIfOnline: false,
+  doToggleCliTutorMode: key => ({ dispatch }) => {
+    const isEnabled = JSON.parse(localStorage.getItem('isCliTutorModeEnabled'))
```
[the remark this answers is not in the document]
jay-dee7: @rafaelramalho19 I did add the suggested pattern however this was the first time I used redux-bundler. Please go through it once more just to validate if I did what was requested. I hope i didn't drift away too far :)

src/components/api-address-form/ApiAddressForm.js
```diff
@@ -18,6 +23,22 @@ const ApiAddressForm = ({ t, doUpdateIpfsApiAddress, ipfsApiAddress = '' }) => {
...
     }
   }
+
+  const showCliTutorIcon = () => {
+    if (isCliTutorModeEnabled) {
+      return (<Fragment>
+        <CopyIcon
+          onClick={setCliTutorModal}
+          className='dib fill-current-color ph2 glow o-80 pointer'
+          style={{ height: '28px', transform: 'scale(1.5)', verticalAlign: 'bottom', color: 'dodgerblue' }}
```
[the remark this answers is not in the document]
jay-dee7: for this, i have now added two options, one using just the bigger pixel size (height is increased) and another one using `msTransform` and `webkitTransform`. I can quickly stick to one solution but as far as i can tell, they both work as expected. `webkitTransform` worked okay for me on safari too

src/components/cli-tutor-mode/CliTutorMode.js
jay-dee7: I've Converted this into a functional component :)

src/components/cli-tutor-mode/CliTutorMode.js
jay-dee7: Thanks for digging this one, I've added comparison with the full command from `cliCmdKeys` now

src/components/cli-tutor-mode/CliTutorMode.js
jay-dee7: added the same comparison here too

src/components/cli-tutor-mode/CliTutorMode.js
jay-dee7: I think thats exactly what was causing the issue. I replaced the call with `() => doOpenCliTutorModal(false/true)` and it has been working just fine

src/components/cli-tutor-mode/CliTutorMode.js
jay-dee7: Removed this and added `command`, since command is a required field

src/files/FilesPage.js
jay-dee7: contextMenu page actually isn't connected but also it doesn't need to. Maybe if we'd want, we do can connect `Dropdown` component to reduce prop chaining

src/files/dropdown/Dropdown.js
jay-dee7: added a css class for this with `webkitTransform` but it would also work if we simple increased the height by 1.3X. Lemme know which way we'd like to move forward with :)

src/settings/SettingsPage.js
jay-dee7: @rafaelramalho19 @lidel i've replace these text lines with keys from translation files :)

src/settings/SettingsPage.js
jay-dee7: replaced it

src/settings/SettingsPage.js
jay-dee7: and replaced this one too

src/components/api-address-form/ApiAddressForm.js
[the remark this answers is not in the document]
rafaelramalho19: It works, the problem is that it renders a pixelated icon in safari (old issue unfortunately), so I would just stick with increasing the height

src/bundles/cli-tutor-mode.js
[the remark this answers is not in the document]
rafaelramalho19: You did good 👍

rafaelramalho19:
After you remove the transforms, LGTM 😄

jay-dee7:
thanks for the amazing review @rafaelramalho19, i've change the css class to use just height and no transforms :)

jessicaschilling:
@lidel -- do you mind going over this when you're able? I think we're very close to this being done.

NOTE: I want to get the "copy to clipboard" icon into ipfs-css before we ship. I can do that while you're reviewing 😊

Thanks!

jay-dee7:
hey @jessicaschilling I just saw the changes that you made and I think they really addch the finishing touches to the feature, left aligned text in the modal looks great plus the new icon actually looks pretty amazing :) However, I did notice once thing though, the Checkbox on CLI Tutor Mode is slightly moved to bottom `lh-copy` class is doing the most of it. Can we do some sort of workaround for that?

I was going through the ipfs documentation and realised that a couple of commands are not correct I think, namely
* `Files Page -> + Import Button -> New Folder`
* `Files Page -> individual file -> three dots -> Rename`
I'll push the change for both of these commands just now but please let me know if rest of them are correct too

image

src/bundles/files/consts.js
```diff
@@ -40,3 +40,31 @@ export const DEFAULT_STATE = {
...
+
+export const cliCommandList = {
+  [cliCmdKeys.UPDATE_IPFS_CONFIG]: () => 'ipfs config replace <path-to-settings.json>',
+  [cliCmdKeys.DELETE_FILE_FROM_IPFS]: (cid) => `ipfs files rm ${cid}`,
+  [cliCmdKeys.DOWNLOAD_OBJECT_COMMAND]: (cid) => `ipfs get ${cid}`,
+  [cliCmdKeys.PIN_OBJECT]: (cid, op) => `ipfs pin ${op} ${cid}`,
+  [cliCmdKeys.RENAME_IPFS_OBJECT]: (oldFileName, newFileName) => `ipfs files mv "${oldFileName}" ${newFileName}`,
+  [cliCmdKeys.ADD_FILE]: () => 'ipfs add <file-name>',
+  [cliCmdKeys.ADD_DIRECTORY]: () => 'ipfs add -r <folder-name>',
+  [cliCmdKeys.CREATE_NEW_DIRECTORY]: () => 'ipfs add -r <folder-name>',
```
lidel, line 66:
```suggestion
  [cliCmdKeys.CREATE_NEW_DIRECTORY]: () => 'ipfs files mkdir <folder-name>',
```

jessicaschilling:
@jay-dee7 @lidel I'd vote for closing the dialog onclick of the "Copy" button. @jay-dee7, can you please change that?

jay-dee7:
Sure @jessicaschilling I'll make the change ASAP :)

jessicaschilling:
@jay-dee7 And thanks for spotting the misaligned checkbox! Fixed it. 😊

jay-dee7:
@jessicaschilling @lidel i've added both the requested points. 
Now by clicking the copy button in the modal, we copy the contents (command) and then close the modal.

Also, I've added a fix for rename command. I realized that we need to preserve the file path while renaming and we should wrap the full path path with quotes ("<path>"), because it'll throw errors if the full path contains spaces (for example a file named `cat image.png` would throw error in cli but if we wrap it with quotes as `"cat image.png"`) it won't. Please let me know if we're good to go with this :)

jessicaschilling:
@lidel - I'm happy but will wait for your OK, please feel free to merge 😊

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "checkJs": true,
     "forceConsistentCasingInFileNames": true,
     "noImplicitReturns": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
lidel, line 7: @Gozala how strongly do we  feel about this?
@jay-dee7 is it possible to add type info and keep this as `true`?
```suggestion
    "noImplicitAny": true,
```

src/files/modals/Modals.js
```diff
@@ -123,15 +129,46 @@ class Modals extends React.Component {
...
 
+    switch (action) {
+      case cliCmdKeys.DELETE_FILE_FROM_IPFS:
+        return cliCommandList[action](activeCid)
+      case cliCmdKeys.DOWNLOAD_OBJECT_COMMAND:
+        return cliCommandList[action](activeCid)
+      case cliCmdKeys.RENAME_IPFS_OBJECT:
+        return cliCommandList[action](path, fileName)
+      case cliCmdKeys.PIN_OBJECT:
+        return cliCommandList[action](activeCid, isPinned ? 'rm' : 'add')
```
lidel, line 163: I believe WebUI pins recursively:
```suggestion
        return cliCommandList[action](activeCid, isPinned ? 'rm' : 'add -r')
```

lidel (changes requested):
@jay-dee7 nearly there! small asks inline :pray:

src/files/modals/Modals.js
jay-dee7: @lidel by recursive you mean that if we pin a directory, all the files in the directory and any of it's sub-directories and their files gets pinned?

src/bundles/files/consts.js
```diff
@@ -40,3 +40,34 @@ export const DEFAULT_STATE = {
...
+  ADD_DIRECTORY: 'addNewDirectory',
+  CREATE_NEW_DIRECTORY: 'createNewDirectory',
+  FROM_IPFS: 'fromIpfs',
+  ADD_NEW_PEER: 'addNewPeer',
+  UPDATE_API_SERVER_ADDRESS: 'updateApiServerAddress'
+}
+
+export const cliCommandList = {
+  [cliCmdKeys.UPDATE_IPFS_CONFIG]: () => 'ipfs config replace <path-to-settings.json>',
+  [cliCmdKeys.DELETE_FILE_FROM_IPFS]: (cid) => `ipfs files rm ${cid}`,
```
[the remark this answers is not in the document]
jay-dee7: @lidel thanks for catching this one. I'll change it to use full filepath and also recursive. I think 
```ipfs files rm -r ${filepath}```
is a much safer option (command would succeed in more cases regardless of a directory/file)

src/files/modals/Modals.js
jay-dee7: i just confirmed, webui doesn;t add/remove pin recursively
ipfs-add-pin
 but please let me know which approach should i follow :)

tsconfig.json
jay-dee7: I just added type information for the methods where typescript transpiler was failing and reverted this option back to `true`. It works now as expected :)

lidel (approved):
Thank you @jessicaschilling, I think the old i18next issue is gone, all looks good.

----

I've made a small fix in 5641184 to ensure CLI tutor lists import actions as the button in GUI 
(`ipfs add && ipfs files cp`, context: https://github.com/ipfs-shipyard/ipfs-webui/issues/676)

Ideally `/<dest-name>` in import commands would be replaced with `"${path}/<dest-name>"` where `path` is the current dir in MFS. I started doing that in 3a337de but had no bandwidth to wire it up.

I think it's ok to merge with static destination, but if anyone has time to ensure `path` is set for [these actions](https://github.com/jay-dee7/ipfs-webui/blob/jay-dee7/cli-tutor-mode/src/files/modals/Modals.js#L157-L160) that would be even better for user experience.

jessicaschilling:
@lidel -- Thanks for calling the subdirectory detail out. I'm inclined to merge this as-is and address that in a separate issue (https://github.com/ipfs-shipyard/ipfs-webui/issues/1601), since I suspect we may want to make some tweaks to tutor mode overall after this gets into the hands of users for a bit.

I've also opened https://github.com/ipfs-shipyard/ipfs-webui/issues/1600 for consolidating repeat-use i18n keys, which I can probably take care of in the next week or two.

Thanks all (especially @jay-dee7!) for all of the work!

Issue #1575, opened 2020-08-05
Improve messaging/continuity when node shut down but UI still open


jessicaschilling:
Originally reported as https://github.com/ipfs-shipyard/ipfs-webui/pull/1571#pullrequestreview-461965298 by @lidel but breaking it out into its own issue.

Original follows:

I've found some UX nits when transitioning "online" → "offline" (eg. when I shut down my go-ipfs node):

1. Connection error is displayed, and the graphs get faded out, but they keep showing the last bandwidth values:
   > ![2020-08-05--20-48-30](https://user-images.githubusercontent.com/157609/89451804-0a1fad80-d75d-11ea-86b3-b9528250120b.png)
   
   I'd expect "offline" state to set them to zeroes.
2. _Files_, _Explore_, _Peers_ in navbar are not being correctly disabled. I can go to each and there is either error or an outdated cached data that does not update. If I refresh page (F5) then those screens are replaced with  connection error on status page. 

3. The flow that starts with clicking on ![2020-08-05--21-29-44](https://user-images.githubusercontent.com/157609/89455618-e5c6cf80-d762-11ea-8f9a-902038a7aa7a.png) → Expanded Status → API multiaddr <kbd>Edit</kbd> →  _Settings_  → API multiaddr input is really nice. It encourages user to intuitively discover things.
  
   The problem: when in offline status, clicking on the ![2020-08-05--21-26-39](https://user-images.githubusercontent.com/157609/89455293-61744c80-d762-11ea-85d4-83d3a0615461.png) does nothing. It can't expand node info, because API is down and that block is hidden. Perhaps the fix is to skip expand step and immediately open the  _Settings_ and scroll to API address input?

jessicaschilling:
@lidel - a few thoughts/questions on your original issue:
1. Good catch!
2. True; @rafaelramalho19 and I had noticed this when working through the earlier welcome screen work. At minimum, we need better detection of node offline status in order to automatically replace outdated data with the "could not connect" variant of Welcome page (current behavior, but only happens if you reload). Even better, we would replace it with a "could not connect" message that didn't have the informational boxes from the Welcome page.
     - Side note on this: I *think* that in current state, *eventually* any page does get replaced by the "could not connect" variant of the Welcome page. However there's an interval where what you're describing definitely happens.
3. For this scenario (Status screen immediately after disconnect) what if we ...
[...]

PR #1577, opened 2020-08-07
feat: Messaging continuity when node connection lost

jessicaschilling:
### About
This PR addresses inconsistency of experience when the user loses API connection but still has Web UI open. See https://github.com/ipfs-shipyard/ipfs-webui/issues/1575 for original issue details.

### Done
- On Welcome page, "in this app" and "what is IPFS" boxes only appear if the user is connected
- On Status page, if connection is lost, displays the same "Could not connect to the IPFS API" troubleshooting box as on the Welcome page
- Assorted cleanup required to facilitate this **(note: would appreciate a close code review on this, thanks in advance!)**
    - `AboutWebUI` and `AboutIpfs` get their own components
    -  `IsConnected` gets its own component
    - `IsNotConnected` gets its own component (which is used on the Status page now too)
    - Removed old Status page component `StatusNotConnected` and its associated i18n/storybook stuff

### To do
**Please note that these will require some help from @lidel and/or @rafaelramalho19 ...**

- Improve detection/reload logic so user doesn't get "stuck" on a page with outdated information, not knowing they've dropped API connection
   - This could mean creating a separate "you're not connected any more" page that includes the `IsNotConnected` component, or it could just mean displaying the `IsNotConnected` component at the top of each individual page if connection is lost (this is how it's currently done on Status page)

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/1575.


jessicaschilling:
@lidel -- per our chat earlier today, I believe this addresses everything except improving the detection/reload logic for when connection is dropped. I did move quite a bit around from `WelcomePage.js` into `IsNotConnected.js`, so would appreciate a close review, but everything appears to be working as expected. Let me know if I can help with anything else. Thanks!

src/components/about-webui/AboutWebUI.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import { withTranslation, Trans } from 'react-i18next'
+import Box from '../box/Box'
+
+export const AboutWebUI = ({ t }) => {
+  return (
+    <Box>
+      <h1 className='mt0 mb3 montserrat fw2 f3 charcoal'>{t('welcomeInfo.header')}</h1>
```
rafaelramalho19, line 8: I wouldn't use a `h1` in this component. There should only be one `h1` in the page and this looks more like an `h2` 😄

jessicaschilling: Good catch 😊 fixed in https://github.com/ipfs-shipyard/ipfs-webui/pull/1577/commits/97b43e81fc3166e9453cb8c8b00523218dba9da8

jessicaschilling:
Regarding nit 1, @rafaelramalho19, do you have any suggestions, since you added this bit in originally? 😊 

Regarding nit 2, looks like anchor tags don't work as expected with react-router: https://stackoverflow.com/questions/48223566/using-anchor-tags-in-react-router-4
We could manually specify scroll position, or install react-router-hash-link, but neither of those feel particularly pretty. :(

Regarding detection/reload logic -- is that something you'd want to do in this PR, or break it out into separate work?

rafaelramalho19:
@jessicaschilling related to nit1, here's the fix: https://github.com/ipfs-shipyard/ipfs-webui/pull/1577

Nit 2 should be addressed with react-router-hash-link 😄

jessicaschilling:
@lidel Unfortunately it looks like using react-router-hash-link isn't going to work unless we pull the default use of # in all our URLs. Is direct-linking critical at this point? I might recommend we punt on this until later.

jessicaschilling:
After discussion with @lidel, rearranging Settings page slightly as a workaround:
image

What changed:
- Swapped API Address and Language boxes
- brought analytics disclaimer “No CIDs, filenames …” full width of page and tightened margin between that and “Configure what is collected”
- Removed the superfluous “API ADDRESS” label above the box for changing API - put that i18n item inside the input itself as aria-labelledby={t('apiAddressForm.apiLabel')} for accessibility -- **@rafaelramalho19, can you please confirm that this is the right way to handle this a11y-wise?**

Re-requesting reviews all around. Thanks!

jessicaschilling:
> Would it be ok to hide the JSON editor when API is offline?
Agreed! @rafaelramalho19, do you mind including this in the changes @lidel mentions above? Thank you 😊

jessicaschilling:
Resolved the conflict created by CLI tutor mode merge.

Also, CircleCI behaving itself again after last week's irregularities, so ... @rafaelramalho19 and @lidel, are we thinking we are ready to merge this?

src/components/api-address-form/ApiAddressForm.js
```diff
@@ -21,8 +21,8 @@ const ApiAddressForm = ({ t, doUpdateIpfsApiAddress, ipfsApiAddress = '' }) => {
 
   return (
     <form onSubmit={onSubmit}>
-      <label htmlFor='api-address' className='db f7 mb2 ttu tracked charcoal pl1'>{t('apiAddressForm.apiLabel')}</label>
       <input id='api-address'
+        aria-labelledby={t('apiAddressForm.apiLabel')}
```
rafaelramalho19, line 25: This shouldn't work, since this attribute required a ID, not the text. I think what you're looking for is: `aria-label={t('apiAddressForm.apiLabel')}`

`aria-label` allows you to provide screen-reader text
vs
`aria-labeledby` allows you to indicate an html element that labels the element you're specifying

jessicaschilling: Thank you! Fixed in https://github.com/ipfs-shipyard/ipfs-webui/pull/1577/commits/b01f2e01338ddaa5cb436a42d7cb2108da8ebddd.

jessicaschilling:
@rafaelramalho19 Thanks for the a11y clarity 😊  
Do you mind looking at the other question from Lidel? Re-pasted below since it gets a little lost in the thread. Thanks!

When I shut down my node and open Files screen (without reloading the page), it displays connection error AND import drawer is expanded with "Imported 0 items" message – it hit me how confusing it looks. Quick fix would be to redirect to status page when offline, or at least hide the import drawer, if possible. If we only had Orange error, then user would instinctively refresh the page:
   >

rafaelramalho19:
I wasn't able to replicate because I got redirected to the welcome page, is this replicable? 🤔

jessicaschilling:
@rafaelramalho19 Afraid I was just able to replicate it on my local :(

lidel:
@rafaelramalho19 to replicate, you need to:
1. be on the Status or Peer screen and when node is online and everything works ok
2. shut down your node, so webui loses connection to the API
3. switch to Files screen (without refreshing the page)

lidel:
Perhaps we should merge this as-is and fill new issue for [connection error AND expanded import drawer](https://user-images.githubusercontent.com/157609/90448562-e574ef80-e0e5-11ea-8e9c-8936d5e537cd.png)?
Most of the users will never end up on that screen with that state.

jessicaschilling:
@lidel This makes good sense, will do.

jessicaschilling:
Files page visual continuity issue is in https://github.com/ipfs-shipyard/ipfs-webui/issues/1606. Per https://github.com/ipfs-shipyard/ipfs-webui/pull/1577#issuecomment-684877590, merging.

PR #1591, opened 2020-08-17
fix: view in gateway url now validates if url is acessible

### LINKED ISSUE #1490: Inspect view: "View on IPFS Gateway" still points to the ip-address, not to .localhost
**Describe the bug**
On the inspect view there's a button labled with "View on IPFS Gateway", which still points (in my case) to `127.0.0.1/ipfs/<CID>` instead of `<CID>.ipfs.localhost`.

**Screenshots**
Screenshot_20200510_162131-1

 - OS: Linux
 - Browser: Brave

rafaelramalho19:
Closes #1490


src/bundles/config.js
```diff
@@ -74,4 +86,16 @@ function getURLFromAddress (name, config) {
...
 }
 
+const checkIfGatewayUrlIsAcessible = memoizee(async (url) => {
+  try {
+    const { status } = await fetch(
+    `${url}/ipfs/bafybeiczsscdsbs7ffqz55asqdf3smv6klcw3gofszvwlyarci47bgf354`
+    )
+    return status === 200
+  } catch (e) {
+    console.error(e)
```
lidel, line 96: Additional context will be super useful here:
```suggestion
    console.error(`Unable to use Gateway at ${url}, the public one will be used as a fallback`, e)
```

src/bundles/config.js
```diff
@@ -34,7 +35,18 @@ bundle.selectApiUrl = createSelector(
...
   'selectConfigObject',
-  (config) => getURLFromAddress('Gateway', config) || 'https://ipfs.io'
+  async (config) => {
+    if (!config) return null
+
+    const url = getURLFromAddress('Gateway', config)
+
+    // More info: https://github.com/ipfs-shipyard/ipfs-webui/issues/1490#issuecomment-671633602
+    if (!await checkIfGatewayUrlIsAcessible(url)) {
+      return 'https://ipfs.io'
```
lidel, line 45: We hardcoded `https://ipfs.io` in [a few places](https://github.com/ipfs-shipyard/ipfs-webui/search?l=JavaScript&q=%22https%3A%2F%2Fipfs.io%22), could we create a constant somewhere and import it instead?

src/bundles/config.js
```diff
@@ -1,3 +1,4 @@
+import memoizee from 'memoizee'
```
lidel, line 1: nit: `p-memoize` is 6KB (vs 266KB of memoizee) 

AFAIK the only place we use memoizee is `/src/bundles/peer-locations.js`, and it does not do anything fancy to justify the size.  Can we switch to the smaller lib?

src/bundles/config.js
jessicaschilling: Could I butt in to suggest small language change?
```suggestion
    console.error(`Unable to use the gateway at ${url}. The public gateway will be used as a fallback`, e)
```

rafaelramalho19: Thanks 🙏

src/bundles/config.js
rafaelramalho19: Good find, I changed it :)

src/bundles/config.js
```diff
@@ -29,12 +32,23 @@ bundle.selectConfigObject = createSelector(
...
 bundle.selectApiUrl = createSelector(
   'selectConfigObject',
-  (config) => getURLFromAddress('API', config) || 'https://ipfs.io'
+  (config) => getURLFromAddress('API', config) || DEFAULT_URL
 )
 
 bundle.selectGatewayUrl = createSelector(
   'selectConfigObject',
-  (config) => getURLFromAddress('Gateway', config) || 'https://ipfs.io'
+  async (config) => {
```
lidel, line 40: I believe this is why E2E fails. If I comment out async/await all tests are green.

I don't feel strong with advice here, but IIUC if you want to have async selector, you need to use [async-selector](https://www.npmjs.com/package/async-selector)/[use-async-selector](https://www.npmjs.com/package/use-async-selector), OR (remove async/await and refactor `checkIfGatewayUrlIsAcessible` to use old school blocking [synchronous XHR request](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#Synchronous_request) (less dependencies, but will produce warning when done in the main thread)

test/e2e/setup/test-environment.js
```diff
@@ -15,6 +15,8 @@ class WebuiTestEnvironment extends PuppeteerEnvironment {
 
     const { ipfs, webuiUrl, page } = this.global
 
+    await ipfs.config.set('Addresses.Gateway', '/dns4/ipfs.io/tcp/443/https')
+
```
lidel, line 19: This is not needed :-)
```suggestion
```

lidel (changes requested):

src/bundles/config.js
rafaelramalho19: Thank you for finding the issue 🙏

test/e2e/setup/test-environment.js
rafaelramalho19: You're right, I removed it :)

src/bundles/config.js
```diff
@@ -74,4 +87,16 @@ function getURLFromAddress (name, config) {
   }
 }
 
+const checkIfGatewayUrlIsAccessible = memoize(async (url) => {
+  try {
+    const { status } = await fetch(
+    `${url}/ipfs/bafybeiczsscdsbs7ffqz55asqdf3smv6klcw3gofszvwlyarci47bgf354`
```
lidel, line 93: I figured that this request for empty directory fetches 80KB of HTML. 

There is _even better_ CID that is always present: [`bafkqaaa`](https://cid.ipfs.io/#bafkqaaa) (inlined empty multihash):

```suggestion
    `${url}/ipfs/bafkqaaa`
```

src/bundles/config.js
```diff
@@ -13,6 +17,15 @@ const bundle = createAsyncResourceBundle({
       conf = JSON.stringify(rawConf, null, '\t')
     }
 
+    // More info: https://github.com/ipfs-shipyard/ipfs-webui/issues/1490#issuecomment-671633602
+    const config = JSON.parse(conf)
+    const url = getURLFromAddress('Gateway', config) || DEFAULT_URI
+
+    if (!await checkIfGatewayUrlIsAccessible(url)) {
+      config.Addresses.Gateway = DEFAULT_MULTI_ADDR
+      return JSON.stringify(config)
```
lidel, line 26: After giving it a spin it works fine on _Files_ and _Explore_ screens, but I don't think this will work for _Settings_.

When `if (true)`  you will see that it breaks IPFS Config on _Settings_ page:

> 

We don't want `Addresses.Gateway` to be changed on the _Settings_ page because if user edits and submits new config that sets `Addresses.Gateway = /dns4/ipfs.io/tcp/443/https`, which will break gateway functionality of their node.

We could either:
- (A) not modify original `config`, but create separate value/bundle to represent "available gateway" and switch _Files_ and _Explore_ to use that value if present (the cleanest solution)
- (B) create separate bundle for _Settings_ screen which does not modify `Addresses.Gateway` (ugly fix)

@rafaelramalho19  I'd got with (A) unless you see a better way

lidel (changes requested):

src/bundles/config.js
rafaelramalho19: Good find! Thank you, fixed 👍

src/bundles/config.js
```diff
@@ -13,6 +17,15 @@ const bundle = createAsyncResourceBundle({
       conf = JSON.stringify(rawConf, null, '\t')
     }
 
+    // More info: https://github.com/ipfs-shipyard/ipfs-webui/issues/1490#issuecomment-671633602
+    const config = JSON.parse(conf)
+    const url = getURLFromAddress('Gateway', config) || DEFAULT_URI
+
+    if (!await checkIfGatewayUrlIsAccessible(url)) {
+      config.Addresses.AvailableGateway = DEFAULT_MULTI_ADDR
+      return JSON.stringify(config)
```
lidel, line 26: nit: without below, it will put everything in one line, making JSON editor useless

```suggestion
      return JSON.stringify(config, null, '\t'))
```

src/bundles/config.js
```diff
@@ -13,6 +17,15 @@ const bundle = createAsyncResourceBundle({
       conf = JSON.stringify(rawConf, null, '\t')
     }
 
+    // More info: https://github.com/ipfs-shipyard/ipfs-webui/issues/1490#issuecomment-671633602
+    const config = JSON.parse(conf)
+    const url = getURLFromAddress('Gateway', config) || DEFAULT_URI
+
+    if (!await checkIfGatewayUrlIsAccessible(url)) {
+      config.Addresses.AvailableGateway = DEFAULT_MULTI_ADDR
```
lidel, line 25: Persisting `AvailableGateway` in `config` smells :biohazard:  
We are mutating config and then lie to user  on _Settings_ screen:




My main worry is this will bite us in the future in a way that breaks users node (invalid config is saved, then node fails to start). Let's avoid this line of hacking at all cost.

@rafaelramalho19 can we leave `config` as-is and store `AvailableGateway` in something like `window.localStorage` instead?

lidel (changes requested):

src/bundles/config.js
rafaelramalho19: Removed this, but thanks for the heads up!

src/bundles/config.js
rafaelramalho19: You're right, I moved it into a separate persistence bundle :)

lidel (approved):
LGTM!

I was about to add a note how to address #1490 in a way that accounts for all edge cases and I ended up implementing it in pseudocode, so I just added it in 7c8b455 to this PR :upside_down_face: – hope that's ok!

@rafaelramalho19 take a look if mentioned commit makes sense

rafaelramalho19:
LGTM @lidel, thank you 🙏

PR #1599, opened 2020-08-25
feat: improve breadcrumbs functionality

### LINKED ISSUE #840: Files: icon for parent directory
The "go one level up" icon is a strong accent at the top of every subdirectory. Bit too strong?
What  if we remove the icon (or just fade it out a bit) and introduce some whitespace that way?

| Now |
| ---- | 
|  |

 No Icon | 
| ---- | 
 |  | 

| Fade Out (`o-20`) |
| ---- |
|  |

Thoughts?

rafaelramalho19:
## Notes

- Improves breadcrumbs component.
- Adds context menu when right clicking a breadcrumb clickable item.

Fixes #840.


jessicaschilling:
Updated Joyride tour in https://github.com/ipfs-shipyard/ipfs-webui/pull/1599/commits/91a0d3a6232613722f5a2734a8eeb4ad8324b387 - screenshot below:
image

jessicaschilling (changes requested):
The following right-click actions for breadcrumbs seem to work perfectly 😊
- Share link
- Copy CID
- Inspect
- Rename
- Delete

The following right-click actions don't quite work:
- Pin: 
   - If a directory is un-pinned, right-clicking its breadcrumb and selecting "Pin" pins it successfully
   - If a directory is already pinned, right-clicking breadcrumb still says "Pin" (not "Unpin")
- Download: This downloads a .txt file that actually contains (valid) dir-index-html HTML for displaying the directory! :rofl: Voting that we should just remove this option entirely from breadcrumbs (it doesn't appear when right-clicking a directory in the files table).

Issues with _moving_ files via dragging:
- None 😊  aside from the known issue in https://github.com/ipfs-shipyard/ipfs-webui/issues/1612

Issues with _importing_ files via dragging:
- I can drag a file into the main file list (not the breadcrumb) when I am in the home directory, or home+1 directory (e.g. `home/cats`), but not any deeper (e.g. not `home/cats/moar cats` or  `home/cats/moar cats/even moar cats`) -- it gives "folder with that name already exists" error. (Using the "Import" button in any directory works just fine.)
- Dragging onto any level of breadcrumb actually just imports it into the presently displayed directory.

Two more things from earlier:
- Per #840 (comment) - would it be possible to display a directory's CID on hover of the breadcrumb, as in the mockup in that comment? That also is a subtle clue that those items are interactive.
- We're using `folderExists` (line 7 in `public/locales/foo/notify.json`) for the error whenever a user tries to import a file with an already existing name, too. Should we just make this more general ("an item with that name already exists") or should we add a separate `itemExists` i18n key?

Thank you so much for continuing to untangle this 😊  

It looks possible to drag onto a higher-level directory in a breadcrumb (i.e. the hover underline activates) if one is in a subdirectory, but I believe it's actually just trying to import to the current directory. For example, say I have orange.png in files/cats but not in the top level. If I am in files/cats in Web UI and I drag my local copy of orange.png to files in the breadcrumb, I get the warning that A folder with that name already exists even though it doesn't in the top-level directory. (Note as well: that warning message should read 'item' instead of 'folder' to account for situations like this one.)
Alternate take on this: Maybe it's actually just trying to upload to the currently displayed directory instead of the higher directory? (Having a hard time teasing out what's actually going wrong here.)

lidel (changes requested):
Love the improved breadcrumbs! Before we merge, need to address what @jessicaschilling said + 

- `Download:` as @jessicaschilling noted, either remove it or fix (I'd vote fix: wire it up to download TAR archive using the same code path as one when a custom directory is created when multiple  items are marked with checkboxes and Download is clicked on the bottom drawer)

- > Should we just make this more general ("an item with that name already exists") or should we add a separate `itemExists` i18n key?

  I vote for making it generic, that's what we did in other places.

- :dizzy: breadcrumb for parent has [all the context menu items](https://user-images.githubusercontent.com/157609/92394466-e49e0f00-f121-11ea-9f44-f482ae1ca763.png), but its parents [only have three items](https://user-images.githubusercontent.com/157609/92394558-09928200-f122-11ea-8acf-4c99ebd5a942.png)
 – bit confusing, are we able to unify this?

- :bug: Found a D&D bug: while in "Files, click on the "Files" in navbar and keep  it clicked, then drag it into Files ;-) – [`Unhandled Rejection (TypeError): entry is null`](https://user-images.githubusercontent.com/157609/92392881-414bfa80-f11f-11ea-825f-5c03235bb939.png) is produced 

- :bug: When I right click on a crumb, its menu is displayed. When I click on another, its menu does not display: [I see menu for previous crumb + browser-native menu instead](https://user-images.githubusercontent.com/157609/92393282-f383c200-f11f-11ea-8ba5-1a76ab4dfc54.png). Was able to reproduce in both Firefox and Chromium. Are we able to close existing menu if a different crumb is right-clicked? 

- :bug: :dizzy:  dragging file item onto the crumb of parent directory _SOMETIMES_ copies it instead of moving
  I've seen https://github.com/ipfs/go-ipfs/issues/7646 but was not able to reproduce it in CLI, then I had it happen once for me in WebUI while testing this PR, but then worked ok.. unclear is its a bug in webui (old dir listing being cached) or ephemeral go-ipfs bug (sometimes not refreshing MFS root CID?). It may be tricky to debug, so as this is just a mild annoyance, and not something that breaks things or lose data, I would not block this PR, and fill separate issue for tracking this.

rafaelramalho19:
- [X] Unpin label in context menu
- [X] Download in a directory - this was actually the assumption that I made that the last item of the breadcrumbs would be always a file 🤦 Sorry about that f(a)il(e), changed the method to use `ipfs.files.stat` to get both CID and Type
- [X] Dragging onto any level of breadcrumb actually just imports it into the presently displayed directory. - I've updated the code from the dragging already uploaded files and forgot to update the one from new files 🤦 

- [ ] @jessicaschilling "would it be possible to display a directory's CID on hover of the breadcrumb, as in the mockup in that comment?" That would force us to run `ipfs.files.stat` for every breadcrumb. We could make this async (after loading with a small delay), but don't you think that the `cursor: pointer` is sufficient feedback for the user? I really don't want to impact the normal usage of the files page because of a small interaction 🤔 

- [X] Changed translation to "an item", makes more sense 👍 

> breadcrumb for parent has all the context menu items, but its parents only have three items – bit confusing, are we able to unify this?

The logic is that if an item is a directory, you can only do the 3 options, if it's a file you can do the full list

- [X]  Found a D&D bug: while in "Files, click on the "Files" in navbar and keep it clicked, then drag it into Files ;-) . Fixed every drag & drop of text inside webui. If you find a weird use-case where this should be possible please tell me


> When I right click on a crumb, its menu is displayed. When I click on another, its menu does not display: I see menu for previous crumb + browser-native menu instead. Was able to reproduce in both Firefox and Chromium. Are we able to close existing menu if a different crumb is right-clicked?

@lidel unfortunately I don't think it's possible with the current solution, since we create an invisible overlay around the dropdown to catch mouse interactions. There's 2 solutions that come to my head:

- Remove the overlay and change the click handling to a global listener that checks if there was a click the dropdown and just close it if that's the case.
- Make a onContextMenu handler on the overlay that simply prevents the default behavior (so that you don't see the native right click action)

What option would you prefer?




-- 


Thanks both of you for your deep review 🙏

jessicaschilling:
Bummer about the hover being expensive. We can leave it out.

lidel:
@rafaelramalho19 IIUC  going with _"global listener that checks if there was a click the dropdown and just close it if that's the case"_ would produce better UX:

- We want to disable native right click action and only display webui ones when interacting with breadcrumbs or file list
- There should be only one context menu at a time
  - if a new one is opened an existing one should be automatically closed

lidel (changes requested):
@rafaelramalho19 [dragging image onto breadcrumb breaks the app](https://user-images.githubusercontent.com/157609/92603882-1d6bee80-f2b0-11ea-85ae-481f87671ebe.gif) – mind applying simililar fix as you did for strings and navbar?

public/locales/en/files.json
```diff
@@ -91,7 +91,10 @@
     },
     "step2": {
       "title": "Breadcrumbs",
-      "paragraph1": "The current work directory, you can navigate through the folder hierarchy by clicking on them."
+      "paragraph1": "These breadcrumbs display the path to your current directory. You can…",
+      "bullet1": "Click an individual breadcrumb to go to that directory",
+      "bullet2": "Right-click a breadcrumb to get the actions menu for that directory",
```
jessicaschilling, line 96:
```suggestion
```

We'll need to remove this Joyride hint if we don't have context menu yet.

jessicaschilling (changes requested):
Quick note on the Joyride helptext, otherwise I'm good whenever @lidel is 😊

public/locales/en/files.json
rafaelramalho19: We have the context menu, what's not possible is to right click on another breadcrumb WHILE the context menu is open :)

jessicaschilling: @rafaelramalho19 apologies, I misunderstood the comment chain. Never mind 😊

jessicaschilling (approved):

lidel (changes requested):
@rafaelramalho19 drag-and-drop works great now, but recent commit introduced regression in how PDF and videos are rendered (images are fine, so we missed it).

before they took entire free space,   now they are a narrow column:


| before | now |
| ---- | ---- |
| ff2-2020-09-09--18-38-26 | ff-2020-09-09--18-34-08 |

lidel (changes requested):
@rafaelramalho19 thanks, PDF previews look fine again.

<del>Q: you mentioned that we should keep joyride for them, buy I've rebuilt  this PR and now I get no context menu on breadcrumbs at all. Mind checking on your end?</del> nvm, it was due to HTTP API I used for testing being slow to respond

lidel (approved):
LGTM. Let's merge?

ps. The slowdown I experienced is not related to this PR, but was already present in go-ipfs.
I filled https://github.com/ipfs-shipyard/ipfs-webui/issues/1618 to track it.

Issue #1576, opened 2020-08-06
Bulk "Download" action does not work with go-ipfs 0.5+ 


lidel:
When more than one file is selected with checkbox, and Download button is clicked, then a TAR archive is prepared, instead of downloading each file separately.

This is done via `/api/v0/get` API with `archive=true` flag. 

This fails because API port requires `POST` now and download call is done with `GET`, which produces HTTP 405 error:

![2020-08-06--23-53-43](https://user-images.githubusercontent.com/157609/89586783-9a82ee80-d840-11ea-8f73-ce0ca9fe74e8.png)

The fix is either:

- switch to `POST` 
- switch URL from API port to Gateway one (which also exposes `/api/v0/get` but allows `GET` as well)

PR #1611, opened 2020-09-04
fix: multiple item download with go-ipfs 0.5+

rafaelramalho19:
Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/1576


lidel (approved):
:+1:

PR #1613, opened 2020-09-04
feat: improved support for remote API

### LINKED ISSUE #836: Make it possible to use Web UI with a remote IPFS API
Lots of people want to use Web UI to monitor and manage IPFS nodes hosted on a remote server rather than on their local machine. That is totally reasonable. 

But **it is not currently supported**. The IPFS API is not authenticated. It currently expects to be accessed by privileged users on the local machine, and should not be exposed to the internet. Once we have an authentication mechanism for the API we'll ensure that Web UI works with it.

Please see: https://github.com/ipfs/go-ipfs/issues/1532

In the meantime we'll make the subset of Web UI that can work against the read-only / gateway API that would be safe expose from a remote machine. See #835

A suitably motivated individual _could_ set up secure proxy + firewall or vpn to access their api port securely from a remote machine. Pervious iterations of Web UI were hardcoded to use an API on localhost, but v2 will do nothing special to prevent trying to use a remote api address. It's worth reiterating that its not a currently supported configuration, and we don't have the bandwidth to deal with support requests if it doesn't work for you. We will make this work in the future, in a secure way, once core supports it.

### Previous reports of this issue

- https://github.com/ipfs-shipyard/ipfs-webui/issues/637
- https://github.com/ipfs-shipyard/ipfs-webui/issues/628
- https://github.com/ipfs-shipyard/ipfs-webui/issues/594
- https://github.com/ipfs-shipyard/ipfs-webui/issues/591
- https://github.com/ipfs-shipyard/ipfs-webui/issues/95
- https://discuss.ipfs.io/t/how-can-i-enable-remote-connection-to-webui/698/3

### LINKED ISSUE #1031: Bug Report: WebUI through browser is not working
**Describe the bug**
I cannot view the status of IPFS from my browser

**To Reproduce**
1. Install via `brew cask install ipfs`
2. Install the ipfs-companion Firefox extension
3. Open "IPFS Desktop.app" and view the successful peering from within the app
4. Select "Open WebUI" from the ipfs-companion menu and get an error that `ipfs daemon` needs to be run.

**Expected behavior**
I expect the WebUI to be accessible from any interface that can load localhost:5001/webui

**Screenshots**
View from IPFS Desktop:
Screenshot 2019-05-07 17 45 24

View from localhost:8080:
Screenshot 2019-05-07 17 45 06

 - OS: macOS 10.14.4
 - Browser: Firefox
 - Version: go-ipfs v0.4.19

### LINKED ISSUE #1194: reverse proxy
i use nginx reverse proxy with ssl for the ipfs webui, and bind a domain for it,

the ipfs conf:
"API": "/ip4/127.0.0.1/tcp/6501"
"Gateway": "/ip4/127.0.0.1/tcp/6502"

on the Files tab, i can not download any file, the download url like:
http://127.0.0.1:6502/ipfs/QmRtUpjjJh3WavoYTBPXS2r7hUZuwvYYc5VvFc96k79Q6x

how can i change the url to my domain form

### LINKED ISSUE #1389: Support for custom API path
Hi,

I have changed the location of the go-ipfs daemon. For example, this is what I get in the networks tab in browser when launching webui.

POST http://localhost:3000/api/v0/stats/bw?stream-channels=true net::ERR_CONNECTION_REFUSED

I would like to have the webui access a similar location, but with /v1/ instead of /v0/. Is there a away to change this to /v1/ in the ipfs-webui code, or otherwise, so that the webui can access the endpoint ?

### LINKED ISSUE #1565: Allow custom address for the gateway
**Describe the bug**
The gateway is set to `http://0.0.0.0:8080/` while I'm connecting from an external computer through a VPN

This bug affects the app since content in not watchable directly in the WebUI

**To Reproduce**
1. Set your API address and gateway address to 0.0.0.0 (ip4) or :: (ip6)
(1.1 Set your CORS allow-origin to "*")
2. Connect to the WebUI with an external computer
3. See that the gateway is the one from the config instead of the real (external) address

**Expected behavior**
We should be able to set a custom gateway address

### LINKED ISSUE #1586: Support API entered as URL with Basic Auth
Support API as URL with Basic Auth was introduced in https://github.com/ipfs-shipyard/ipfs-redux-bundle/pull/28/
Deprecating ipfs-redux-bundle and switching to ipfs-provider (https://github.com/ipfs-shipyard/ipfs-webui/pull/1563) introduced a small regression: it is no longer possible to connect to API defined as URL with basic auth credentials. Entering URL and pressing Submit does nothing:

> 

We need to restore this feature in webui (and release new ipfs-provider, if needed).

Ref. 
- Simple proxy that adds basic auth to existing API can be found in https://github.com/ipfs-shipyard/ipfs-redux-bundle/pull/28/#pullrequestreview-310399308 (useful for testing this)

lidel:
> Closes #1586
> Closes #1565
> Closes #836
> Closes #1194
> Closes #1389
> Closes #1031

This PR adds support for connecting to remote API that is protected by HTTP Basic Auth (#1586), enables entering custom configuration (a hidden  feature for advanced users) and adds end-to-end tests.

## Terse summary

- There is UI for changing the API address on the _Settings_ screen
- Custom API address is stored in and read from `window.localStorage[ipfsApi]`
- `ipfsApi` is a `string` or a `null`
- If `ipfsApi` is `null`, [ipfs-provider](https://github.com/ipfs-shipyard/ipfs-provider) will try default address at `/ip4/127.0.0.1/tcp/5001` (and same-origin for a good measure)
- `ipfsApi` can be a  multiaddr, URL or JSON with config for ipfs-http-client used by webui
- User can enter URL with inlined basic auth  user and password, we convert them to advanced config before passing to http client.
- E2E tests for using remote API with basic auth (URL or JSON, by setting localStorage[ipfsApi] or via Settings screen)
- Updated some dependencies related to E2E and made things bit more robust

### Settings page

It is now possible to enter URL with inlined basic auth:

> Screenshot_2020-09-04 Settings IPFS

JSON with equivalent config for ipfs-http-client is also accepted:
> Screenshot_2020-09-04 Settings IPFS(1)

Note: this is sort of an easter egg, we should not mention it in UI. This is mostly future-proofing. If someone has a very custom setup, webui will be able to support it without the need for a new release or a fancy UI for that one person.

### Status page

Default state is the same(multiaddr), so most of the users will see no difference:

> Screenshot_2020-09-04 Status IPFS(1)

If a custom URL was entered on _Settings_, we also show URL on _Status_:

> Screenshot_2020-09-04 Status IPFS

If JSON was entered, then we show it as-is. Here, a custom config with Basic Auth header:

> Screenshot_2020-09-04 Status IPFS(3)

I believe this is so niche, power user feature, that there is no point in making UI for this. Edit button is enough.
Instead of showing JSON, we could just print `[JSON config]` placeholder or something – @jessicaschilling thoughts?


jessicaschilling:
@lidel LGTM -- thank you for the super-detailed screenshots 😊
Agree that just showing a message rather than the full JSON config is a good idea - plus reduces the risk of folks not inadvertently sharing stuff in screenshots.
How about `Custom JSON configuration`? (note no square brackets)

jessicaschilling (approved):

lidel:
@jessicaschilling done:

> 

One can test this even without basic auth setup, this JSON will work: `{"url":"http://127.0.0.1:5001/api/v0"}`

lidel:
Oops, I need to fix E2E tests, sorry!

lidel:
Passing config object to `ipfs-http-client` turned out tot be brittle and risky (eg. right now the object gets mutated and user sees different JSON than the one she entered). 

I've addressed that in https://github.com/ipfs-shipyard/ipfs-provider/pull/24 but need to update this PR to use documented parameters instead of `url`. 
Probably safer to do it after some sleep, so will get back to this tomorrow.

lidel:
Switched to ipfs-provider 1.1.0, which does not mutate the custom config, so the "undocumented `url` attribute"  is no longer visible to the user. Only [documented](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-http-client#importing-the-module-and-usage) parameters for `ipfs-http-client` are used:

Screenshot_2020-09-09 Settings IPFS

@rafaelramalho19 ready for your final review.

test/e2e/remote-api.test.js
```diff
@@ -0,0 +1,247 @@
...
+    proxyReq.setHeader('Host', new URL(remoteApiUrl).host)
+  })
+
+  proxy.on('error', function (err, req, res) {
+    res.writeHead(500, { 'Content-Type': 'text/plain' })
+    res.end(`proxyd error: ${JSON.stringify(err)}`)
+  })
+
+  proxyd = http.createServer((req, res) => {
+    // console.log(`${req.method}\t\t${req.url}`)
```
rafaelramalho19, line 60: You forgot to remove this? :)

rafaelramalho19 (approved):

test/e2e/remote-api.test.js
lidel: no, left in on purpose in case someone wants to debug this and is unsure where to add console log (there is server, then proxy inside of server, then we override stuff...) ;-)

src/bundles/ipfs-provider.js
```diff
@@ -104,33 +104,49 @@ const readAPIAddressSetting = () => {
...
 /**
- * Attempts to turn cast given value into `URL` instance. Return either `URL`
- * instance or `null`.
+ * Attempts to turn cast given value into URL.
+ * Return either string instance or `null`.
  * @param {any} value
  * @returns {string|null}
  */
 const asURL = (value) => {
+  if (!value || typeof value !== 'string') return null
```
Gozala, line 116: NIT: I don't think it's really necessary, because URL will stringify and if it's not valid URL it will error anyway. Before this change passing a URL instance would have worked with this change it no longer will, I don't remember if that can be a problem.

src/bundles/ipfs-provider.js
```diff
@@ -104,33 +104,49 @@ const readAPIAddressSetting = () => {
...
+ * Attempts to turn cast given value into Multiaddr.
+ * Return either string instance or `null`.
  */
 const asMultiaddress = (value) => {
-  if (value != null) {
-    try {
-      return multiaddr(value).toString()
-    } catch (_) {}
-  }
+  if (!value) return null
```
Gozala, line 128: NIT: I'm not fan of `!value` as it's too ambiguous (in terms of conveying intent) and requires knowledge of some of the JS quirks to fully understand e.g. `!''` is `true` and I often wonder if author intended that or simply did not consider.

src/bundles/ipfs-provider.js
```diff
@@ -239,7 +255,31 @@ const initIPFS = async (store) => {
   store.dispatch({ type: 'IPFS_INIT_STARTED' })
 
   /** @type {Model} */
-  const { apiAddress } = store.getState().ipfs
+  let { apiAddress } = store.getState().ipfs
+
+  // if a custom JSON config is present, use it instead of multiaddr or URL
```
Gozala, line 260: Would be nice to encapsulate this `apiAddress` reading into a single function.  Maybe `asAPIAddress` should become `asAPIOptions(string) => string | { ... } | null`. If I recall correctly `apiAddress` is passed through `JSON.stringify` anyway before it's stored to localStorage.

Gozala (approved):

src/bundles/ipfs-provider.js
lidel: Good idea. I've removed double JSON serialization and renamed `asAPIAddress` to `asAPIOptions` in [`83009bb` (#1613)](https://github.com/ipfs-shipyard/ipfs-webui/pull/1613/commits/83009bbca5728ba2a3a5bd19ceb09c66f830048b)

PR #1616, opened 2020-09-08
feat: peers table enhancements

### LINKED ISSUE #1602: feat: Peers page table enhancements
Note: This issue includes and therefore supersedes https://github.com/ipfs-shipyard/ipfs-webui/issues/1037 and https://github.com/ipfs-shipyard/ipfs-webui/issues/1117.

### Summary
Enhance Peers page by adding table columns for transfer rates and IPFS version, and removing unused "Notes" column.

### To do
- [x] Remove "Notes" column from Peers page table
- [ ] Add sortable "In/Out" column between "Latency" and "Peer ID" columns, formatted per mockup:
     - Display `TotalIn` and `TotalOut`, or `--` if `sum(TotalIn, TotalOut)` = 0
     - Sort by `sum(TotalIn, TotalOut)`
     - For units, express as the largest unit that can be used while keeping the value above 1 (i.e. let's avoid things like `0.2 kB` in favor of `211 B`)
     - If a transfer is currently happening (e.g. if `sum(RateIn, RateOut)` > 0 ), add an "indicator light" whose colors are the same as the in/out colors on Status page; for edge case in which out/in values are the same, split the circle in two as per mockup
     - Include browser-native hover state indicating current `RateIn` and `RateOut` as well as `TotalIn` and `TotalOut`, per mockup
- [x] For "Peer ID" column, make peer IDs clickable, with an action of copying that peer ID to clipboard and feedback using green snackbar: "Peer ID copied to clipboard."
- [x] For "Connection" column, make row content clickable, with an action of copying that multiaddr to clipboard and feedback using green snackbar: "Connection address copied to clipboard."
- [x] For "Connection" column, also append `direction` to the hover text as such: `/multi/addr (outbound)`  (note this doesn't appear if using js-ipfs)
- [ ] Add sortable "Agent" column at the far right, formatted per mockup, using `AgentVersion` via `ipfs id <PeerId>`
     - [x] Include hover state indicating what type of activity occurred, i.e. `DHT`, `Bitswap` (note this doesn't appear if using js-ipfs)
- [ ] <del>Make items in "Peer ID" column links that open the Explore page with that particular IPNS address</del> 
  (peers dont publish anything under their PeerID by default, this would be broken for most of peers)
- [x] Replace monospace font throughout table with standard Inter for everything except the values in the "Peer ID" column

### Mockup screenshot
peers-page-transfer-status

jessicaschilling:
> Closes #1602

### Summary
This PR stubs out but doesn't finish #1602. All elements are on the page, with dummy values and visual styling as outlined in https://github.com/ipfs-shipyard/ipfs-webui/issues/1602, but the following still need to be wired up:

#### "In/Out" column:
- Real values for `rateIn`, `rateOut`, `totalIn`, `totalOut`  in `peer-locations.js` to display in table body and hover state
- Fallback to `--` if `sum(TotalIn, TotalOut)` = 0
- Displaying human-friendly units as described in https://github.com/ipfs-shipyard/ipfs-webui/issues/1602#issue-686601595
- Indicator light logic as described in https://github.com/ipfs-shipyard/ipfs-webui/issues/1602#issue-686601595 (SVG is inline now)
- Sortability as described in https://github.com/ipfs-shipyard/ipfs-webui/issues/1602#issue-686601595

#### "Peer ID" column:
- Make Peer IDs clickable, with an action of copying that Peer ID to clipboard
- Upon copy, feedback using green snackbar: "Peer ID copied to clipboard."

#### "Connection" column:
- Make row content clickable, with an action of copying that multiaddr to clipboard
- Upon copy, feedback using green snackbar: "Connection address copied to clipboard."
- [x] Append `direction` to the hover text as such: `/multi/addr (outbound)` (note this doesn't appear if using js-ipfs)

#### "Agent" column:
- Real value for `agent` in `peer-locations.js` based on `AgentVersion` (via `ipfs id <PeerId>`)
- Sortability (A-Z should be fine here)
- [x] Include hover state indicating what type of activity occurred, i.e. `DHT`, `Bitswap` (note this doesn't appear if using js-ipfs)

#### Screenshot
image


jessicaschilling:
@lidel and/or @rafaelramalho19 -- I realize it's a stretch to get this into 2.11, but wanted to at least stub out something that would be easier to wire up in 2.12. Happy to plug away at any guidance you might want to give yourselves if that's easier than just asking you to do it. 😊

lidel:
Sidenote: when we ship these improvements we should also update geoip database (https://github.com/ipfs-shipyard/ipfs-geoip/issues/68#issuecomment-707784491), because the moment we bring people's focus to the Peers screen, they will notice faster-than-lings pings ;) 
It does not need to be part of this PR, but should land in the same release for the biggest "improvement" impact.

lidel:
Wired up copying:

I realized showing the green snackbar was not the best UI for providing feedback, especially when details behind each cell can be copied separately.

After some trial&error I implemented something different, a form of in-place confirmation:

> copying-ui

IMO its less noisy, and more intuitive for this specific UI, but lmk your thoughts @jessicaschilling

jessicaschilling:
@lidel - great idea!

jessicaschilling:
@lidel One quick thought - could we shorten the dwelltime of the phrase "Copied!"? 2sec should be enough.

lidel:
Quick note on unblocking 80% of this PR:
- comment out IN/OUT and AGENT columns for now + fill issue to follow-up:
  - getting this data is too expensive atm, some  additional caching would need to happen
  - mixing active and totals may be too confusing
- restore NOTES column in place of AGENT one

lidel:
@jessicaschilling  sadly I was not able to make things performant enough to run decently on low-power machines:
    
- in/out adds 1 HTTP request PER peer
  (and cant be cached as it is dynamic)
- agent also adds one (but could be cached)
    
We need better API than HTTP, maybe RPC over websockets or pubsub.
    
To salvage this PR I replaced mentioned columns with "open libp2p streams" column, which does not require any additional requests, but is pretty useful, gives us insight if peer was used only for kad DHT, bitswap,   pubsub etc:

Screenshot_2021-03-29 Peers IPFS


Mind doing quick :eyes:?

lidel (approved):

jessicaschilling:
@lidel Works for me. Two thoughts:
- Is there any point in having all cell values in this table click-to-copy? For example, copying the latency value doesn't really seem to have a use case. Might be worth limiting this just to the Peer ID and Connection columns.
- The 🤔 in the Open Streams column is charming, but could be more useful if the hover text replaced the 🤔 with what that actually means.
Thanks!

lidel:
Thanks! Should be ready:
- Removed interaction from Latency column, I kept copying for everything else as I find them useful
- Hover text for :thinking: is  `[unnamed]`

PR #1620, opened 2020-09-10
Have CI publish to the Collab Cluster

mburns:
similar to https://github.com/ipfs/ipfs-docs/pull/195

When CI/CD runs and generates a CID, we need to pin that CID directly to the Websites (collab) cluster. Otherwise, when CircleCI's build finishes, the ipfs instance will dissapear and the CID will be irretrievable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,7 @@ jobs:
         environment:
           DOMAIN: webui.ipfs.io
           BUILD_DIR: build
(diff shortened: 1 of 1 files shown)


rafaelramalho19 (approved):

lidel (approved):
Should be fine: iiuc this will be pinning to collab cluster only on `master` and release (tag), won't pin PRs.

mburns:
@lidel correct, whenever the  `deploy` runs.

Issue #1621, opened 2020-09-11
The PIN icon disappears unexpectedly


akimoto72738:
**Describe the bug**
The pin icon will disappear and can not unpinned if you want check and list files you have pinned (pins button) 

**To Reproduce**

> 1. Go to 'file' #pin icon is normal
> 2. Click on 'pins'
> 3. back to 'file' and check pin icon
> 4. See error

**Screenshots**
### At the beginning, everything was fine
![捕获](https://user-images.githubusercontent.com/67720156/92909633-e33e5080-f459-11ea-983a-5fcf63220191.JPG)
### until you clicked the button
![捕获1](https://user-images.githubusercontent.com/67720156/92909955-32848100-f45a-11ea-8ed7-c0ce2da742ed.JPG)
### this page will show all file you pinned,Should it show pin icons? I don't know, The focus is on the next image
![捕获3](https://user-images.githubusercontent.com/67720156/92910488-acb50580-f45a-11ea-96b9-64b0fca9a763.JPG)
### Click the button to return to the page we just started, and the pin icon disappears
![捕获4](https://user-images.githubusercontent.com/67720156/92911095-24833000-f45b-11ea-95db-50dd1e8881e2.JPG)
### also files can not be unpinned
![捕获5](https://user-images.githubusercontent.com/67720156/92911739-bc811980-f45b-11ea-8ce5-22b639389eca.JPG)

**Desktop (please complete the following information):**
 - Browser [Opera Lastest Version]
 - WebUI Version [2.11.0]

jessicaschilling:
Thanks, @akimoto72738 - we'll go through this in our next weekly triage.

jessicaschilling:
@Gozala and/or @rafaelramalho19, can you please investigate? Thank you!

PR #1633, opened 2020-09-15
fix: files pin status now shows correctly

rafaelramalho19:
Fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/1621


lidel (approved):
Agree, `.toString()` should be enough: both CIDs come from the same source.

(When we get a CID from unknown third party, then `.toV1().toBaseEncodedString('base32')` is safer as always produces CID in deterministic version and base)

Gozala:
> .toV1().toBaseEncodedString('base32')

Note that `toBaseEncodedString` was / is been deprecated so `toString('base32')` would be a way to go.

Issue #1631, opened 2020-09-14
Files: missing actions in folder context menus


lidel:
I noticed some actions are available via bottom drawer menu (when selecting a single folder), but are missing from the context menu (file list, breadcrumbs):

> ![2020-09-14--21-15-18](https://user-images.githubusercontent.com/157609/93128249-7c29d180-f6cf-11ea-9211-44ebdb30385a.png)

Any reason we should not show the same actions in both places? 

I did quick check and drawe menu actions work as expected: 

- :green_heart: Share link of a folder works as expected
- :green_heart: Download of a folder works as expected with fix from #1630 
- :green_heart: Delete of a folder works as expected
- :green_heart: / :broken_heart: Rename of a folder works fine, unless the directory is named `test` :upside_down_face:  (or is it just me?)

rafaelramalho19:
Maybe you already had a directory called `test`?

jessicaschilling:
AFAIK absolutely no reason not to put those options in right-click context menu ... if they work.

jessicaschilling:
Added labels to this, moved it into "Quick Fixes" and the v2.12 milestone -- let's discuss if this doesn't seem viable.

lidel:
cc @Swedneck – do you have similar issue with rename?

rafaelramalho19:
@jessicaschilling I would change this to v2.11.2, since this is a important patch before next desktop release

jessicaschilling:
Sounds good.

Sovxx:
#1405 looks pretty similar

ldeffenb:
@jessicaschilling I see the "bottom drawer menu" in the illustration above, but how do you get that in the Files page of the webui?  When I click on anything in the list, it drills into it if a directory or displays it if a file.  What's the magic to just "select" it to see the bottom drawer menu?

jessicaschilling:
@ldeffenb The checkboxes to the left of each item in the list enable you to select multiple (or even just one) items and act on them:

![image](https://user-images.githubusercontent.com/1507828/94305414-5c47a880-ff2e-11ea-8114-eb73ac91c2af.png)

ldeffenb:
[...]

PR #1639, opened 2020-09-17
fix: context menu now has all options when it's a directory

rafaelramalho19:
Fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/1631#issuecomment-692903563

- This was a regression on behavior compared to pre-breadcrumbs.


lidel (approved):
:ok_hand:

PR #1646, opened 2020-09-18
fix: allow json and text fiels to be in preview file component

### LINKED ISSUE #1642: Preview of .json files is broken
To reproduce:
1. Import .json file
2. Try opening it in Files
3. Seems that plain text file with JSON gets stringified incorrectly:
   > 

Example:
- https://webui.ipfs.io/#/ipfs/QmUzaX5HWh68LTrTu2crvNwCsPebZgn5Hxs7Nwx3y8Dktz/app.json

### LINKED ISSUE #1647: Failed to view TXT file via WebUI
**Describe the bug**
Failed to view TXT file via WebUI

**To Reproduce**
1.click txt file to view content

**Screenshots**
屏幕截图 2020-09-19 104603
屏幕截图 2020-09-19 104633

 - Browser [Opera Lastest version]
 - Version [2.11.1]

rafaelramalho19:
Closes #1642 
Closes #1647
Important info: https://github.com/achingbrain/uint8arrays#tostringarray-encoding--utf8

## Questions

- Did `ipfs.cat` start returning uint8arrays recently?
- Do we need to parse uint8arrays in another function?

### DIFF
--- a/src/files/file-preview/FilePreview.js
+++ b/src/files/file-preview/FilePreview.js
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react'
+import React, { useCallback, useEffect, useRef, useState } from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
 import { connect } from 'redux-bundler-react'
@@ -9,7 +9,8 @@ import ComponentLoader from '../../loader/ComponentLoader.js'
 import './FilePreview.css'
 import CID from 'cids'
 import { useDrag } from 'react-dnd'
-import concat from 'it-concat'
+import fromUint8ArrayToString from 'uint8arrays/to-string'
+import Button from '../../components/button/Button'
 
 const Preview = (props) => {
   const { name, size, cid, path } = props
@@ -27,13 +28,28 @@ const Preview = (props) => {
   </div>
 }
 
-const PreviewItem = ({ t, name, cid, size, type, availableGatewayUrl: gatewayUrl, read }) => {
+const PreviewItem = ({ t, name, cid, size, type, availableGatewayUrl: gatewayUrl, read, onDownload }) => {
   const [content, setContent] = useState(null)
+  const [hasMoreContent, setHasMoreContent] = useState(false)
+  const [buffer, setBuffer] = useState(null)
 
-  const loadContent = async () => {
-    const buf = await concat(await read())
-    setContent(buf.toString())
-  }
+  const loadContent = useCallback(async () => {
+    const readBuffer = buffer || await read()
+    if (!buffer) {
+      setBuffer(readBuffer)
+    }
+
+    const { value, done } = await readBuffer.next()
+    const previousContent = content || ''
+
+    setContent(previousContent + fromUint8ArrayToString(value))
+    setHasMoreContent(!done)
+  }, [buffer, content, read])
+
+  useEffect(() => {
+    loadContent()
+  }, // eslint-disable-next-line react-hooks/exhaustive-deps
+  [])
 
   const src = `${gatewayUrl}/ipfs/${cid}`
   const className = 'mw-100 mt3 bg-snow-muted pa2 br2 border-box'
@@ -79,7 +95,6 @@ const PreviewItem = ({ t, name, cid, size, type, availableGatewayUrl: gatewayUrl
       }
 
       if (!content) {
-        loadContent()
         return <ComponentLoader pastDelay />
       }
 
@@ -87,11 +102,19 @@ const PreviewItem = ({ t, name, cid, size, type, availableGatewayUrl: gatewayUrl
         return cantPreview
       }
 
-      return (
-        <pre className={`${className} overflow-auto monospace`} style={{ userSelect: 'all' }}>
+      return <>
+        <pre className={`${className} overflow-auto monospace`}>
           {content}
         </pre>
-      )
+        { hasMoreContent && <div className="w-100 flex items-center justify-center">
+          <Button onClick={ loadContent }>
+            { t('loadMore')}
+          </Button>
+          <Button className="mh2" onClick={ onDownload }>
+            { t('app:actions.download')}
+          </Button>
+        </div>}
+      </>
     }
   }
 }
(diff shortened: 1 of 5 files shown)


lidel (approved):

src/files/file-preview/FilePreview.js
```diff
@@ -27,8 +29,8 @@ const PreviewItem = ({ t, name, cid, size, type, availableGatewayUrl: gatewayUrl
   const [content, setContent] = useState(null)
 
   const loadContent = async () => {
-    const buf = await read()
-    setContent(buf.toString('utf-8'))
+    const buf = await first(await read())
```
lidel, line 32: Hm.. this will only show single block. To show full file you need to use [concat](https://www.npmjs.com/package/it-concat) or we need some UI to indicate truncation.

Example: try previewing text file that is bigger than a single block, it will get truncated:

- https://www.gutenberg.org/files/2600/2600-0.txt

>

lidel (changes requested):

lidel (changes requested):
nit: it is not possible to select text because it tries to drag&drop entire preview.
I think we should disable d&d for text previews, as its more usefulf for people to be able to select part of text and copy it.

src/files/file-preview/FilePreview.js
lidel: I'd vote for truncation (if file was bigger than 256 KB, add footer with link to full file), as its easier on resources, but will lean on @jessicaschilling to make a call what is the better fix here.

jessicaschilling: Totally agree with @lidel.

rafaelramalho19: Thank you, didn't thought of testing big text files 😢 


Changed to concat, should be 👌 now

rafaelramalho19:
> nit: it is not possible to select text because it tries to drag&drop entire preview.
> I think we should disable d&d for text previews, as its more usefulf for people to be able to select part of text and copy it.

Done! Also added `user-select: all`, so if the user double clicks the text/json, it copies everything :)

lidel (changes requested):
@rafaelramalho19  thanks! 

last nit: you switched to `concat`, which shows full file, but @jessicaschilling [agreed](https://github.com/ipfs-shipyard/ipfs-webui/pull/1646#discussion_r491065338) we should be keeping `first` to avoid loading unbound text files into browser.  Mind changing it back to `first` and if `buf.length > 256000`  show a footer at the end of preview informing user that preview was truncated?

lidel (changes requested):
@rafaelramalho19 something is still fishy: (1) clicking on "Load more" loaded entire thing instead of just one more chunk, and still displayes the footer. When i click on "Load more" again the footer disapears (2) footer is displayed for small text file, where is nothing more to load

Are you able to reproduce?

achingbrain:
> Did ipfs.cat start returning uint8arrays recently?

Yes - in `ipfs@0.50.x` and `ipfs-http-client@47.x.x`.  See: https://github.com/ipfs/js-ipfs/issues/3220 and https://blog.ipfs.io/2020-09-14-js-ipfs-0-50/

rafaelramalho19:
> @rafaelramalho19 something is still fishy: (1) clicking on "Load more" loaded entire thing instead of just one more chunk, and still displayes the footer. When i click on "Load more" again the footer disapears (2) footer is displayed for small text file, where is nothing more to load
> 
> Are you able to reproduce? (I'm on Firefox)

Yes @lidel , but the weird thing is that the buffer says that `done: false` and then on the second run just returns `value: undefined, done: true`

lidel:
@rafaelramalho19 I believe that is expected? When `value` is present then `done` is `false`, even if it is the last chunk, because iterator has no knowledge of what future will bring (its operates on chunk level without global state).

Some ideas: 
- Either account for that and execute additional `next` to see `done: true`
- ..or, instead of relying on `done`, you could check if currently read content length is equal  to the file size reported by `ipfs.files.stat` in `size` (size of raw data without metadata)

src/files/file-preview/FilePreview.js
```diff
@@ -18,20 +20,43 @@ const Preview = (props) => {
...
-  const src = `${gatewayUrl}/ipfs/${cid}`
+    setHasMoreContent(hasMore)
+  }, [buffer, content, read, size])
+
+  useEffect(() => {
+    loadContent()
+  }, // eslint-disable-next-line react-hooks/exhaustive-deps
+  [])
+
+  const src = `${gatewayUrl}/ipfs/${cid}?filename=${encodeURIComponent(name)}`
```
lidel, line 59: FYI I've added `filename` here, so its preserved when "Download" is clicked or when someone tries to download via browser's context menu.

lidel (approved):
@rafaelramalho19 works as expected and UX when opening huge text files is pretty nice now, thank you! :ok_hand:  

Feel free to merge this and prepare a patch release, so we can include it in https://github.com/ipfs-shipyard/ipfs-desktop/pull/1638 :ship:

achingbrain:
> Some ideas:
>
> * Either account for that and execute additional next to see done: true
> * ..or, instead of relying on additional next call, you could check if currently read content length is equal to the file size reported by ipfs.files.stat in size (size of raw data without metadata)

Please do the first and not the second.  If you do the second, the iterator will not know that it's not going to be invoked any more so will not perform any resource cleanup, tear down underlying streams etc, and could leak resources.

For example:

```javascript
async function * gen () {
  yield 'hello'

  console.info('i will do cleanup now')
}

async function main () {
  console.info('using low-level API')
  const iter = gen()
  const val = await iter.next()
  console.info('val', val)
  console.info('not calling .next again..')

  console.info('---')

  console.info('using for...await of')
  for await (const val of gen()) {
    console.info('val', val)
  }
  console.info('all done')
}

main()
```

Running this file results in:

```
using low-level API
val { value: 'hello', done: false }
not calling .next again..
---
using for...await of
val hello
i will do cleanup now
all done
```

Note how when using the low-level `.next()` API, control is not passed back to the `gen` function after the `yield` statement.

lidel:
Thanks @achingbrain, I've filled https://github.com/ipfs-shipyard/ipfs-webui/issues/1652 with an idea how to avoid the cleanup problem. 

----

@rafaelramalho19 I don't believe this should block `v2.11.2` release of webui – it has other urgent bugfixes that we should prioritise. Please tag  `v2.11.2` and bump webui to `v2.11.2` in https://github.com/ipfs-shipyard/ipfs-desktop/pull/1638 


(we will address #1652 in a later release of webui)

PR #1653, opened 2020-09-28
patch tiny-secp256k1

dignifiedquire:
Uses https://github.com/bitcoinjs/tiny-secp256k1/pull/49 to avoid build issues on macos


jessicaschilling:
Note: Keeping this as draft until we see what happens upstream with https://github.com/bitcoinjs/tiny-secp256k1/pull/49 and https://github.com/bcoin-org/bcrypto/issues/50.

PR #1655, opened 2020-09-30
feat: upload progress ui

Gozala:
> This pull request is a draft that contains more changes than it should, namely it fixes some of the storybook stories that I appeared broken, I will factor out those changes into separate PR.

This pull request adds stories corresponding to the mockups at https://www.figma.com/file/BQjzJcaKuNrNgeDnoZ664D/Web-UI-upload%2Fdownload-progress

image
image
image
image
image
image

@jessicaschilling I was not very successful in styling such that there will be no whitespace below progressbar, with what tachyons provides. If you could share some tips that would be greatly appreciated.

---

Now the bad news is (as @rafaelramalho19 mentioned it during the call) react is unable to catch up with all the updates that are happening to put it very gently. Some profiling showed an extreme high pressure on both CPU. On my macbook Air it's **0fps**, in fact total block time time (TBT) which should be kept <50ms was whooping 86412ms 😱. Put it differently while uploading 1gig file app was able to render only two frames.

image

I did some instrumentation and here are some early findings:

1. There are way too many re-renders happening. I am not sure yet where these updates coming from, but backlog of re-renders is so massive that chrome console is unable to store only the ones that happen when progress is at 0. Which suggests that those updates aren't caused by file add itself.

2. I can see state updates happening on file add. While state is updated to progress 32% react is still trying to process re-renders queue of updates where state is still 0%.
    > It is ridiculous that react still does re-render on every update instead of doing it on animation frames BTW, something that other libraries had been doing for a long time.

---

I am afraid this is not going to be easy, because scope of this problem is well beyond specific components. I'll need to investigate this further before I can tell if there are some shortcuts we could take, but it may well require significant changes to get a reasonable re-render loop.


jessicaschilling:
Boo, I suspected tachyons' weird height unit incrementing was going to foul that up. Feel free to explicitly give the progress bar itself a normal-measurement height.

Gozala:
Turns out all the progress updates start firing only after the request is complete, and that caused the flood of updates that react was unable to keep up with. I'll try to figure out why updates are starting to come in only after request is complete.

Gozala:
With the following changes applied:

- [ ] https://github.com/ipfs/js-ipfs-utils/pull/60
- [ ] https://github.com/ipfs/js-ipfs/pull/3310

I finally was able to resolve all the issues and get a progress bar working


Screen Recording 2020-10-02 at 12 03 15 PM 2020-10-02 12_10_09

src/files/FilesPage.js
```diff
@@ -36,7 +36,7 @@ const defaultState = {
   }
 }
 
-class FilesPage extends React.Component {
+export class FilesPage extends React.Component {
```
Gozala, line 39:
```suggestion
class FilesPage extends React.Component {
```

Gozala:
Cleaned up pull request and to remove unrelated changes and re-targeted it to storybook-fixes branch instead as those changes are needed to get the storybook stuff working here. Once that branch lands I'll change target back to master.

Please note that in order to test any of these changes you'd need to manually update `ipfs-http-client` dependency to include changes mentioned earlier. Unfortunately it does not appears like there is a way to do it with git urls because of the monorepo setup.

Gozala:
Fixed the styling so there is no whitespace below the progressbar and that we only display it if import is pending. I think it's ready to be reviewed even if it will take some time to get all the dependencies to propagate before it can land.

rafaelramalho19 (changes requested):

src/bundles/files/actions.js
```diff
@@ -248,21 +248,21 @@ const actions = () => ({
...
+    for await (const loaded of progress) {
+      // Reduce amount of progress updates by producing message by yielding
+      // only if progress changes by a precentage.
+      if (loaded > uploadState.progress) {
+        uploadState = { progress: loaded, entries }
+        yield uploadState
+      }
     }
 
     try {
```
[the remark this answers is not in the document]
Gozala: @rafaelramalho19 that is an issue, but I don't think it was introduced by changes here. Can we address it separately ? We could also probably ship fix to that sooner.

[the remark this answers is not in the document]
Gozala: Created a separate issue for that https://github.com/ipfs-shipyard/ipfs-webui/issues/1669

Gozala:
## Status update

I have run into problems in updating this pull. For whatever reason "react scripts" seems to modify [`isolatedModules`](https://www.typescriptlang.org/tsconfig#isolatedModules) to `true` which than causes issues with new ipfs-http-client (which we need to get progress events):

```
Failed to compile.

/Users/gozala/Projects/ipfs-webui/node_modules/ipfs-core-types/src/index.ts
TypeScript error in /Users/gozala/Projects/ipfs-webui/node_modules/ipfs-core-types/src/index.ts(5,3):
Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.  TS1205

    3 | 
    4 | export {
  > 5 |   RootAPI,
      |   ^
    6 | 
    7 |   AbortOptions,
    8 |   Await,
```

This is really annoying and I was unable to get it resolved yet. I hope we won't have to ship changes in http-client to fix this, although I am starting to run out of ideas. Even if I delete `tsconfig.json` `react scripts` seems to create one and produce above error.

Gozala:
Worked around the issue for now

Gozala:
Now we're running into https://github.com/Gozala/web-encoding/issues/1

Gozala:
@lidel not sure why this things are failing on CI, they don't on my device. I also end up disabling some tests because mocking `redux-bundle` does not seems to work with new react scripts, which were needed to address Gozala/web-encoding#1. I have spend several hours trying to figure out why jest mocking no longer works, but ultimately gave up. Do you by a chance have any clue what could be going wrong there, according to my reading of docs it should work but mocked `createSelector` seems to be returning `undefined` no matter how they're morked.

lidel:
I believe @rafaelramalho19  is experiencing the same or  similar issue with tests in #1713
I'll block some time to look into this, maybe I'll be lucky.

lidel:
Hm.. not sure if this is a good lead, but I see a weird behavior of `ipfs.ls(cid)` here
(this seems to cause e2e tests of Files screen to fail when i test with `npm run build && DEBUG=true npm run test:e2e`)

https://github.com/ipfs-shipyard/ipfs-webui/blob/cfe03a78398394b0352eeb92d481a6ed8c60c2ac/src/bundles/files/actions.js#L582-L587

When I open Files screen and the repo is empty  (`mfsRoot` being empty dir `QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn`)  `ipfs.ls(mfsRoot)`  returns `res` with a single object:

> 


Which looks like a bug. `ipfs.ls('QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn')` should return empty list. 
Instead, it returns... itself with `type` `unknown`  (instead of a directory) and `name` is missing, which is then causing error on line 587.

Will get some sleep and look at this tomorrow, but lmk if this rings any bells.

lidel:
Updating CI to the latest LTS Node (14.15.4) fixed one of tests:

- 12.x: Tests:       4 failed, 22 passed, 26 total
- 14.x: Tests:       3 failed, 23 passed, 26 total

lidel:
@Gozala any luck here?
I'd love to land this and then rebase #1713 and #1721 to include fixes from this PR.

Gozala:
I figured out why I was not able to reproduce the problems on my device - I was using yarn which is deduplicating dependencies, with npm I'm able to replicate the issue locally. I am also getting bunch of issues caused by multiple versions of same dependecy. Trying to get versions aligned now & hope it will get it all fixed.

Gozala:
Identified problem which is caused by the following change:
https://github.com/ipfs/js-ipfs/commit/f243dd1c37fcb9786d77d129cd9b238457d18a15#diff-451c33c54d61ab1f32e767b878ccd03a507ba13b852934fd40d32472df2f81f2R76-R78

Specifically `ipfs.ls(emptyDir)` contains CID of the `emptyDir` now, while prior to this change it returned no entries.

Gozala:
Submitted an issue https://github.com/ipfs/js-ipfs/issues/3566

Gozala:
I am trying to figure what is causing the remote-api.test failures, so far without much success. So far I have noticed that if I just have one test enabled it passes (regardless of the test). If I add bunch of alerts in the code things pass as well.

All this leads me believe that there is some timing problem possibly with with starting helper servers, but all this combined makes it really difficult.

lidel (approved):
I fixed e2e tests in https://github.com/ipfs-shipyard/ipfs-webui/pull/1655/commits/fc4b2dc152619b3c11c8d22ebd86340644886b8d and https://github.com/ipfs-shipyard/ipfs-webui/pull/1655/commits/39eaccc078d6e26dd7b5ad6e5502766d2a4091bc, but its only a testament we should move to playwright at some point.


CI is green, LGTM.

I tested import of 800MB ZIM file, and it looks ok.

src/bundles/peer-locations.test.js
```diff
@@ -30,7 +30,7 @@ jest.mock('hashlru', () => () => ({
 }))
 
 describe('reactPeerLocationsFetch', () => {
-  it('should declare its dependencies', () => {
+  it.skip('should declare its dependencies', async () => {
```
lidel, line 33: @Gozala I wonder if failures here are related to weird behavior I see on Peers screen.
It is unable to fetch geoip data, and requests are for the same CID:



Perhaps there is a regression somewhere  in  `ipfs.swarm.peers` or  `ipfs.object.*` API, similar to one we had with `ipfs.ls` in https://github.com/ipfs/js-ipfs/issues/3566 ?

lidel (changes requested):

src/bundles/peer-locations.test.js
lidel: Hm.. now issue is gone. I think it was just a fluke caused by my node looking for geoip root.

lidel:
@Gozala  is there anything left in this PR that you'd like to do before we merge?

I don't want to block this on those commented tests, perhaps we merge this to unblock rebase of Rafael's PRs and fill issue for fixing tests in separate PR?

Gozala:
> @Gozala is there anything left in this PR that you'd like to do before we merge?
>
> I don't want to block this on those commented tests, perhaps we merge this to unblock rebase of Rafael's PRs and fill issue for fixing tests in separate PR?

Can’t thank you enough for helping with CI issues! I think merging and doing everything as followups is the best way to go

lidel (approved):
Thank you @Gozala for all the time that went into this :heart: :rocket: 
(not just this PR, but even more work happened upstream in js-ipfs libs)

I'm merging this so we can rebase all the other PRs on top of the latest http-client and fixed CI/tests.

PR #1657, opened 2020-10-02
fix: broken file list stories

Gozala:
I think regression in the FileList stories was introduced by #1568, this pull request fixes it.


rafaelramalho19 (approved):
LGTM, thanks

PR #1658, opened 2020-10-03
feat(i18n): cache locale files

lidel:
This adds `i18next-localstorage-backend` which caches fetched translation files in `window.localStorage` for a week, which removes the need for seven requests for `.json` files.

ps. I also discovered that https://webui.ipfs.io/?lng=zh-CN works, so added it to README ;-)


jessicaschilling:
Will this introduce a need to repeatedly clear caches when hacking on text in master `en` locale?

lidel:
Good catch! Fixed in 09ba001b2bdff9c96ffb33540c184a4b5a2f53f2 –  disabled cache when developer server provided by `npm start` is used.

jessicaschilling (approved):

package.json
```diff
@@ -121,7 +123,7 @@
     "go-ipfs": "0.6.0",
     "http-proxy": "^1.18.1",
     "http-server": "^0.12.3",
-    "ipfs": "^0.48.1",
+    "ipfs": "0.50.2",
```
rafaelramalho19, line 126: is this intended in this PR? Just want to make sure we test everything so we don't problems from upgrading ipfs :)

rafaelramalho19 (approved):

rafaelramalho19:
Don't forget to update ipld-explorer-components translation array to match the one on this PR 😄

package.json
lidel: Good catch! js-ipfs is used only in e2e tests. 

Not related to this PR, but I noticed we were not using hard-coded version so fixed it and bumped to the latest version to ensure we are green with the latest version.

lidel:
@rafaelramalho19 added it to  https://github.com/ipfs-shipyard/ipld-explorer-components/pull/281 :ok_hand:

Issue #1659, opened 2020-10-05
Files → Import → File bug: "Imported NaN items"


lidel:
> small nit found in https://github.com/ipfs-shipyard/ipfs-desktop/issues/1661#issuecomment-703650125:

When "Import" → "File" button is used, and more than a single file is picked, the number of imported items is reported as `NaN` or `0`:
 
> ![image](https://user-images.githubusercontent.com/1507828/94972025-887e9e80-04c5-11eb-9cea-99e078603e8d.png)

Also: 

> Also, when dragging, It wrote that I imported 0 items:
![image](https://user-images.githubusercontent.com/8818223/94988972-87a34680-0571-11eb-9e90-2a18f1d79cf5.png)

**Update:** I confirmed this bug in Electron AND Firefox, so this is a generic bug.

PR #1660, opened 2020-10-06
fix: import / importing items count

Gozala:
This fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/1659, I would like to figure out how to automate tests for that issue, but I also don't want to block release so submitting this now & will try to figure out how to test in the followup.

According to @jessicaschilling we expect that when you drop or select `n` files for importing it would show "Importing n items" and once import is done show "Imported n items", however if you drop / select a file and a directory of n files it should show "Importing 2 items" and later "Imported 2 items" (Unless you had more items imported prior).

However code seemed to count number of import tasks for showing "importing n items" and then counting number of total files for showing "Imported n items", so neither was correct. Additionally non-existing `state` field was accessed which lead to `NaN`.

This change factors out general logic of grouping by directory name used in the view and uses it to count both pending and imported item numbers.

### DIFF
--- a/src/files/file-import-status/FileImportStatus.js
+++ b/src/files/file-import-status/FileImportStatus.js
@@ -13,37 +13,75 @@ import GlyphTick from '../../icons/GlyphTick'
 import GlyphCancel from '../../icons/GlyphCancel'
 import GlyphSmallCancel from '../../icons/GlyphSmallCancel'
 
-const File = (job, t) => {
-  const pathsByFolder = job.message.entries.reduce((prev, currentEntry) => {
-    const isFolder = currentEntry.path.includes('/')
-    if (!isFolder) {
-      return [...prev, currentEntry]
-    }
-
-    const baseFolder = currentEntry.path.split('/')[0]
-
-    const alreadyExistentBaseFolder = prev.find(previousPath => previousPath.path.startsWith(`${baseFolder}/`))
-
-    if (alreadyExistentBaseFolder) {
-      alreadyExistentBaseFolder.count = alreadyExistentBaseFolder.count + 1
-
-      return prev
-    }
-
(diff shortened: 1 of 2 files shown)


lidel (approved):
LGTM, thank you!

@rafaelramalho19 @jessicaschilling ok with merging this and shipping ipfs-webui v2.11.3?

@Gozala We have a basic end-to-end in https://github.com/ipfs-shipyard/ipfs-webui/blob/master/test/e2e/files.test.js which could be extended. But I agree, that should not block this PR, can be added later, unblocking ipfs-desktop release is the priority.

jessicaschilling (approved):
Tested in Chrome (Mac) - LGTM, thank you!

src/files/file-import-status/FileImportStatus.js
```diff
@@ -13,37 +13,75 @@ import GlyphTick from '../../icons/GlyphTick'
...
+  for (const entry of entries) {
+    const name = baseDirectoryOf(entry)
+    if (name == null) {
+      groupedEntries.set(entry.path, { type: 'file', ...entry })
+    } else {
+      // add `/` to avoid collision with file names.
+      const directory = groupedEntries.get(`${name}/`)
+      if (directory) {
+        directory.entries.push(entry)
+        directory.size += entry.size
```
rafaelramalho19, line 58: ❓ Wouldn't it be best to not mutate the Map item but instead set it again? Maybe something like

```javascript
if (directory) {
    groupedEntries.set(`${name}/`, {
          type: 'directory',
          size: directory.size + entry.size,
          path: name,
          entries: [...directory.entries, entry]
        })
}
```

rafaelramalho19 (approved):
LGTM, my suggestion is not mandatory, it's just a small improvement if you agree with it 😄

PR #1678, opened 2020-10-26
feat: add pinning to files page

rafaelramalho19:
Description:
➕  Updated/added pin icons
➕ Added new translations for pinning context menu + modal (please review them @jessicaschilling ?)
🐛  Fixed a bug where the file list with a lot of files (enough to have a scrollbar) wouldn't show the overlay correctly when dragging files to upload
💻  Refactored whole files page component to make it modern and faster ❤️


src/files/modals/pinning-modal/PinningModal.js
```diff
@@ -0,0 +1,70 @@
...
+
+    return setSelectedServices(selectedServices.filter(s => s !== key))
+  }
+  return (
+    <Modal {...props} className={className} onCancel={onCancel} >
+      <ModalBody title={t('pinningModal.title')}>
+        <div className="pinningModalContainer">
+          <button className="flex items-center pa1 hoverable-button" key={t('pinningModal.localNode')} onClick={() => selectService(t('pinningModal.localNode'))}>
+            <Checkbox className='pv3 pl3 pr1 flex-none' checked={selectedServices.includes(t('pinningModal.localNode'))}/>
+            <GlyphPin fill="teal" width={24} height={24} className="mr1 flex-shrink-0"/>
```
jessicaschilling, line 28:
```suggestion
            <GlyphPin fill="currentColor" width={32} height={32} className="mr1 aqua flex-shrink-0"/>
```
To match larger text, and aqua color from Files page row icons.

src/files/modals/pinning-modal/PinningModal.js
```diff
@@ -0,0 +1,70 @@
...
+            <p className="f6 w-100">{ t('pinningModal.localNode') }</p>
+          </button>
+          { availablePinningServices.map(({ icon, name }) => (
+            <button className="flex items-center pa1 hoverable-button" key={name} onClick={() => selectService(name)}>
+              <Checkbox className='pv3 pl3 pr1 flex-none' checked={selectedServices.includes(name)}/>
+              <img className="mr1" src={icon} alt='' width={24} height={24} style={{ objectFit: 'contain' }} />
```
jessicaschilling, line 34:
```suggestion
              <img className="mr1" src={icon} alt='' width={32} height={32} style={{ objectFit: 'contain' }} />
```
To match larger text.

jessicaschilling (changes requested):
Thank you! 😊  Added some inline comments for some small visual tweaks, but a few other notes, please:

1. Please add column header of `Pin status` per mockups; ideally this would be sortable but OK to omit for initial release if that's too painful.
image

2. Per mockups, please add `Set pinning` option in multi-select bottom bar:
image

3. I'm having trouble viewing the third-party pinning service icons, for what that's worth:
image

4. Are the checkboxes meant to work yet? (They don't for me.)

5. I think the icons in all the modals died 😢  Example: there's no trash icon in the Delete modal any more ...
image

src/files/modals/pinning-modal/PinningModal.js
rafaelramalho19: chore: update pinning modal pin icon

rafaelramalho19:
I fixed all the icons in all modals, I broke it by mistake. Thanks for noticing it 😄

jessicaschilling (changes requested):
This seems great (issues will probably crop up during implementation, but this seems like good "furniture" for now) 😊

One thing: On the Files page, no columns seem sortable now? Even if the Pin Status column isn't sortable, Name and Size should still be.

rafaelramalho19:
> This seems great (issues will probably crop up during implementation, but this seems like good "furniture" for now) 😊
> 
> One thing: On the Files page, no columns seem sortable now? Even if the Pin Status column isn't sortable, Name and Size should still be.

That's actually broken in master after the actions refactor (I believe), I'll fix it here too 😄

jessicaschilling (approved):
much good, very sort
thanks! 😊

lidel (commented):
lgtm! 

nit: I've found small UI bug in Firefox when you quicky scrolling a big directory.
Not sure if its Linux-specific, but try Browsing `QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm` and then scrolling it quickly in Firefox. I see this flashy behavior:
 
firefox-scroll-artifacts


@rafaelramalho19  / @jessicaschilling are you able to reproduce (on Firefox)?

rafaelramalho19:
@lidel Couldn't reproduce, is that something that happens to you in master too?

lidel (approved):
@rafaelramalho19 yup, its also in `master`, so not in scope of this PR.

LGTM!

public/locales/en/app.json
```diff
@@ -68,6 +69,7 @@
     "peers": "Peers",
     "pinNoun": "Pin",
     "pins": "Pins",
+    "pinStatus": "Pin Status",
```
lidel, line 72: nit: do we need this label at all? It looks bit noisy, and confusing when there are no pins.

jessicaschilling: @lidel I'd say yes - eventually this will be a sortable field.

lidel: Ah, indeed, sorting by pin status would be handy. 
Nvm then, we can refile label later (if needed).

PR #1694, opened 2020-11-23
feat: overwrite files modal

### LINKED ISSUE #1612: Add file override confirmation
Whenever a file is moved to another folder, we don't check for the existence of a file with the same name & extension. Leading to potential unintentional overrides.

**To Reproduce**
1. Have a `/dogs/oof.png` and `/dogs/breeds/oof.png`
2. Drag the first one into the `breeds` folder
3. No confirmation modal is being used

### LINKED ISSUE #1667: Renaming file will override existing one without warning
**Describe the bug**
When you `rename` a file to match a file that already exists with that name, it overrides the later without any warning.

**To Reproduce**
1. Go to files page
2. Have 2 files: `1.png` and `2.png`
3. Rename `2.png` to `1.png`
4. No warning that original `1.png` will be permanently lost with the rename

**Expected behavior**
@jessicaschilling thoughts on the expected behavior? Maybe open a modal that requires confirmation before actually replacing file?

rafaelramalho19:
This PR adds a feature that creates a modal for each file that has a conflicting name.
Fixes #1612
Closes  #1667


hacdias (approved):
LGTM (the code, didn't run it)!

lidel (changes requested):
@rafaelramalho19 mind rebasing on top of https://github.com/ipfs-shipyard/ipfs-webui/pull/1685 ? It should fix the CI.

src/files/files-list/FilesList.js
```diff
@@ -41,16 +37,29 @@ export const FilesList = ({
...
       const { filesPromise } = monitor.getItem()
 
-      addFiles(filesPromise, onAddFiles)
+      const filesToAdd = normalizeFiles((await filesPromise) || [])
+
+      const filePaths = files.map(f => f.path.replace('/files/', ''))
+      const existingFiles = filesToAdd.filter(({ path }) => filePaths.includes(path))
+
+      if (existingFiles.length) {
+        setExistingFiles(existingFiles)
```
Gozala, line 56: What if the same file is dropped twice, it would produce same CID but second drop would still add cid suffix (if user approves) even though there is no real conflict.

src/bundles/files/actions.js
```diff
@@ -288,15 +311,15 @@ const actions = () => ({
           try {
             await ipfs.files.cp(src, dst)
           } catch (err) {
-            throw Object.assign(new Error('Folder already exists.'), {
-              code: 'ERR_FOLDER_EXISTS'
-            })
+            // Item already exists, so let's remove it first.
```
Gozala, line 314: Assumption that either:

1. User agreed to overwrite a file
2. Or that `src` is suffixed and it's ok to delete existing path.

Make me really uncomfortable, despite the little change of them been invalid, because they will be invalid in following scenarios:

1. Directory entries may change by concurrent actor while import was in progress. So it is possible that conflicting path was added after drop and sure has not conceded to overwrite it.
2. Assumption that it is safe to delete `dst` seems to be based on the one that unless user chose to overwrite suffixed path will be for the same file so it's ok to remove it. Which is flawed because:
    1. There is a chance that directory contains unrelated file with name that by a chance matches suffixed file name.
    2. Collision of last 4 chars of hash is unlikely, but possible.

Gozala (changes requested):
I found it really difficult to follow the code paths. I think I have read through this at least 5 times to get my head around, while myself and react hooks should take some share of blame, I do think that fact that some decisions are made in view code (like identifying conflicting paths) and then actions deciding what to do based on added field is genuinely confusing. I think it would a lot easier to comprehend if all of the logic was encapsulated in `doFilesWrite` instead, e.g. on name conflicts conflicting file could be collected and maybe separate action could be dispatched to trigger dialog ? 

I think that would also help address following flaws, in the current implementation:

1. If same file is dropped twice, second drop will prompt user and on consent add same file with suffix. I would expect it should just be a noop, we could still show it as successful import (not to confuse or start unnecessary dialogs with user).
2. There are some assumptions about why error occurs in `doFilesWrite` in which case overwrite is performed. I believe those assumptions may not hold, even though probability is very low.

I would recommend handling name collisions in `doFilesWrite` and more specifically do them in `catch` block that way:

1. It is possible to just ignore imports for the same file.
2. Get a consent from the user after files are imported, which will reduces chance of tree changing during import & handle case when that occurs.
3. Properly handle case where unrelated file exists that conflicts with hash suffixed name.

src/files/files-list/FilesList.js
rafaelramalho19: It overrides the duplicated file (the one that has the appended CID), I believe it's fine since it required user interaction and awareness for that to happen?

src/bundles/files/actions.js
rafaelramalho19: Due to the current workflow it's assured that the user agreed to overwrite the file, however your point about the concurrent actor is relevant, even though it's unlikely.

Do you have any suggestions to fix that? I have no idea how to change the flow to fix the points you raised 🤔

Gozala: I wrote proposal on how to go about it in https://github.com/ipfs-shipyard/ipfs-webui/pull/1694#pullrequestreview-559244930

rafaelramalho19: In reply to your comment, I started by implementing this feature that way but I was blocked by the fact that there wasn't a way to handle the conflict inside `doFilesWrite`, if I recall correctly, I needed access to the files list beforehand, and the files depends only on the view that's currently being used (FilesList.js has different Files structure than App.js/File.js for example). Also, I needed access to the CID in order to perform the append, which forced me to change to the solution I have currently.


I don't believe your proposed changes would fix the edge-cases you pointed out, or am I wrong?

lidel (commented):
I tried to trigger modal via button and via drag&drop without any success:
- I had a directory with `guardian.jpg` and then imported different `guardian.jpg` (different CID) and there was no confirmation, the existing file got overwritten without any modal. 
- I also renamed `foo.jpg` to `guardian.jpg` (different CIDs as well) and overwrite happened without any modal.

@rafaelramalho19  Am I doing something wrong?

rafaelramalho19:
@lidel you just found a missing code path I forgot, the import button instead of the drag & drop! Thanks, I'll fix it 😓

src/bundles/files/actions.js
Gozala: > I needed access to the files list beforehand, and the files depends only on the view that's currently being used (FilesList.js has different Files structure than App.js/File.js for example).

I am not sure I follow all this, it might be worth going about this in sync channel. The way I understand it you have a `source` representing list of files to be written and a destination. So you can:

1. Go over each imported file (should have a CID and name).
2. Try to cp file to the dst.
  - If failed check whether file at destination is actually the same
    - If same skip and continue
    - if not the same push it into conflict list
  - if did not fail continue to the next file
  - At the end of the loop check if conflicting list is empty
    - If empty all is done
    - if not empty prompt user with conflict resolution strategy modal
      - If overwrite just go over the list and overwrite (rm + cp)
      - otherwise iterate over the conflicting files
         - derive suffixed name and try to cp
           - If error check if file at the destination is the same
              - If same move to next one
              - If not the same derive name with a a longer suffix and repeat until no error occurs (there is a little chance there to be a conflicting file named `name-fullcid` that is different, but chances are probably so tiny that we can error in that case)

In the end I believe you should have a result with all edge-cases addressed except when `file-cid` happens to exist. Significant difference here is that conflicts are not detected prior to import but after, as import can take some time increasing a chance that directory has changed.

rafaelramalho19: @Gozala A sync discussion sounds good, as it's a bit tricky to deal with this decision.

The issue that lead me to not follow your suggested flow (which I tried at the start), was due to the `cp` not failing when a file already existed. It simply returned success but didn't actually do anything.

So the current flow where I have to check for the existing files before copying it happened because of that.

Gozala: > The issue that lead me to not follow your suggested flow (which I tried at the start), was due to the `cp` not failing when a file already existed. It simply returned success but didn't actually do anything.

Hmm, you're right, [according to docs](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#notes-1) contents will just be replaced, which I find to be very strange

>
> If `from` has a single value and `to` exists and is a file, `from` must be a file and the contents of `to` will be replaced with the contents of `from` otherwise an error will be returned.
>

I'll try to figure out rational behind this API decision, it makes non racy `cp` impossible. That said, in the meantime it should be possible to create utility `cp` function that checks for path before performing actual `cp` (throwing if path exists). It still could have a race condition, but still time frame in which it can occur is far smaller than it is currently which can be as long as uploading a file would take.

rafaelramalho19: I guess so, I'll try to focus on changing this PR after pinning services

SELENITACOUSTICS:
---------- Forwarded message ---------
Šalje: SELENIT J.D.O.O Finance & Accounting <selenitjdoo@gmail.com>
Date: sri, 13. sij 2021. 13:47
**_Subject: IMPORTANT: [ipfs-shipyard/ipfs-webui] Regarding bandwidth api response (#1710)_**
To: ipfs-shipyard/ipfs-webui <reply+AL43F45MNRRADWUOLCRIN256BKP7VEVBNHHC5SKGOY@reply.github.com>

This business account of my small firm which I use trough gmail for corespod with some clients and that is all...although I did make long time now a github account with this adress mail but I do not use that lack of time and understanding...So obviously..this is hacked account...and some one with knoledge in correspondenting with you and your help ...I recently saw bunch of mails from your part but did not think much. But. .couse I do not know wtf is all about I think now tgat it is some third partie included..and I do not have clue at all.
Deep apologiez and please I am so so sorry 
 for every bit inconvenience
L&L
Ana Burazer Siriscevic - Selenit j.d.o.o.

SELENITACOUSTICS:
---------- Forwarded message ---------
Šalje: *SELENIT J.D.O.O Finance & Accounting* <selenitjdoo@gmail.com>
Date: sri, 13. sij 2021. 13:47
Subject: IMPORTANT: [ipfs-shipyard/ipfs-webui] Regarding bandwidth api
response (#1710)
To: ipfs-shipyard/ipfs-webui <
reply@reply.github.com>


This business account of my small firm which I use trough gmail for
corespod with some clients and that is all...although I did make long time
now a github account with this adress mail but I do not use that lack of
time and understanding...So obviously..this is hacked account...and some
one with knoledge in correspondenting with you and your help ...I recently
saw bunch of mails from your part but did not think much. But. .couse I do
not know wtf is all about I think now tgat it is some third partie
included..and I do not have clue at all.
Deep apologiez and please I am so so sorry
 for every bit inconvenience

Love and light

Ana Burazer Siriscevic - Selenit j.d.o.o.

sri, 13. sij 2021. u 15:51 Rafael Ramalho <notifications@github.com>
napisao je:

SELENITACOUSTICS:
---------- Forwarded message ---------
Šalje: SELENIT J.D.O.O Finance & Accounting <selenitjdoo@gmail.com>
Date: sri, 13. sij 2021. 13:47
Subject: IMPORTANT: [ipfs-shipyard/ipfs-webui] Regarding bandwidth api response (#1710)
To: ipfs-shipyard/ipfs-webui <reply+AL43F45MNRRADWUOLCRIN256BKP7VEVBNHHC5SKGOY@reply.github.com>


This business account of my small firm which I use trough gmail for corespod with some clients and that is all...although I did make long time now a github account with this adress mail but I do not use that lack of time and understanding...So obviously..this is hacked account...and some one with knoledge in correspondenting with you and your help ...I recently saw bunch of mails from your part but did not think much. But. .couse I do not know wtf is all about I think now tgat it is some third partie included..and I do not have clue at all.
Deep apologiez and please I am so so sorry 
 for every bit inconvenience

Love and light

Ana Burazer Siriscevic - Selenit j.d.o.o.

src/bundles/files/actions.js
SELENITACOUSTICS: ---------- Forwarded message ---------
Šalje: SELENIT J.D.O.O Finance & Accounting <selenitjdoo@gmail.com>
Date: sri, 13. sij 2021. 13:47
Subject: IMPORTANT: [ipfs-shipyard/ipfs-webui] Regarding bandwidth api response (#1710)
To: ipfs-shipyard/ipfs-webui <reply+AL43F45MNRRADWUOLCRIN256BKP7VEVBNHHC5SKGOY@reply.github.com>


This business account of my small firm which I use trough gmail for corespod with some clients and that is all...although I did make long time now a github account with this adress mail but I do not use that lack of time and understanding...So obviously..this is hacked account...and some one with knoledge in correspondenting with you and your help ...I recently saw bunch of mails from your part but did not think much. But. .couse I do not know wtf is all about I think now tgat it is some third partie included..and I do not have clue at all.
Deep apologiez and please I am so so sorry 
 for every bit inconvenience

Love and light

Ana Burazer Siriscevic - Selenit j.d.o.o.

lidel (changes requested):
We merged latest http-client with https://github.com/ipfs-shipyard/ipfs-webui/pull/1655 so connect-deps should no longer be necessary.

@rafaelramalho19 mind rebasing this PR?

PR #1709, opened 2021-01-11
fix(line-wrap): Fixed tooltips of file names

melegiul:
### Summary
This PR fixes the problem of displaying long file names when hovering the file.
- Set break-word property to tooltip.
- Changed overflow of rows to visible.
- Set tooltip width to not exceed the menus width.

Fixes: #1464

### Screenshot
longFileName

### DIFF
--- a/src/components/tooltip/Tooltip.js
+++ b/src/components/tooltip/Tooltip.js
@@ -48,8 +48,10 @@ export default class Tool extends React.Component {
         <div style={{
           bottom: '-10px',
           left: '50%',
-          transform: 'translate(-50%, 100%)'
-        }} className={`nowrap white z-max bg-navy-muted br2 pa1 f6 absolute ${(show && overflow) ? 'db' : 'dn'}`}>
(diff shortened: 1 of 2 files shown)


jessicaschilling:
Hi @melegiul -- thank you so much for fixing something that's been one of my own pet peeves! ;)  

Confirmed to work as intended on Mac (Big Sur) in latest Chrome, Brave, Safari, Firefox:

Long file name (Chrome):
image

CID on narrow screen (Chrome):
image

Note that Safari has always displayed these differently:
image

Also note that this does **not** address the issue with hover text clipping on the last item in a files list due to the react-virtualized grid acting as a boundary, but that's not within the scope of this issue.
image

I'm marking this PR as LGTM, but asking @lidel to have a quick look before we merge. Thanks again!

jessicaschilling (approved):
LGTM - just waiting on @lidel OK before merging. Thanks!

lidel (approved):
LGTM, thank you @melegiul :heart:

PR #1718, opened 2021-01-30
fix: trim input of "Import from IPFS"

FledgeXu:
# Description
This PR adds a default trim to `TextInputModal` and it should improve kind of user experiences.
Like below:
`before`:
image

`after`:
image


jessicaschilling:
Thanks! We'll look in more detail on Monday. 😊

jessicaschilling:
(Edit - the team is deferred this week but this is on our radar for review as soon as possible!)

lidel (approved):

PR #1719, opened 2021-02-03
consistently use Du instead of Sie

marten-seemann:
We probably should also switch to using "du" instead of "Du". Since the orthography reform of 1996 it's not supposed to be capitalized any more ;)


jessicaschilling:
@marten-seemann Sorry for any duplication of effort, but we use Transifex for internationalization - can you please make edits there? https://www.transifex.com/ipfs/ipfs-webui/translate/

Issue #1502, opened 2020-05-28
Pinning integration: Make Files screen functional


jessicaschilling:
_Note: This issue is part of a larger [pinning service integration epic](https://github.com/ipfs/ipfs-gui/issues/91) undertaken spring/summer 2020._

### In brief
Make the UI elements for the Files screen that were stubbed out in #1500 functional by integrating with the to-be-built IPFS Pinning API.

**⚠️ IMPORTANT:** Work on this issue is blocked by issues #1500 and #1504 in addition to completion of the API itself.

lidel:
Closed as part of https://github.com/ipfs/ipfs-gui/issues/91

PR #1713, opened 2021-01-21
feat: remote pinning settings page

rafaelramalho19:
## Description

- Unstubs UI Elements to use Remote Pinning API (https://github.com/ipfs/js-ipfs/pull/3293)
- The CI is probably going to fail until we have a go-ipfs@8+ release and bump this PR to use it

## Testing

To test this locally, follow these steps: 
- Checkout js-ipfs into local dir and switch to @Gozala branch (feat/pin-remote)
- Go to js-ipfs/package/ipfs-http-client and try to build it via `npm ci`.
- Go back to ipfs-webui dir, and execute `connect-deps link /path/to/js-ipfs/packages/ipfs-http-client` and then `connect-deps connect`
if successful, you should be able to run `npm i && npm start` and have getIpfs().pin.remote

PR #1721, opened 2021-02-09
feat: remote pins on files page

rafaelramalho19:
Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/1502


lidel (changes requested):
Not sure if issues I had were caused by using different version of `ipfs-http-client` with `connect-deps`, but here is quick feedback:

- error handling seems to be missing. when I unpinned a file from one of services that are broken on purpose, I got [this page crash](https://user-images.githubusercontent.com/157609/107532080-c8653900-6bbd-11eb-81d2-3cf37991ba44.png)

- assuming http-client issues caused below ones, but mentioning them for completeness:
   - I was not able to get cloud pin to show on Files screen
   - I have multiple services defined with Pinata endpoint, but I don't see them in "Set pinning" modal, I only see a single Pinata entry

lidel (changes requested):
We merged latest http-client with https://github.com/ipfs-shipyard/ipfs-webui/pull/1655 so connect-deps should no longer be necessary.

@rafaelramalho19 mind rebasing this PR?

lidel:
I've merged fixes from main branch but there is some remaining work here:

- [ ] resolve conflicts 
- [ ] rebase after #1535  is merged 
- [ ] when I have multiple Pinata services the "Set pinning" modal shows "Pinata"  only once
- [ ] when one of services is offline entire Files screen crashes ([screenshot]((https://user-images.githubusercontent.com/157609/109990203-779eb700-7d09-11eb-9b76-cb955df39701.png))

rafaelramalho19:
@lidel > when I have multiple Pinata services the "Set pinning" modal shows "Pinata" only once (we need to merge #1713 first and rebase)

This was actually never done apparently, ıt was so long ago I can't remember but I guess I didn't finish it, so I done it today 😄 

I also replicated the colors from the settings page here
image

rafaelramalho19:
> when one of services is offline entire Files screen crashes 

- What's the best way to reliably test this? Should I make an e2e test and try to reproduce this behavior there?
- It crashes when in dev mode right? If you "close the debugging modal" it should continue as normal right?


>  rebase after #1535 is merged

What do you mean by this? That PR was merged a long time ago 🤔

lidel:
>   when one of services is offline entire Files screen crashes

This seems to be fixed in the latest version: the screen no longer crash.
Instead, when I pin a file  to a service which is offline, no error is displayed and the file gets cloud pin badge, buf if I refresh the page the badge disappears. 

> What's the best way to reliably test this? Should I make an e2e test and try to reproduce this behavior there?

For dev just add a service with invalid endpoint or access token (so every request always fail).

I think the only way to do automatic tests is to  add `test/e2e/remote-pins.test.js` with tests for adding/removing service and pinning files using invalid service and a valid one.

For valid one you can use  [js-mock-ipfs-pinning-service](https://github.com/ipfs-shipyard/js-mock-ipfs-pinning-service):
- runs in memory, so should be acceptable for CI
- you can force it to set status to 'pinned' by using  CID with identity multihash like `bafkqaaa`  (you can import it from IPFS via Files→Import→Import from IPFS → `/ipfs/bafkqaaa`)


> That PR was merged a long time ago 

My bad, meant https://github.com/ipfs-shipyard/ipfs-webui/pull/1713 :)

PR #1644, opened 2020-09-17
feat: delete-modal pin guidance, improved language

jessicaschilling:
### Summary
This PR prepares for pinning service continuity and improves overall UX of the Files screen "Delete" modal as follows:
- Replaces "delete" language with "remove" and clarifies re garbage collection and possible presence on other nodes
- Adds checkbox option for removing an item's local pin (if one exits)
-  Adds checkbox option for unpinning from all pinning services
- _Note: Doesn't add a checkbox for GC per concerns in https://github.com/ipfs-shipyard/ipfs-webui/issues/1634#issuecomment-693672213_

Closes https://github.com/ipfs-shipyard/ipfs-webui/issues/1634.
Closes #1258 

### Done
- [x] Update modal title, text, buttons for all variants (item, folder, file)
- [x] Add (but don’t wire up) checkboxes
- [x] Update context-menu item title
- [x] Update multi-select menu item title
- [x] Replace `actions.delete` with `actions.remove` in `app.json`; make sure this doesn’t cause any other issues
- [x] Rename `deleteModal`/`DeleteModal` to `removeModal`/`RemoveModal` and update file names accordingly
- [x] Replace “delete” with “remove” throughout, both in UI messaging and in codebase where it’s more apropos (NOTE: I veered on the side of conservative here, @lidel may have other ideas)
- [x] Update modal styling per @ericronne suggestions (note: standardized modals on Files page for visual consistency, but didn’t go further since that’ll be covered in https://github.com/ipfs-shipyard/ipfs-webui/issues/1643)
- [x] Fixed unrelated file list refresh issues: close #996, close #1740 and close #1742

### To do
- [x] Wire up checkboxes (first box should be checked by default)
- [x] Address  https://github.com/ipfs-shipyard/ipfs-webui/issues/1258 while we're here by adding the name of the (file, folder) to be deleted if only one item is being deleted: _Are you sure you want to remove "catphoto.gif"?_

### Open question
Should we change CLI tutor mode guidance in any way to match?

### Screenshot
(Note this doesn't include the language _Are you sure you want to remove "catphoto.gif"?_ - can update screenshot if desired)
image

PR #1741, opened 2021-03-30
fix: batch caching of remote pin lookups

rafaelramalho19:
Still WIP

- Refactored FilesList to functional components (speeds up the application loading & responsiveness, as well as allowing for this new optimization to work)
- Refactored all loadable components to use the new library (@loadable/component) so it doesn't throw errors in the console anymore due to latest version of react


lidel (changes requested):
@rafaelramalho19 I've played a bit with this PR, and it mostly worked as expected: – no longer fetching all remote pins  :tada: :))

The only missing part was avoiding sending check for CIDs that were already probed, 
so I've added some optimizations in https://github.com/ipfs/ipfs-webui/pull/1741/commits/9792b476f68b9f648c53bbc8de5617ec64742787:
  - It  adds new cache for "not-pins" per each remote service, so we can remember remote pin state ("pinned" or "not pinned") of each item and avoid sending an unnecessary request every time we see the same CID anywhere in Files.

If you could take a look at remaining TODO here: 

- [ ] On the initial load, Files screen does not run the check for MFS root. One needs to go to a subdirectory to trigger remote pin checks, and then if one goes back to MFS root, the check is finally executed.  We are probably missing `doFetchRemotePins(<MFS root files>)` somehwere, or it gets skipped because some state is not ready yet.
- [ ] `ipld-explorer-components` still requires `react-loadable` so either you need to add it back for now, or refactor that lib as well.

rafaelramalho19:
> On the initial load, Files screen does not run the check for MFS root. One needs to go to a subdirectory to trigger remote pin checks, and then if one goes back to MFS root, the check is finally executed. We are probably missing doFetchRemotePins(<MFS root files>) somehwere, or it gets skipped because some state is not ready yet.

Is this still happening? Can't replicate 🤔

lidel:
I am able to reproduce this in both Firefox and Brave. 
There is no API call to `/api/v0/pin/remote/ls` the  very first time Files screen is opened. 
Do you have any remote pins in the root of Files?

lidel (approved):
Ok, managed to fix the problem. It manifested on my box because I use Firefox and have a loooot files in the root directory, so initial remote pin check was slow enough to trigger performance issues and race condition due to unnecessary rendering.

https://github.com/ipfs/ipfs-webui/pull/1741/commits/6207f68caf3ed27b0edc58b52d0ee29311287afd  simplifies the way we pass information about remote pins to cache by batching all pin updates into a single event. 
This way we avoid delays in UI caused by unnecessary re-renders.

Additional fixes:
- requests are only sent to services that are confirmed to be online
- CIDs to check are deduplicated (in case same item is present multiple times)
- remote pins are created with `background:true` 
  - we don't have UI for "ongoing pin" nor "failed pin" anyway
  - above can be added in the future, but it is  best-effort for now
- initial load of Files shows remote pins, but we fetch remote pins only if at least one service is online

@rafaelramalho19 looks ok to merge on my end, but  please take a look and give :+1:  if no concerns.

lidel:
Ugh, I've merged  #1644   and now we need to resolve conflicts.
Merging master to this branch produced some weird errors around react, so rebase may be a safer way. 
Will look into it tomorrow.

lidel:
@rafaelramalho19 resolved conflicts – give this a quick look and  :+1:  / :-1:  in a spare moment.

src/bundles/pinning.js
```diff
@@ -140,33 +184,36 @@ const pinningBundle = {
...
-          dispatch({
-            type: 'ADD_REMOTE_PIN',
-            payload: {
-              ...pin,
-              id: `${service.name}:${pin.cid}`
-            }
-          })
-          await ipfs.pin.remote.add(cid, { service: service.name, name })
+          adds.push({ id, ...pin })
+          // TODO: remove background:true and add pin job to queue.
```
rafaelramalho19, line 202: Maybe let's add a github issue with this?

lidel: Yes, filled https://github.com/ipfs/ipfs-webui/issues/1751 and https://github.com/ipfs/ipfs-webui/issues/1752
(we should add error handling first, tracking progress will be more complex)

PR #1753, opened 2021-04-01
fix: files page size

rafaelramalho19:
Fixed regression in files size
Closes https://github.com/ipfs/ipfs-webui/issues/1738


lidel (approved):
:heart:

PR #1761, opened 2021-04-12
wip: fetching parent dir cids to check if pinned

rafaelramalho19:
Aims to close https://github.com/ipfs/ipfs-webui/issues/1736

## WIP PR

- Missing function `doCheckIfMultipleArePinned` in `bundles/ipfs-provider.js` with a strategy to fetch the parent CIDs (we had performance concerns about doing X HTTP requests depending on the number of parent dirs)
- Missing variable and UI style for the `setParentLocallyPinned` state

### DIFF
--- a/src/bundles/files/actions.js
+++ b/src/bundles/files/actions.js
@@ -27,6 +27,7 @@ import { IGNORED_FILES, ACTIONS } from './consts'
  * @property {string} path
  * @property {boolean} pinned
  * @property {boolean|void} isParent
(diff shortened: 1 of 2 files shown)


lidel:
I'm parking this one for now. While nice to have, lack of indirect pin indicators is not the end of the world.
Note for later: we should try to tackle this together with "tracking pin progress" (https://github.com/ipfs/ipfs-webui/issues/1752)

PR #1768, opened 2021-04-26
fix: do not send consent removal event

olizilla:
- Saying no to analytics explicitly calls `Countly.opt_out`. Without this removing all consents triggers a "consent removed" event to be tracked which is not what we want.
- Removing all individual consents explicitly calls `Countly.opt_out`
- Granting consent for 1 or more analytic types opts back in to analytics by calling `Countly.opt_in`.
- Saying yes to analytics explicitly calls `Countly.opt_in`, for completeness.

ref: https://github.com/ipfs/ipfs-webui/issues/1041#issuecomment-825770656

License: MIT

### DIFF
--- a/src/bundles/analytics.js
+++ b/src/bundles/analytics.js
@@ -203,6 +203,7 @@ const actions = {
    * @returns {function(Context):void}
    */
   doDisableAnalytics: () => ({ dispatch, store }) => {
+    root.Countly.opt_out()
     removeConsent(consentGroups.all, store)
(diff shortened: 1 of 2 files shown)


olizilla:
I'm updating the test to verify this change

olizilla:
I need to figure out if this plays nicely with ipfs-desktop

lidel (approved):
LGTM, confirmed the countly is no longer pinged. Thank you @olizilla :heart: 

If you want to test against Desktop do it from PR branch of https://github.com/ipfs/ipfs-desktop/pull/1807  to ensure the latest electron is used.

PR #1772, opened 2021-04-28
fix: work around dot in pin service name

### LINKED ISSUE #1770: Pinning problems when a service name includes dot
## Problem in GUI

Found in v0.12.1 while testing https://nft.storage  – having "." in name means bad time in both GUI and CLI.

> 

Quick fix for GUI is to replace "." with "-" or "_" when Service is added  from GUI, but  that won't solve the problem if  one added it via CLI:

## Problem in CLI

```console
$ ipfs pin remote service add nft.storage https://nft.storage/api [secret]
$ ipfs config Pinning.RemoteServices.nft.storage.Policies.MFS.Enable
Error: failed to get config value: "Pinning.RemoteServices key has no attributes"
```

## How to fix this?

@alanshaw @aschmahmann given that we use "nft.storage" in `pin remote`  examples on https://nft.storage, we should address this before the problem escalates when people start using it :sweat_smile: 

I tried explicit percent encoding `%2E` but it does not work  (we would have to modify both webui and go-ipfs to allow that). 

Is there a preexisting way to pass `.` correctly?

If not, we need to either
- (a)  tweak go-ipfs to percent-decode service name to make `%2E` work,
- (b) ban `.` from service names and replace with `-` in docs
- (c) ?? (is there other way to resolve this?)

lidel:
This is a temporary workaround that maintains most of the functionality while being compatible with `pin remote` and `config` commands in go-ipfs 0.8.0:

- when service is added via webui, we replace `.` with `_` so the issue (#1770) never surfaces
- if service with `.` in name was already added or is added via CLI, we disable button for enabling auto upload of MFS, because it is not possible to make it work in go-ipfs 0.8.0, but everything else works as expected (manual pinning to "nft.storage" works fine)

Closes  #1770 (but we will have a separate fix in go-ipfs to support `.` in key names)

@alanshaw just for completeness, would it be ok for you to change `nft.storage` to `nft-storage` on https://nft.storage docs about `pin remote service add`?

### DIFF
--- a/src/bundles/pinning.js
+++ b/src/bundles/pinning.js
@@ -20,7 +20,12 @@ const parseService = async (service, remoteServiceTemplates, ipfs) => {
 }
 
 const mfsPolicyEnableFlag = memoize(async (serviceName, ipfs) => {
-  return await ipfs.config.get(`Pinning.RemoteServices.${serviceName}.Policies.MFS.Enable`)
+  try {
+    return await ipfs.config.get(`Pinning.RemoteServices.${serviceName}.Policies.MFS.Enable`)
+  } catch (e) {
+    console.error(`unexpected config.get error for "${serviceName}"`, e)
+    return false
+  }
 }, { maxAge: 3000 })
 
 const uniqueCidBatches = (arrayOfCids, size) => {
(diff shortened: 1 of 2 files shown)


alanshaw:
✅ https://github.com/ipfs-shipyard/nft.storage/commit/deff12ea7504dff463fdeb1b58c39a761fd254ef

olizilla:
changed on the spec too https://github.com/ipfs/pinning-services-api-spec/pull/76

lidel:
Thanks! :heart:

Issue #1737, opened 2021-03-25
Fix  end-to-end tests under js-ipfs


lidel:
> Extracted from https://github.com/ipfs/ipfs-webui/pull/1713

Right now e2e tests do not pass against js-ipfs:

```
npm ci && npm run build && E2E_IPFSD_TYPE=js npm run test:e2e
```

Note: we are unable to ship webui with js-ipfs without fixing this. 
In theory, it is ok to ship release that only works with go-ipfs and fix this later, using tribute rotation, but would be way better to ship release to both JS and GO.

lidel:
For reference:

```
(node:22646) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
 FAIL  test/e2e/navigation.test.js
  Navigation menu
    ✓ should work for Status page (152 ms)
    ✓ should work for Files page (246 ms)
    ✕ should work for Explore page (16 ms)
    ✓ should work for Peers page (373 ms)
    ✓ should work for Settings page (434 ms)

  ● Navigation menu › should work for Explore page





  console.error
    Unhandled error

      at process.uncaught (../../node_modules/jest-config/node_modules/jest-jasmine2/build/jasmine/Env.js:248:21)
      at processEmit [as emit] (../../node_modules/signal-exit/index.js:161:32)
      at handleError (../../node_modules/jest-environment-puppeteer/lib/PuppeteerEnvironment.js:17:11)
      at ../../node_modules/puppeteer/vendor/mitt/src/index.ts:81:75
          at Array.map (<anonymous>)
      at Object.emit (../../node_modules/puppeteer/vendor/mitt/src/index.ts:81:56)

  console.error
    

      at process.uncaught (../../node_modules/jest-config/node_modules/jest-jasmine2/build/jasmine/Env.js:249:21)
      at processEmit [as emit] (../../node_modules/signal-exit/index.js:161:32)
      at handleError (../../node_modules/jest-environment-puppeteer/lib/PuppeteerEnvironment.js:17:11)
      at ../../node_modules/puppeteer/vendor/mitt/src/index.ts:81:75
          at Array.map (<anonymous>)
      at Object.emit (../../node_modules/puppeteer/vendor/mitt/src/index.ts:81:56)

 PASS  test/e2e/files.test.js
  Files screen
    ✓ should have the active Add menu (245 ms)
[...]

PR #1789, opened 2021-05-19
fix: fix download files feature

### LINKED ISSUE #1783: Unable to download files
**Describe the bug**
Nothing happens when trying to download files or directories.

**To Reproduce**
1. Go to 'Files'
2. Click on 'Import'
3. Click on 'New Folder'
4. Type 'foo'
5. Click on 'Create'
6. Click on the three dots to the right of "foo"
7. Click on 'Download'
8. Nothing happens

**Expected behavior**
I expect to be able to download files.

 - OS: macOS 10.15.7
 - Browser Safari 14.1
 - Version Web UI v2.12.2

**Additional context**
Downloading works with v2.11.4 but not with v2.12.0+.

hugomrdias:
Abort download was being called right after starting to fetch data, because if useState receives a function that function is called to get the final value.
To fix the issue we just wrap the abort function with another function.

closes #1783

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16834,28 +16834,28 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "",
+          "resolved": false,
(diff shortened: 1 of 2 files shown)


lidel (approved):
Thank you!
e2e for js-ipfs is unrelated – we track it in #1737

Luflosi:
Thank you!

PR #1790, opened 2021-05-19
fix: fix modal open when removing files

### LINKED ISSUE #1785: Removing more than one file via bottom drawer's  "Remove" does not work
> 

- bottom drawer Remove action is broken, produces error
- removing via context menu works fine

hugomrdias:
closes #1785


lidel (approved):
LGTM.

(failure of e2e for js-ipfs is unrelated – we track it in #1737)

PR #1802, opened 2021-06-10
feat: replace puppeteer with playwright for e2e tests

### LINKED ISSUE #1164: Improve Test Suite

yusefnapora:
I started pulling at this thread yesterday, since I hit some issues with puppeteer (possibly related to #1737).

This swaps out puppeteer for playwright. I haven't tested to see if that resolves the issue with js-ipfs in #1737, since I just got this working a few minutes ago and I should have switched gears a couple hours ago :) There's probably some stray console.logs and lint errors strewn about; I'll fix those up tomorrow.

cc @lidel  closes  #1164


yusefnapora:
This seems odd:

```
Error: Jest: Got error running globalSetup - /home/runner/work/ipfs-webui/ipfs-webui/test/e2e/setup/global-init.js, reason: go-ipfs binary not found, it may not be installed or an error may have occured during installation
    at Object.module.exports.path (/home/runner/work/ipfs-webui/ipfs-webui/node_modules/go-ipfs/src/index.js:18:9)
    at globalSetup (/home/runner/work/ipfs-webui/ipfs-webui/test/e2e/setup/global-init.js:48:37)
    at async /home/runner/work/ipfs-webui/ipfs-webui/node_modules/@jest/core/build/runGlobalHook.js:135:13
    at async waitForPromiseWithCleanup (/home/runner/work/ipfs-webui/ipfs-webui/node_modules/@jest/core/node_modules/@jest/transform/build/ScriptTransformer.js:207:5)
    at async /home/runner/work/ipfs-webui/ipfs-webui/node_modules/@jest/core/build/runGlobalHook.js:126:9
    at async pEachSeries (/home/runner/work/ipfs-webui/ipfs-webui/node_modules/p-each-series/index.js:8:23)
    at async _default (/home/runner/work/ipfs-webui/ipfs-webui/node_modules/@jest/core/build/runGlobalHook.js:110:5)
    at async runJest (/home/runner/work/ipfs-webui/ipfs-webui/node_modules/@jest/core/build/runJest.js:352:5)
    at async _run10000 (/home/runner/work/ipfs-webui/ipfs-webui/node_modules/@jest/core/build/cli/index.js:408:7)
    at async runCLI (/home/runner/work/ipfs-webui/ipfs-webui/node_modules/@jest/core/build/cli/index.js:261:3)
```

It says it can't find the `go-ipfs` binary... I don't think I changed anything related to that.

lidel (changes requested):
@yusefnapora thanks for switching to playwright ! :heart: 

### On go-ipfs binary on CI

I  wonder if  the go-ipfs binary problem is related to https://github.com/ipfs/npm-go-ipfs/issues/36
Could be something else because the npm version did not change: same version was able to find go-ipfs binary just fine 2 days ago:
https://github.com/ipfs/ipfs-webui/runs/2787746156?check_suite_focus=true#step:3:11

## On js-ipfs

I see running `E2E_IPFSD_TYPE=js npm run test:e2e`  still fails, but now with playwright we should have more confidence when fixing it.

yusefnapora:
It seems like the go-ipfs issue was happening because I was using npm v7 locally, which generated a v2 lockfile. This apparently causes the postinstall script in go-ipfs to not run on ci. I re-generated the lockfile using npm v6, but then I started getting errors like this:

```
TypeError: virtualMocks.get is not a function

      at Resolver._getVirtualMockPath (../../node_modules/jest-resolve/build/index.js:447:25)
```

It seems there are some incompatibilities between `jest-playwright-preset` and jest v27 (see https://github.com/playwright-community/jest-playwright/issues/665)

Oddly, the jest issue doesn't show up using npm v7 - it's only broken if you install using npm v6. So there must be some significant change in dependency resolution logic between v6 and v7 that's causing it to break. Fun stuff.

I just pinned jest to v26, which seems to fix the issue. It's not ideal to be stuck on v26 though, so it may be worth investigating further to see what's actually causing the problem. Or maybe the `jest-playwright-preset` maintainers will sort it out for us before we need anything from v27 😄

yusefnapora:
haha, okay, using lockfile v1 and jest v26 fixed the e2e tests (with go-ipfs anyway). But now the unit tests are failing with `ReferenceError: TextDecoder is not defined` 🤕 I now get the same on my machine when running with node v15 also...

I found https://github.com/ipfs/js-ipfs/issues/3620, but it seems odd to be getting this error with node v15, since `TextDecoder` should definitely exist...

yusefnapora:
Well, I squashed all the commits, then split out the dependency changes so I could revert `package.json` and the lockfile back to the state they're in on `main`. Then I re-applied all the dependency changes by `npm installing` things one-by-one, and this time the unit tests pass without any `TextDecoder` errors... Not sure what was actually causing that issue, but it seems to be cleared up - I'm guessing it was some weird consequence of switching between npm versions.

At any rate, the go e2e tests and unit tests are back in business, so I'm going to quit while I'm ahead 😆

lidel (approved):
Thank you @yusefnapora  this should be ok to merge.
(I did cosmetic tweaks to `waitForText` in 387f8c6 just to make CI logs more intuitive)
 
 We will tackle js-ipfs in https://github.com/ipfs/ipfs-webui/issues/1737 and evaluate running against other vendors in #1804

PR #1806, opened 2021-06-18
fix: move output below input to be less confusing

### LINKED ISSUE #1797: Traffic Graph always shows outgoing abve incoming, even when it is lower
**Describe the bug**
The outgoing line is always higher than incoming

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to http://webui.ipfs.io.ipns.localhost:8080/#/
2. Scroll to the traffic graph
3. Wait for incoming to be higher than outgoing

**Expected behaviour**
If the outgoing is lower, the line should go down below incoming

**Screenshots**
Screen Shot 2021-05-29 at 11 38 48

 - OS: macOS 11.4
 - Browser Firefox
 - Version 90.0a1 (2021-05-29)

**Additional context**

olizilla:
...in some scenarios. it's a stacked graph, so the total height is in+out, when out is tiny and in is large it is not obvious that out is a tiny sliver stacked on top.

Screenshot 2021-06-18 at 15 05 06

fixes #1797
License: MIT

### DIFF
--- a/src/bundles/node-bandwidth-chart.js
+++ b/src/bundles/node-bandwidth-chart.js
@@ -22,7 +22,7 @@ function createNodeBandwidthChart (opts) {
     doUpdateNodeBandwidthChartData: (bw, timestamp, chartData) => ({ dispatch }) => {
       chartData = {
         in: chartData.in.concat({ x: timestamp, y: parseInt(bw.rateIn.toFixed(0), 10) }),
-        out: chartData.out.concat({ x: timestamp, y: parseInt(bw.rateOut.toFixed(0), 10) })
+        out: chartData.out.concat({ x: timestamp, y: parseInt(bw.rateOut.toFixed(0) * -1, 10) })
       }
 
       const startIndex = chartData.in.findIndex(d => d.x >= timestamp - opts.windowSize)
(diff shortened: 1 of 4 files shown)


olizilla:
Up next, actually seperate out in vs out.

olizilla:
it's also confusing that the 2 guages do not use the same range, but are ranged per direction... it makes visual comparison of the 2 gauages meaningless. They need to share the same max value.

olizilla:
last commit fixes it so out going bandwidth is negative so you can visually compare in vs out

Screenshot 2021-06-18 at 16 16 42

lidel (changes requested):
Thank you @olizilla, this approach looks like a good way to fix confusion causing issues like #1797 

Two asks:
1. Can you rebase this on top of `main` branch, so this PR is on top of v0.12.3  (right now this PR is based on v0.12.0 – perhaps you have old `master` locally? It got renamed a while ago). 
   - This is important because we changed units in https://github.com/ipfs/ipfs-webui/pull/1791 (v2.12.3) and we talked about reverting units for bandwidth during friday's triage.
2. (cosmetic) Negative numbers in [point popup and left axis look bit weird](https://user-images.githubusercontent.com/157609/122790144-4f3ac880-d2b8-11eb-9823-994ca6d8ca2d.png) but if its not possible to fix it without changing the lib responsible for plotting graph,

lidel:
Rebased and reverted BW unit change from https://github.com/ipfs/ipfs-webui/pull/1791 – WebUI should now show the same numeric values as `ipfs stats bw`.
I also tweaked units, values in tooltip and legend to be unified as well:

>

lidel:
Just to be sure, units+values match confirmed with quick eyeballing:

>

lidel (approved):

PR #1810, opened 2021-06-24
feat: link to pinning service docs if available

lidel:
This PR adds a link to docs next to "Secret access token", making it easier for users of predefined services like Pinata to learn how to generate a new token. 

The link to docs is defined in https://github.com/ipfs/ipfs-webui/blob/dadaf04219f2cc38118ddcde48f4b42fb1de2643/src/constants/pinning.js#L9

>

### DIFF
--- a/public/locales/en/settings.json
+++ b/public/locales/en/settings.json
@@ -39,7 +39,7 @@
     "apiEndpoint": "API endpoint",
     "apiEndpointPlaceholder": "URL for its API endpoint",
     "secretApiKey": "Secret access token",
-    "secretApiKeyHowToLink": "How to generate a new token?",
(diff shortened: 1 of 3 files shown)


public/locales/en/settings.json
```diff
@@ -39,6 +39,7 @@
     "apiEndpoint": "API endpoint",
     "apiEndpointPlaceholder": "URL for its API endpoint",
     "secretApiKey": "Secret access token",
+    "secretApiKeyHowToLink": "How to generate a new token?",
```
momack2, line 42:
```suggestion
    "secretApiKeyHowToLink": "How to generate a new token",
```

momack2 (approved):
LGTM - would just remove the "?"

PR #1811, opened 2021-06-29
fix: add api input placeholder

### LINKED ISSUE #1807: Add a placeholder to the API URL textbox
If the IPFS Desktop client can't connect to a previously available node, then the user needs to enter a new URL to the API. The problem is that the format of the URL is not obvious, making it difficult to guess the right text to enter.

It would be great to have a placeholder (gray text in the empty textbox) if no value was entered as API URL.

The textbox could be also prefilled with the previously used URL even if that node is not reachable at the moment.

**Additional context**
Is this the right format?
image

If the textbox is empty, a validly formatted example URL should be displayed as the background of the textbox.
image

lidel:
Closes #1807

>  Screenshot 2021-06-30 at 00-07-47 Settings IPFS

### DIFF
--- a/public/locales/en/app.json
+++ b/public/locales/en/app.json
@@ -42,6 +42,9 @@
     "connectingToIpfs": "Connecting to IPFS…",
     "couldNotConnect": "Could not connect to the IPFS API"
   },
+  "apiAddressForm": {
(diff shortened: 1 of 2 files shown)


jessicaschilling (approved):
LGTM!

PR #1826, opened 2021-07-24
Try to get a custom shareble link.

feliciss:
Is there a way to get a custom shareable link but omitting ensureMFS function?

replace ipfs.io to 

```
return `${gatewayUrl}/ipfs/${cid}${filename || ''}`
```


lidel:
Thank you for poking at this! 
We've added GUI for customizing the public gateway in https://github.com/ipfs/ipfs-webui/pull/1834, so this PR can be closed.

PR #1828, opened 2021-07-24
Change availableGatewayUrl to gatewayUrl.

feliciss:
Currently, the availableGatewayUrl in src/files/file-preview/FilePreview.js is always returning ipfs.io. it means you cannot preview a file on the custom host machine. I believe this patch will solve the problem - if I am no wrong in looking at the code.


lidel:
Superseded by https://github.com/ipfs/ipfs-webui/pull/1834

PR #1834, opened 2021-08-14
fix: local and public gateway links when cannot preview

### LINKED ISSUE #1767: The private network browsing file references the ipfs.io gateway
**Describe the bug**
The private network browsing local cluster file references the `ipfs.io` gateway.
image

hacdias:
Fixes #1767 by implementing the suggested solution: link to both local gateway and to ipfs.io.

Please note that `availableGatewayUrl` can be either the local gateway or ipfs.io and that is still used for all other `type`s.

I could not reproduce #1767 since my computer always got the correct local gateway URL as the available gateway. To show both links, we hard code the ipfs.io link and fetch the actual `localGatewayUrl` from settings.

image

This PR superseeds #1826.

License: MIT


lidel (changes requested):
Looks good, but we should remove hardcoded ipfs.io and enable users to use gateways of their choosing.

1. rename "with ipfs.io" to "public gateway"
1. hide "local gateway" if it is same as public one
1. make public gateway customizable
   - add "Public Gateway" section to "Settings" screen  (you can reuse "API" section for look and feel)
   - it should read value from `window.localStorage.get('ipfsPublicGateway')` and if missing fallback to showing https://dweb.link
   - replace `ipfs.io` hardcoded in the app with value from `window.localStorage` (with fallback to https://dweb.link)

hacdias:
### If Public Gateway === Available Gateway (local)

Screen Shot 2021-08-17 at 17 29 33

### Otherwise

Screen Shot 2021-08-17 at 17 29 41

### Public Gateway Setting

Screen Shot 2021-08-17 at 17 29 47

### If bad http url

Screen Shot 2021-08-17 at 17 29 54

### If empty

Screen Shot 2021-08-17 at 17 30 00

### Default: https://dweb.link

hacdias:
Updated @lidel

lidel (changes requested):
Looks sweet!  

I'm excited for this one – we will leverage the `window.localStorage` trick in brave and ipfs-companion, to ensure webui uses gateway user set in respective GUI :heart_eyes_cat: 

### Needs review

- [ ] Confirm with native speaker that labels look ok 
  - (cc @autonome  mind quickly eyeballing changes in https://github.com/ipfs/ipfs-webui/pull/1834/files → ` public/locales/en/*` ? preview in screenshots in comment above)

### Needs fixing:

- [ ]  Opening non-unixfs DAG like http://localhost:3000/#/ipfs/baf4bcfg4ep767tjp5lxyanx5urpjjgx5q2volvy produces malformed message: 
  > 
- [x] `Share link` should use gateway set on Settings, instead of hardcoded `ipfs.io`: 
  > 
  >  Screenshot 2021-08-18 at 18-11-14  aaa_test Files IPFS

hacdias:
@lidel both fixed. The first point "Opening non-unixfs DAG like" is translation related (see e12efec). I only updated for English as I'm assuming everything else needs to go through Transifex.

hacdias:
This PR superseeds and thus closes #1826.

lidel (approved):
Thank you! 

Tweaked UI a bit so buttons are active only when they actually do something:

Screenshot 2021-09-01 at 21-33-02 Settings IPFS

Merging + waiting with release till Friday for new labels to appear on Transifex.

autonome:
bah just saw this - the language and changes look good to me!

PR #1836, opened 2021-08-18
fix: do not load binary content

### LINKED ISSUE #1819: freeze when open a epub file in files tab
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

image

**Smartphone (please complete the following information):**

**Additional context**

hacdias:
Fixes #1819 by stopping to load the content after we determine it is binary.

I'm not completely sure why it works. I tried playing with the `loadContent` function to see if anything was wrong, but couldn't find the culprit.

@rafaelramalho19 I'm assigning you to review to make sure that the removal of that line doesn't have any issues. I'm not sure why we need to call `loadContent` after we determine that it is a binary file.

License: MIT

### DIFF
--- a/src/files/file-preview/FilePreview.js
+++ b/src/files/file-preview/FilePreview.js
@@ -104,7 +104,6 @@ const PreviewItem = ({ t, name, cid, size, type, availableGatewayUrl: gatewayUrl
       }
 
       if (isBinary(content)) {
-        loadContent()
         return cantPreview
       }
 


lidel (commented):
I suspect `loadContent` is related to progressive load of big text files added in https://github.com/ipfs/ipfs-webui/pull/1646 – @rafaelramalho19 may remeber more.

hacdias:
> I suspect `loadContent` is related to progressive load of big text files added in #1646 – @rafaelramalho19 may remeber more.

Exactly, but such thing is not needed when we detect that we're looking at a binary file. So there's no point on keeping loading it I think. So let's wait for @rafaelramalho19

lidel (approved):
I did some poking and removal of `loadcontent` lgtm – most likely a scar tissue from past refactors.

While at it, improved `isBinary` check in 39b172e – takes filenames into account + should behave better around small binary files.

Issue #1798, opened 2021-06-01
Import/export CAR DAG archives


lidel:
Given that:
- `ipfs dag import|export` exists
- go-ipfs 0.9+ supports exporting DAG archive via any public gateway (`/api/v0/dag/export`)
- tools like [ipfs-car](https://www.npmjs.com/package/ipfs-car) and [ipfs-content-addressed-archiver](https://github.com/olizilla/ipfs-content-addressed-archiver) make it easier for people to produce/consume content-addressed archives

We should support "DAG import/export"  in ipfs-webui somehow.

## Import DAG

Import could  work the same as "From IPFS" :

> ![2021-06-01--13-01-05](https://user-images.githubusercontent.com/157609/120312856-8a5d6380-c2d9-11eb-85a4-a6e602c93020.png)

We most likely want to add an explicit "From CAR Archive"  item here  (personally, I like the idea of a whimsical car icon  :)), but "stroke data" from [ipfs-css](https://github.com/ipfs-shipyard/ipfs-css#icons) would also do.

## Export DAG

Add complimentary "Export DAG" to context menu on Files screen:

> ![2021-06-11--16-08-29](https://user-images.githubusercontent.com/157609/121699752-7ed71d00-cacf-11eb-873e-6b73b1ef0abd.png)

Gozala:
Seems like it would be a good idea to do it after https://github.com/ipfs/js-ipfs/issues/2745

hacdias:
@lidel as I mentioned on https://github.com/ipfs/ipfs-webui/pull/1838#issuecomment-901733616, we'll need to update to the latest HTTP client across WebUI and the Explore page to support the import part. The export is done on #1837. I don't think I should work on the upgrade right now as that would take a long time working only 10h a week. Updating the http client will also require updating everything else connected: multiformats, ipld, multicodecs, etc.

I think we could merge the export button and evaluate how much effort is needed for the import part.

hacdias:
@lidel I agree with your worry. I didn't think about that, but it's a good point.

lidel:
Reopening since we also need https://github.com/ipfs/ipfs-webui/pull/1838

hsn10:
Add maximum CAR size so output can be split into multiple files. Standard between services seems to accept 100 MB CAR files for uploads.

lidel:
AS of 2025Q1 we do have support for CAR export:

[...]

PR #1837, opened 2021-08-18
feat: download as CAR via the context menu 

hacdias:
Partially addresses #1798.

Screen Shot 2021-08-18 at 11 56 20.

License: MIT

### DIFF
--- a/public/locales/en/app.json
+++ b/public/locales/en/app.json
@@ -27,7 +27,8 @@
     "setPinning": "Set pinning",
     "submit": "Submit",
     "unpin": "Unpin",
-    "unselectAll": "Unselect all"
+    "unselectAll": "Unselect all",
+    "downloadCar": "Download as CAR"
   },
   "cliModal": {
     "description": "Paste the following into your terminal to do this task in IPFS via the command line. Remember that you'll need to replace placeholders with your specific parameters."
--- a/src/lib/files.js
+++ b/src/lib/files.js
@@ -148,6 +148,26 @@ export async function getShareableLink (files, gatewayUrl, ipfs) {
 }
 
 /**
+ *
+ * @param {FileStat[]} files
+ * @param {string} gatewayUrl
+ * @param {IPFSService} ipfs
+ * @returns {Promise<string>}
+ */
+export async function getCarLink (files, gatewayUrl, ipfs) {
+  let cid, filename
+
+  if (files.length === 1) {
+    cid = files[0].cid
+    filename = files[0].name
+  } else {
+    cid = await makeCIDFromFiles(files, ipfs)
+  }
+
+  return `${gatewayUrl}/ipfs/${cid}?format=car&filename=${filename || cid}.car`
+}
+
+/**
  * @param {number} size in bytes
  * @param {object} opts format customization
  * @returns {string} human-readable size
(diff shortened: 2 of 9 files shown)


rafaelramalho19 (approved):

lidel (commented):
@rvagg @olizilla @mikeal   @alanshaw 
Would appreciate some opinionated  bikeshed around labels related to DAG export/import in UI (webui/desktop).

Does the above screenshot look sensible?

Mainly, I want this to be consistent with messaging we have in dev docs in other places (are we using "DAG export" or "CAR download" etc)?

rvagg:
I think this is probably OK, we've put IPFS' "export" inside the "dag" subsection and it's technically correct that it's exporting an entire DAG starting from a root.

We have other place that are front-loading "CAR", like web3.storage which has a `putCar()` API (though not in docs yet I think), but I'm not sure that's the ideal approach as it would be good to have some wriggle room around transports for moving DAGs around and not box ourselves too tightly into a form. Even the `putCar()` API in web3.storage just means "give me an object that conforms to the `CarReader` API, which it then just uses as a simple root provider blockstore (I faked this for percid.va.gg so I could use that API but not deal with CARs).

lidel (changes requested):
LGTM but let's park this until we have a PR with import  as noted in https://github.com/ipfs/ipfs-webui/issues/1798#issuecomment-910930951

src/lib/files.js
```diff
@@ -147,6 +147,29 @@ export async function getShareableLink (files, gatewayUrl, ipfs) {
...
+  let cid
+
+  if (files.length === 1) {
+    cid = files[0].cid
+  } else {
+    cid = await makeCIDFromFiles(files, ipfs)
+  }
+
+  return {
+    url: `${apiUrl}/api/v0/dag/export?arg=${cid}`,
```
lidel, line 167: @hacdias go-ipfs 0.13 will allow for fetching CAR from gateway, mind refactoring this, so we use below?

```suggestion
    url: `${gatewayUrl}/ipfs/${cid}?format=car&filename=${filename || cid}.car`,
```

(would be cool if we could preserve filename, if not, cid is fine too)

public/locales/en/app.json
```diff
@@ -27,7 +27,8 @@
     "setPinning": "Set pinning",
     "submit": "Submit",
     "unpin": "Unpin",
-    "unselectAll": "Unselect all"
+    "unselectAll": "Unselect all",
+    "exportDag": "Export DAG"
```
lidel, line 31: All docs in ecosystem refer to CARs, maybe it is better to rename this to:
```suggestion
    "exportDag": "Download as CAR"
```

lidel (changes requested):

hacdias:
@lidel I rebased and made the necessary updates to use the new gateway URLs and the nomenclature. I will now update #1838 such that it matches and is rebased.

Besides, I added some of the mechanics we use in #1894 such that we don't store the whole file in memory downloading it. We create an artificial `a` block and click on it.

On a different note, the gateway download doesn't follow the `filename` query parameter. I opened an issue in https://github.com/ipfs/go-ipfs/issues/9027 about it.

lidel (approved):
LGTM, this is unblocked because  go-ipfs 0.13 supports `?format=car` ([release notes](https://github.com/ipfs/go-ipfs/releases/tag/v0.13.0)) and we now have PR with gateway specs at https://github.com/ipfs/specs/pull/283

Needs small unicode fix below, but I'll apply it shortly. 

Unsure why CI deps install fail, probably the usual NPM rot. 
Will regenerate package-lock and see if that helps.

PR #1838, opened 2021-08-18
feat: import DAG CAR

hacdias:
License: MIT


hacdias:
Okay, so this will involve updating the ipfs-http-client to have access to `dag.import` which will be a larger task than expected.

SgtPooki:
we should be able to import a car using helia similar to https://github.com/ipfs/helia/blob/8a5bc6f03d7d87a8a07ead00216ac2704f36d9ab/packages/interop/src/car.spec.ts#L44-L75

without having to deal with car file support in js-kubo-rpc-client

hacdias:
@SgtPooki that sounds like a better approach indeed. I'm closing this PR as it's quite outdated anyways.

PR #1846, opened 2021-09-02
feat: check if gateway is valid using image

### LINKED ISSUE #1844: Settings: validate public gateway before saving it
## Current state 

#1834 introduced ability to customize public gateway used for shareable URLs  and download via preview screen.
The only validation we do is making sure the entered string is a valid URL.

## Proposed improvement

We should validate that the gateway is functional before storing it in `localStorage` to avoid users breaking their experience by entering an invalid gateway address (typo, server being down, etc).

### Note on how to validate

Some gateways may not allow cross-origin requests (CORS), so we can't test them using regular `fetch`.
Workaround: use Image load instead.`` tags are not being validated using CORS rules.  

Prior art is `checkViaImgSrc` –  can be found [here](https://github.com/ipfs/public-gateway-checker/blob/8c5b63e8815269d3ac9aae70d2c691298ca78e17/app.js#L83-L108).

hacdias:
Closes #1844 by using the same code we use in Companion (check mentioned issue).

It verifies if the gateway is valid only when submitting. I feel like validating on change would be better in terms of UX, but it can also be problematic as we have a 15 seconds timeout to account for the gateway to fetch the actual content.

License: MIT

### DIFF
--- a/src/bundles/gateway.js
+++ b/src/bundles/gateway.js
@@ -1,6 +1,7 @@
 import { readSetting, writeSetting } from './local-storage'
 
 export const DEFAULT_GATEWAY = 'https://dweb.link'
+const IMG_HASH = 'bafybeibwzifw52ttrkqlikfzext5akxu7lz4xiwjgwzmqcpdzmp3n5vnbe' // 1x1px image
 
 const readPublicGatewaySetting = () => {
   const setting = readSetting('ipfsPublicGateway')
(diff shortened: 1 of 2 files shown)


lidel (approved):
LGTM. :+1:

I think checking on submit is perfectly fine for basic sanity check that protects users against URL that does not act as a gateway.

We could check on every change, but then we have to add "Testing..." status that disables submit button until test  is done + account for various edge cases. Fine to do this as follow-up PR, but i'm merging this as-is because it's already an improvement  :sparkles:

PR #1850, opened 2021-09-10
feat: add peer permanently by default

### LINKED ISSUE #1849: Peers: ability to add permanent peering
## Current UX

> Screenshot 2021-09-09 at 22-33-16 Peers IPFS

Right now, only a simple,  one time connection request can be executed via Add Connection button. 
It executes `ipfs swarm connect` behind the scenes, but does not persist the connection across restarts.

## New UX

We should improve existing Add Connection modal and add checkbox "Add to the permanent peering config" which is checked on by default.

When checked, this new action should execute `ipfs swarm connect` just like the old did, but if that is successful, 
it should add peer details to the config file under [peering section](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#peering), making sure it is persisted across restarts and disconnects.

Doing this is important, as it aligns webui with peering UI provided by Brave.

hacdias:
Closes #1849.

@lidel I added the configuration change in the same bundle function as the swarm connect. Do you think it would be better to change to the config bundle?

Also, I noticed that when you change the config, if you go to the Settings page, the config isn't changed and you need to refresh. Somehow we need to trigger the config re-fetching here.

### DIFF
--- a/src/bundles/peers.js
+++ b/src/bundles/peers.js
@@ -32,7 +32,7 @@ bundle.selectPeersCount = createSelector(
   }
 )
 
-bundle.doConnectSwarm = (addr, permanent) => async ({ dispatch, getIpfs }) => {
+bundle.doConnectSwarm = (addr, permanent) => async ({ dispatch, getIpfs, store }) => {
   dispatch({ type: 'SWARM_CONNECT_STARTED', payload: { addr } })
   const ipfs = getIpfs()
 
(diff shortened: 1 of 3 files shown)


src/bundles/peers.js
```diff
@@ -31,12 +32,34 @@ bundle.selectPeersCount = createSelector(
...
+          // add new addr to existing address list for the peer
+          preexisting.Addrs.push(rawAddr)
+        }
+      } else {
+        // add new peer to the list
+        peers = [...peers, { ID: peerId, Addrs: [rawAddr] }]
+      }
+
+      await ipfs.config.set('Peering.Peers', peers)
+      await store.doMarkConfigAsOutdated() // force Settings screen to re-fetch
```
lidel, line 61: @hacdias fyi this takes care of triggering config refresh for Settings screen (iiuc this is hidden/built-in feature of react/redux)

lidel (approved):
LGTM, thanks for adding this, super useful improvement  :heart: 

- switched to the Checkbox we already used on Settings 
- tweaked the way we modify `Peering.Peers` to follow format from https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#peering
- made config on _Settings_ screen pick up changes to  `Peering.Peers` section done here

PR #1852, opened 2021-09-13
Fix navbar links to the github repo.

AyumuKasuga:
These links at the bottom:

image

Currently redirect does not work for `Revision` link and we get 404 page there.


lidel (approved):
Nice catch, thanks! :+1:

PR #1857, opened 2021-09-20
feat: IPNS Publishing

### LINKED ISSUE #1482: Support for publishing to IPNS
It's annoying to have to use the terminal to publish content to an IPNS key, and it alienates non-powerusers.

It would be nice if the webui let you generate, organize, and name IPNS keys that files and folders could then be published to via the dropdown menu.

hacdias:
Closes #1482.

-  [x] Add IPNS key management UI to Settings page
-  [x] Publishing to IPNS id via context action on Files page

## A bunch of screenshots

Screen Shot 2021-09-24 at 13 29 10

The context menu is a bit unaligned. The same happens with the Pinning Services context menu. The target reference seems correctly set so I'm not sure why's this miscalculation happening.

Screen Shot 2021-09-24 at 13 29 27

Screen Shot 2021-09-24 at 13 37 16

Screen Shot 2021-09-24 at 13 37 27

Screen Shot 2021-09-24 at 13 37 35

Screen Shot 2021-09-24 at 13 37 50

image

### DIFF
--- a/.github/workflows/tx-pull.yml
+++ b/.github/workflows/tx-pull.yml
@@ -5,35 +5,17 @@ on:
     - cron: '0 0 * * 0'
 
 jobs:
-  install-tx-client-and-pull-translations:
+  tx-sync:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Cache public folder
-      uses: actions/cache@v3
-      with:
-        path: public
-        key: ${{ github.sha }}
     - name: Install Transifex client
       run: |
         curl -o- https://raw.githubusercontent.com/transifex/cli/master/install.sh | bash
+        chmod +x ./tx
     - name: Pull translations from Transifex
       run: |
-        tx -t ${{ secrets.TX_TOKEN }} pull -a
-
-  pull-request:
-    runs-on: ubuntu-latest
-    steps:
-    - name: Get cached public folder
-      uses: actions/cache@v3
-      with:
-        path: public
-        key: ${{ github.sha }}
-
-    # - name: Setup node
-    #   uses: actions/setup-node@v2
-    #   with:
-    #     node-version: 16.14.x
+        ./tx -t ${{ secrets.TX_TOKEN }} pull -a -f
     - uses: stefanzweifel/git-auto-commit-action@49620cd3ed21ee620a48530e81dba0d139c9cb80
       with:
         # Optional. Commit message for the created commit.
@@ -43,7 +25,7 @@ jobs:
         # Optional. Local and remote branch name where commit is going to be pushed
         #  to. Defaults to the current branch.
         #  You might need to set `create_branch: true` if the branch does not exist.
-        branch: tx-pull-${{ github.sha }}
+        branch: i18n-sync
 
         # Optional. Options used by `git-commit`.
         # See https://git-scm.com/docs/git-commit#_options
@@ -54,7 +36,7 @@ jobs:
         # See the `pathspec`-documentation for git
         # - https://git-scm.com/docs/git-add#Documentation/git-add.txt-ltpathspecgt82308203
         # - https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec
-        file_pattern: public/**/*.json
+        file_pattern: public/locales
 
         # Optional. Local file path to the repository.
         # Defaults to the root of the repository.
@@ -62,7 +44,11 @@ jobs:
 
         # Optional. Options used by `git-add`.
         # See https://git-scm.com/docs/git-add#_options
-        add_options: '-u'
+        add_options: '-A'
+
+        # Optional. Options used by `git-push`.
+        # See https://git-scm.com/docs/git-push#_options
+        push_options: '--force'
 
         # Optional. Disable dirty check and always try to create a commit and push
         skip_dirty_check: true
@@ -80,11 +66,11 @@ jobs:
         # Optional. Create given branch name in local and remote repository.
         create_branch: true
     - name: pull-request
-      uses: repo-sync/pull-request@ff8726ae82c9970bde6527e419133aa167297c40
+      uses: repo-sync/pull-request@65785d95a5a466e46a9d0708933a3bd51bbf9dde
       with:
-        source_branch: "tx-pull-${{ github.sha }}"
+        source_branch: "i18n-sync"
         destination_branch: "main"
-        pr_title: "chore: Pull transifex translations"
+        pr_title: "chore: pull new translations"
         pr_body: "Automated PR created by .github/workflows/tx-pull.yml"
         pr_label: "area/i18n/translations"
         pr_draft: false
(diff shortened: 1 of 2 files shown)


hacdias:
This one is ready @lidel :D

src/bundles/ipns.js
```diff
@@ -0,0 +1,63 @@
...
+
+  doGenerateIpnsKey: (name) => async ({ getIpfs, store }) => {
+    const ipfs = getIpfs()
+
+    // Stronger defaults than the listed on the docs:
+    // https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/KEY.md#ipfskeygenname-options
+    await ipfs.key.gen(name, {
+      type: 'ed25519',
+      size: 4096
+    })
```
lidel, line 39: I believe those defaults changed a while ago in both js-ipfs and go-ipfs.
- [ ] TODO: check if we can switch to the default now.

lidel (commented):
Finally, got to spend some time with this :)
Works really good, want to include it in go-ipfs 0.11 if possible.

Before we merge this, need to add some e2e tests tomorrow:
- [ ] add key and confirm pet name is added to the table
  - [ ] publish to the key from files
    - [ ] confirm link on the settings screen got updated 
  - [ ] remove key and confirm it's gone from the table


For now, pushed small changes which tweak  some modals a bit:


> Screenshot 2021-11-25 at 02-07-29  animals Files IPFS

> Screenshot 2021-11-25 at 02-10-31 Settings IPFS

> Screenshot 2021-11-25 at 02-10-00 Settings IPFS

lidel (changes requested):
Added E2E tests and cleaned them up a bit, should  be pretty robust now that we use playwright. 

--- 

@hacdias I think we have UX gap around publishing, which blocks this from being shipped.

Need some sort of confirmation that publishing was successful. 

## Visual feedback on IPN Publish

Right now I click Publish and nothing happens. 
We need three states:
- "Publishing in progress"
- "Publishing successful" / "Publishing failed"

A nice flow would be if clicking on "Publish"  displayed spinner until `ipfs.name.publish` returns result (success/error), and then:
- if error, show error inside of the modal.
- if success, show "Succes, published at {keyid}" and below the same UI we have on "Share" modal, where user can copy a shareable link to `/ipns/` address at a public gateway.


@hacdias will you have bandwidth to add this? 
I think it is better to wait and polish the experience, so I'm descoping this from go-ipfs 0.11, so there is no rush.

src/bundles/ipns.js
hacdias: Ah, both `js-ipfs` and `go-ipfs` default to keys of type `ed25519`. As far as I understand, the size does not influence this types of keys, so I will clean that up.

hacdias:
@lidel I'll work on it.

hacdias:
@lidel it's still missing the actual spinner. Could you just validate if the workflow is good?

lidel:
@hacdias yes, workflow is :ok_hand:  

Figuring out an engaging spinner will be tricky – some thoughts:
- IPNS publishing can take up to a minute atm because (iirc) it puts IPNS record in 20 peers, and it has to find the right ones first.  (It is faster if [Accelerated DHT Client](https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md#accelerated-dht-client) is enabled, but can't use it by default yet).
- A fake progressbar that assumes to take ~1m should be enough of feedback / setting expectations. We can improve it later.
- If you want to get creative, we could inform user about DHT traversal, eg. show random DHT walk with  [await first(ipfs.log.tail())](https://github.com/ipfs/js-ipfs/blob/484d2559cfe120f01eb9c5dc728d1f7cfbd89730/packages/ipfs-http-client/test/log.spec.js#L21-L32) or fake it.

ps. fysa pushed two changes:
- Finished E2E tests for IPNS publishing – they now confirm IPNS record was updated by doing `ipfsn.name.resolve`.
- Added some copy text above the shareable link. (just a draft, when we are done we will ask a native speaker to proofread and refine messaging and UI)
  >  Screenshot 2021-12-03 at 00-35-33  animals Files IPFS

hacdias:
> A fake progressbar that assumes to take ~1m should be enough of feedback / setting expectations. We can improve it later.

I will do this. Seems to be the easiest. Using a spinner would probably not be the greatest idea idea, indeed, as it does not give feedback.

hacdias:
@lidel I added a progress bar that is initially 60 seconds. After each publishing, we average the current expected time with the new time it take. I think that this way, it should improve and have better expectations over time.

hacdias:
Ping @lidel

src/files/modals/publish-modal/PublishModal.js
```diff
@@ -0,0 +1,157 @@
...
+  }, [selectedKey, start])
+
+  useEffect(() => {
+    doFetchIpnsKeys()
+  }, [doFetchIpnsKeys])
+
+  useEffect(() => {
+    if (!start) return
+
+    const interval = setInterval(() => {
```
lidel, line 32: :thought_balloon:  nit: I wonder if we could remove `setInterval` here and use [css animations](https://stackoverflow.com/question/15945288/animating-progress-bars-with-css) instead somehow?  (would look less choppy, we could also make it more engaging by making it non-linear). 

I'm thinking setting `progress` to 100% and using  `animation: progressBar ${expectedPublishTime}s ease-in-out`  for customizing the progress animation instead.

Just an idea, feel free to ignore.

lidel (changes requested):
@hacdias really nice! 

* I adjusted the rolling average to overshoot rather than undershoot (apps fake progress and usually finish around 75% to give that "snappy feeling", we guess based on past times, which is better, but still should overshoot to avoid being stuck at 100%) + added brief explainer what is happening, so the wait is a bit more justified:
  > 
- I'm marking this as blocked on go-ipfs roadmap:
  - We need to figure out  if it is ok to ship the GUI for IPNS before we have things like  https://github.com/ipfs/go-ipfs/issues/8591 and https://github.com/ipfs/go-ipfs/issues/8586 in go-ipfs, or are those worth waiting for.  Will pick this up and make the decision in January.

hacdias:
@lidel I added an option that will make `ProgressBar` use CSS animations. I also added an `overflow: hidden` to the parent `div` so that the colored part/progress does not overflow.

lidel (changes requested):
@hacdias I don't  believe https://github.com/ipfs/go-ipfs/issues/8591 or https://github.com/ipfs/go-ipfs/issues/8586 will happen this quarter,  let's not wait anymore. I think it is ok for us to ship it because    IPNS publishing should be in IPFS Desktop, and unlike the upstream kubo (go-ipfs), IPFS Desktop  has IPNS over pubsub enabled by default, so performance will be acceptable.

Mind resolving conflicts in this PR and pinging me when ready?  :pray:

SgtPooki:
@juliaxbow this is a great candidate for your review and help!

hacdias:
I tried a new way of rebasing and I guess I _messed up_ this PR. Sorry for that and won't happen again. The branch is fine, but this PR got closed and I can't re-open it. 

**Continue on #1973**.

PR #1890, opened 2022-01-14
fix: download HAMT-sharded directories

### LINKED ISSUE #1885: Downloading Multiple Large Files Fails
**Describe the bug**

Downloading large files fails due to the usage of `ipfs object add-link`. See https://github.com/ipfs/go-ipfs/issues/7421.

**To Reproduce**

1. Go to Files
2. Click on multiple large files
3. Click on 'Download'

**Expected behavior**
Download the files.

**Additional context**

See https://github.com/ipfs/go-ipfs/issues/7421. We should move to the MFS API, perhaps through some kind of temporary file on the MFS? 

Perhaps something such as `/temp/Download {timestamp}`. cc @lidel

hacdias:
fixes #1885

License: MIT

### DIFF
--- a/src/lib/files.js
+++ b/src/lib/files.js
@@ -68,18 +68,19 @@ async function downloadSingle (file, gatewayUrl, apiUrl) {
  * @returns {Promise<CID>}
  */
 export async function makeCIDFromFiles (files, ipfs) {
-  let cid = await ipfs.object.new({ template: 'unixfs-dir' })
+  const dirpath = `/zzzz_${Date.now()}`
+  await ipfs.files.mkdir(dirpath, {})
 
-  for (const file of files) {
-    cid = await ipfs.object.patch.addLink(cid, {
-      name: file.name,
-      // @ts-ignore - can this be `null` ?
-      size: file.size,
-      cid: file.cid
-    })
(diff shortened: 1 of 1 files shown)


lidel (approved):
LGTM.

Disclaimer for a future reader: this is the best we could do, given what go-ipfs 0.11 provides:
- `object patch` is deprecated. It is not aware of HAMT-sharding, nor it will keep block size under 1MB
  - unsafe to use, we have to remove this ASAP
- https://github.com/ipfs/go-ipfs/issues/8106 made MFS pretty smart: it will autoshard/unshard and keep block sizes in check
  - the downside is that this MFS dance looks a bit awkward: it "pollutes" MFS root with temporary directory, but we prefix it with `zzzz` so it should not impact GUIs even when Files screen is opened at the time of execution.

In the future:
- we will have `dag patch`, but it will work at the DAG level (too low, does not abstract sharding)
- .. so we need to find a better abstraction for creating ephemeral directories for use cases like this one
  - could be `ipfs files patch` that works on MFS paths but also on `/ipfs/{cid}`, becoming true replacement for `ipfs object patch`
  - ..or ability to have more than one  MFS root
  - .. or something else

PR #1891, opened 2022-01-14
fix: replace .substr by .substring

hacdias:
`.substr` is deprecated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

License: MIT


lidel (approved):
Thanks!

PR #1894, opened 2022-01-16
feat: use direct links to download all files

### LINKED ISSUE #1887: File → menu → Download buffers entire file into a blob
- OS: Windows 10 x64
 - Version of IPFS Desktop [0.18.1]

**Describe the bug**

https://streamable.com/9z175t

**To Reproduce**
1. Go to the file
2. Click on more
3. Scroll down to Download
4. Blank Screen after some seconds.

**Log files:**

[combined.log](https://github.com/ipfs/ipfs-desktop/files/7834484/combined.log)
[error.log](https://github.com/ipfs/ipfs-desktop/files/7834485/error.log)

**PD:** I tried reinstalling, but same problem.

hacdias:
This closes #1887 by using the gateway URLs to download the files. There's two things to mention:

1) I tried creating an anchor element to download the file with a name via the `download` attribute, but it seems to still use the name `get.tar.gz` sadly. Any suggestions?
2) There's some other code that used to show download progress. Should I remove it? I don't think it is used anywhere else in the code so it feels safe to remove unless it can be used for any planned feature.

License: MIT


src/files/FilesPage.js
```diff
@@ -57,14 +54,12 @@ const FilesPage = ({
...
-      return downloadAbort()
-    }
+    const { url, filename } = await doFilesDownloadLink(files)
 
-    const updater = (v) => setDownloadProgress(v)
-    const { url, filename, method } = await doFilesDownloadLink(files)
-    const { abort } = await downloadFile(url, filename, updater, method)
-    setDownloadAbort(() => abort)
+    const link = document.createElement('a')
+    link.download = filename // TODO: filename isn't working. Keep getting 'get.tar.gz' - is it being overwritten?
```
lidel, line 60: Oh wow, indeed, `download` attribute only works on links within the same Origin or.. `blob:` ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download)), given that users load webui from different port or domain than the RPC API, we can't use it here :crying_cat_face: 

We can link to standalone files directly, using gateway link, 
but for  directories that need to be returned as `tar.gz` we have no good option for setting filename on ipfs-webui side.

Unless someone comes up with a creative hack, I am afraid we need to park this PR until we get something like  [`?format=tar.gz`](https://github.com/ipfs/go-ipfs/issues/8234) on Gateways, or  `/api/v0/get` with support for  implicit or explicit `filename` parameter that modifies response and adds two HTTP headers:
```
Content-Type: application/octet-stream
Content-Disposition: attachment; filename=cid.tar.gz
```

I'd  rather wait for clean `?format=tar.gz`  (hoping to start working on it this quarter) as we want to remove `/api/v0` from gateways.

lidel (changes requested):
Re 1 – see inline comment
Re 2 – ok to remove the `dowload-file.js`, do you have something more in mind?

src/files/FilesPage.js
hacdias: @lidel please see #1945

hacdias:
The TARs are a thing on the gateway now: https://github.com/ipfs/kubo/pull/9029

This PR can be unblocked and worked on 😃

src/files/FilesPage.js
lidel: For anyone reading this in the future, this got resolved in https://github.com/ipfs/kubo/pull/9029, we can switch to that.

hacdias:
I decided to revitalise this PR since we now have TAR response formats on our gateway (https://github.com/ipfs/kubo/pull/9029). That lets us download directories directly from the gateway without the need to use the API v0 POST method. Therefore, we do not need to keep anything in memory and it is all streamed to the user. This allows to massively simplify the code, removing a lot of old bits that are no longer relevant 🪩

This can be tested with Kubo from master, and with 0.17.0 RC-1. I'll mark this as ready for review as its been open for many months, but it's finally possible to review and test this PR.

src/lib/files.js
```diff
@@ -87,42 +76,20 @@ export async function makeCIDFromFiles (files, ipfs) {
...
-export async function getDownloadLink (files, gatewayUrl, apiUrl, ipfs) {
+export async function getDownloadLink (files, gatewayUrl, ipfs) {
   if (files.length === 1) {
-    return downloadSingle(files[0], gatewayUrl, apiUrl)
+    return getDownloadURL(files[0].type, files[0].name, files[0].cid, gatewayUrl)
   }
 
-  return downloadMultiple(files, apiUrl, ipfs)
+  const cid = await makeCIDFromFiles(files, ipfs)
+  return getDownloadURL('directory', '', cid, gatewayUrl)
```
lidel, line 92: @hacdias passing empty `name` here will always produce `download_<cid>.tar`.
It has been a while, do you remember if there is any reason why we can't preserve directory name and do something like `<dir_name>_<cid>.tar` ?

lidel (approved):
lgtm, small UX question below

src/files/FilesPage.js
```diff
@@ -58,28 +55,13 @@ const FilesPage = ({
...
-      link.click()
-    } else {
-      const updater = (v) => setDownloadProgress(v)
-      const { abort } = await downloadFile(url, filename, updater, method)
-      setDownloadAbort(() => abort)
-    }
+    const url = await doFilesDownloadLink(files)
+    const link = document.createElement('a')
+    link.href = url
+    link.click()
```
SgtPooki, line 61: non-blocking Q: instead of creating an element and then programmatically clicking on it can we have the thing the users click have the appropriate URL?

SgtPooki (commented):
The changes look okay, but we're removing some functionality: progress updates and download aborting. Is that intended, or am I missing something?

src/files/FilesPage.js
hacdias: > Note: I do agree that this is a significant improvement, and resolves https://github.com/ipfs/ipfs-webui/issues/1887 but I'm wondering if there is a way we can keep the ability to cancel a download, or display download progress. Fine with a separate issue for those items

Downloads will be handled by the browser UI. The user can just cancel it through the browser UI. We only had the option to cancel and display progress on the Web UI itself because we were storing **every single** byte of data in-memory before downloading.

hacdias: > non-blocking Q: instead of creating an element and then programmatically clicking on it can we have the thing the users click have the appropriate URL?

Yes and no. If you select multiple files, what link do you put? You can't as we calculate the CID dynamically.

whizzzkid: Can't we just redirect the users to download? it will prompt the user to download and not go away from current page too?

src/files/selected-actions/SelectedActions.js
```diff
@@ -70,37 +69,12 @@ class SelectedActions extends React.Component {
...
-    if (this.props.downloadProgress === 100) {
-      return this.props.t('finished')
-    }
-
-    return this.props.downloadProgress.toFixed(0) + '%'
-  }
-
   render () {
-    const { t, tReady, animateOnStart, count, size, unselect, remove, share, setPinning, download, downloadProgress, rename, inspect, className, style, isMfs, ...props } = this.props
+    const { t, tReady, animateOnStart, count, size, unselect, remove, share, setPinning, download, rename, inspect, className, style, isMfs, ...props } = this.props
```
whizzzkid, line 77: Can we sort these alphabetically?

whizzzkid (commented):
non-blocking nits

src/files/FilesPage.js
hacdias: It doesn't go away from the current page AFAIK. At least not for me?

src/files/selected-actions/SelectedActions.js
hacdias: Unless you know how to do it automatically, please don't make me do it 😅

src/files/FilesPage.js
hacdias: This is the behaviour of the current PR:

https://user-images.githubusercontent.com/5447088/201318295-aa0d3d56-d6d7-43a7-ba6b-d8e15022bc30.mp4

hacdias:
@SgtPooki @whizzzkid I addressed most of your comments (see replies). This is the behaviour of the current PR:

https://user-images.githubusercontent.com/5447088/201318295-aa0d3d56-d6d7-43a7-ba6b-d8e15022bc30.mp4

SgtPooki (approved):
Thanks a ton for remembering this PR and coming back to close it out @hacdias !

SgtPooki:
Do we expect any issues with how this will work within electron?

hacdias:
> Do we expect any issues with how this will work within electron?

I think it should be fine. We already had download links we this behaviour before and they were not a problem.

PR #1903, opened 2022-02-09
refactor: on-demand remote pin status checks

### LINKED ISSUE #1900: Pinning to Pinata disabled due to "Rate limit exceeded"
- OS: Windows
 - Version of IPFS Desktop 0.18.1

**Describe the bug**
Pinata integration stopped working. So it worked before and after some update in the recent past it stopped working. I thought it's a temporary outage as I saw it first. But that's quite a while ago, maybe a month or two, and it's still not working.

**To Reproduce**
1. -> Settings
2. Click "+ Add Service"
3. Select Pinata
4. Go to Pinata's API page (which is linked)
5. Create a new key
6. Add it to the dialog opened in step 3
7. Click on Save
8. Pinata's status is "Pins: Failed to fetch"

**Expected behavior**
The integrated service connection should work.

**Screenshots**
IPFS Desktop Pinata fail

lidel:
This is bare minimum to restore remote pinning with rate-limited Pinata and close #1900
Not a perfect solution, but a fast-tracked fix to restore pinning functionality **_at all_** (right now Pinata is effectively broken in webui and ipfs-desktop – details in https://github.com/ipfs/ipfs-webui/issues/1900#issuecomment-1032030124).

Key changes:

- [x]  _Files_ screen does not make automatic requests to remote pinning service every time a new directory is opened
  - we progressively build a cache with remote pin states every time user opens "Set pinning" modal, and send requests only for files that we did not check the remote state before
  - we persist  the `pinning` bundle state with remote pin states with  `createCacheBundle` (src/bundles/index.js) and `persistActions` (src/bundles/pinning.js), which store data in browser's IndexedDB (uses `money-clip` as cache backend)
  - new users should not see any regression:  the pin would be added to the cache during pinning operation, and thanks to persisted cache, cloud pin will show every time _Files_ screen loads.
  - old users may need to manually go to "Set pinning" to trigger status fetch, but that is the best we can do and a fair compromise, given nothing worked before this.  
- [x] The expensive `ipfs pin remote service ls --stat` checks  are executed  only on _Settings_ screen – _Files_ does not need them.
- [x] Remote pin status check is triggered by opening "_Set pinning_"  modal via context menu or the pin icon, or by clicking on "Pin Status" column header.
  - [x] Clicking on "Pin Status" header bypassess remote pin status cache and forces online status re-check for all items in current directory (bit of a hidden feature / escape hatch for frustrated, existing users)
- [x] Errors returned by pinning service are now correctly reported to the user. The CLI and ipfs-webui users will now see the same error message:
  >  
  >

### DIFF
--- a/src/bundles/pinning.js
+++ b/src/bundles/pinning.js
@@ -2,12 +2,13 @@
 import { pinningServiceTemplates } from '../constants/pinning'
 import memoize from 'p-memoize'
 
+const CID_PIN_CHECK_BATCH_SIZE = 10 // Pinata returns error when >10
+
 const parseService = async (service, remoteServiceTemplates, ipfs) => {
   const template = remoteServiceTemplates.find(t => service.endpoint.toString() === t.apiEndpoint.toString())
   const icon = template?.icon
   const visitServiceUrl = template?.visitServiceUrl
-  const autoUpload = await mfsPolicyEnableFlag(service.service, ipfs)
-  const parsedService = { ...service, name: service.service, icon, visitServiceUrl, autoUpload }
+  const parsedService = { ...service, name: service.service, icon, visitServiceUrl }
 
   if (service?.stat?.status === 'invalid') {
     return { ...parsedService, numberOfPins: -1, online: false }
@@ -15,8 +16,9 @@ const parseService = async (service, remoteServiceTemplates, ipfs) => {
 
   const numberOfPins = service.stat?.pinCount?.pinned
   const online = typeof numberOfPins === 'number'
+  const autoUpload = online ? await mfsPolicyEnableFlag(service.service, ipfs) : undefined
 
-  return { ...parsedService, numberOfPins, online }
+  return { ...parsedService, numberOfPins, online, autoUpload }
 }
 
 const mfsPolicyEnableFlag = memoize(async (serviceName, ipfs) => {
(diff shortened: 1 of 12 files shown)


src/bundles/pinning.js
```diff
@@ -103,35 +101,41 @@ const pinningBundle = {
...
-          if (!cidsToCheck.length) return // skip if no new cids to check
+          if (!cidsToCheck.length) continue // skip if no new cids to check
           const notPins = new Set(cidsToCheck.map(cid => cid.toString()))
-          const pins = ipfs.pin.remote.ls({ service: service.name, cid: cidsToCheck })
-          for await (const pin of pins) {
-            const pinCid = pin.cid.toString()
-            notPins.delete(pinCid)
-            adds.push({ id: `${service.name}:${pinCid}`, ...pin })
+          try {
+            // TODO: execute remote.ls with +1m backoff per service when response Type == "error" and Message includes "429 Too Many Requests"
```
lidel, line 123: :information_source: Curiously, hardcore caching is enough to send less than 6 requests per minute (I was not able to trigger rate-limiting at Pinata). 

If the user pins more than 6 things in one minute, they will get  a meaningful error via `notify.js` (yellow Toaster stating):

> Unable to set Pinning at Pinata: Error: reason: "Rate limit exceeded", details: "You have exceeded the API rate limit. Wait a minute and slow down your requests.": 429 Too Many Requests

lidel:
Fast-tracking this, as we need to  restore remote pinning on Pinata in go-ipfs 0.13 and ipfs-desktop 0.19.

CryptoUnchained (commented):
``

PR #1911, opened 2022-03-03
fix: fetch local pins stats

### LINKED ISSUE #1906: Settings show 0 local pins
i've recently uninstall/re-install IPFS Windows (Windows 10). 

i recall that there was a directory like "IPFS-Desktop" in my previous installation, but not the current one (see screenshot)
Untitled

i pinned the whole directory on my IPFS Windows and it shows ok on the Files page, but NOT on the Settings Page
Untitled

Is this normal or did it miss anything?

hacdias:
Fixes #1906: regression introduced in #1903, specifically by removing [local pin size calculation](https://github.com/ipfs/ipfs-webui/pull/1903/files#diff-7898dde76b2bbdcbb1d0fd5916ea316d6e7344901f01cbf5adb1822e140bc053), which is tied with the update of the amount of local pins.

Instead of just partially reverting the code, I added it to the pinning bundle and simplified, renamed and cleaned a few things.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -109,7 +109,7 @@
         "eslint-plugin-standard": "^4.0.1",
         "fake-indexeddb": "^3.1.2",
         "get-port": "^5.1.1",
-        "go-ipfs": "0.9.1",
+        "go-ipfs": "0.12.0",
         "http-proxy": "^1.18.1",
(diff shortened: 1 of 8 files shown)


lidel (approved):
Thanks!

PR #1916, opened 2022-03-23
fix: FilesExplorerForm icons for differing browser text sizes

jayschwa:
I run my browser with a non-standard text size and I noticed that the `FilesExplorerForm` icons do not look correct. This changes the icon heights to relative units from the absolute (default) pixels.

Very small and very large text sizes *before* this change:

very_small_broken
very_large_broken

After:

very_small_fixed
very_large_fixed

I am using the Dark Reader extension for the dark theme. It is not part of this PR.


Joelsamijon (approved):

lidel (approved):
Thanks!

I guess there is no harm in merging this (fwiw I see no regression with default font size)

PR #1919, opened 2022-04-08
feat: track remote pins in progress

### LINKED ISSUE #1752: Files: track pins in progress
Right now we do `pin.remote.add` with `background:true`.

It's fine for smaller files which get pinned pretty fast, but not really honest when it comes to bigger files.

We should:

- pin with `background:false` and wait until its pinned or throw
  - this should not be blocking UI, add ongoing pinning tasks to async queue
- mark remote pin as "in progress" and indicate that in UI  (could be "breathing" effect or something)
   - if long-running pin fails (throws) we should...  (TBD)
     - .. mark it as "failed" and indicate that in UI (icon tinted in red?)
     - clicking on failed cloud pin should open pin modal with (TBD.. failed item unchecked?)

### LINKED ISSUE #1991: Unable to set pinning at local: Timeout
Hi, I'm pretty new to this topic so please forgive me if the solution to this is obvious. I have downloaded the latest version of the desktop ui for windows (0.21.0) and wanted to download and share the wikipedia en snapshot. From my understanding I have to open the cid bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze in the explorer and pin it. I did this and the download starts, but after a while (about 20 minutes) I get an error message "Unable to set pinning at local: TimeoutError: Request timed out". What can I do about this? Restarting the pinning helps, but I can't do this every 20 minutes for the next week or so. Any ideas?

edit: using the cli seems to work, so the timeout must somehow come from the UI app

hacdias:
Closes #1752
Closes #1991

- [x] Persistance via `persistActions`
- [x] Breathing cloud when pending
- [x] Red cloud when failed
- [x] Periodic checks per storage with back-off
- [x] Dismiss failed pin when clicking red cloud (failed)
- [x] Replace `/pins` by a list of pending and done pin jobs that can be cleared (similar to the upload drawer)

There are some things you can test via the Storybook. However, there are some aspects, such as dismissing a failed pin on the Files page that I wasn't able to test.


hacdias:
Screenshot 2022-04-27 at 12 04 37

lidel (commented):
Thank you @hacdias!

I won't be able to look before I'm AFK for a week,  but  would be cool if @SgtPooki took it for a spin in the meantime, with a fresh set of eyes.  (not a full review, just try to use this and write down and comment on the UX)

lidel:
@hacdias mind resolving conflicst?  I think we should review and ship this, so this work is not lost.

SgtPooki:
@juliaxbow another one for you

SgtPooki:
I tried running this locally and got the following error:

image

hacdias:
@SgtPooki I think I fixed the issue you were having with the checkbox.

lidel (changes requested):
@hacdias mind rebasing/resolving conflicts again? I want to review it this/next week :)

lidel (changes requested):
Thank you @hacdias!

Will you be able to address two UX nits?
- There is no obvious UI for opening `/#/pins` screen to see ongoing jobs.
  Perhaps an easy fix for this is to add a breathing, clickable  cloud icon to the top, that would open Pins screen?
  > 
- There should be  a  ❌ button next to each ongoing pin job, allowing the user to cancel every job (remove the pin), not only the failed ones.

hacdias:
Regarding the UI element to navigate to the Pins page: I would like to add something that would fit in. However, I'm not having great ideas.

image

**Problem**: the remanning information blocks on this style are not clickable.

image

**Problem**: better, as the button indicates an action. However, we do not usually use buttons for navigation, so it also does not feel natural.

image

I pushed your suggestion, but it doesn't make me necessarily happy.

image

We could also keep it minimal without the button surrounding it. I just feel like it's still a bit out of place too.

What do you think?

hacdias:
Tweaked the interface to include an 'x' for all types of pins:

image

lidel:
Thanks @hacdias!  

@juliaxbow any suggestions on what could be the best way to indicate "there are remote pins in progress"? 
Design requirement: we want it to be visually noticeable, so the user won't miss the fact there are pinning jobs in progress.
See initial mockups in https://github.com/ipfs/ipfs-webui/pull/1919#issuecomment-1202685324

juliaxbow:
I like the idea of a breathing cloud! Something like below. Tested this on two people who both were inclined to click the cloud
Pinning Cloud

Separately I would also suggest an additional visual clue (aside from color) for failed pin to make it more accessible for users. 
Like this
Broken Pin
or the more straightforward version (which I prefer):
Failed Pin

hacdias:
I updated the breathing cloud to match your idea:

https://user-images.githubusercontent.com/5447088/183380046-1a2b69e6-5e6a-4d29-8ffe-6dd5c00c3635.mov

Regarding the icon for the failed pins: I think it'd be a great addition, but maybe we should add it to https://github.com/ipfs-shipyard/ipfs-css first. /cc @lidel

src/files/header/BreathingCloud.css
```diff
@@ -0,0 +1,21 @@
+.BreathingCloud {
```
lidel, line 1: I've just realized that having two different animations for indicating the same thing using the same icon  looks a bit weird.

Can we replace this with the more subtle one  from the File list? (without drop shadow)

>

src/pins/PinsStatuses.js
```diff
@@ -0,0 +1,57 @@
...
+        <GlyphCancel className='fill-aqua' />
+      </button>
+    </span>
+  </li>)
+}
+
+const PinsStatuses = ({ pendingPins, failedPins, completedPins, onDismissFailedPin, onDismissCompletedPin, doCancelPendingPin, t }) => {
+  return (
+    <div>
+      <h3>Pins</h3>
```
lidel, line 46: This should be added to locale files and have two values:
- "All done, no pins in progress" (when list is empty)
- "Pinning in progress"

lidel (changes requested):
Nice!

- If you have time, adding new icons to it  to ipfs-css is :+1: 
  - .. but if you use the same icon for the failed pins, and  its color to red (`fill-red`) its also fine.
- I like the idea of a breathing cloud. Small nit here: see suggestions to reuse existing animation below.

src/pins/PinsStatuses.js
hacdias: @lidel you mean, add local pins to this page, right?

hacdias:
@juliaxbow could you please send the bottom icon (cloud with the slash) as an SVG so I can add it to the ipfs-css repository? You can see how the other icons are formatted and sizes, here: https://github.com/ipfs-shipyard/ipfs-css/tree/main/icons

src/pins/PinsStatuses.js
lidel: I meant translation files https://github.com/ipfs/ipfs-webui/tree/main/public/locales/en :)

hacdias: Oeh, done 😃

juliaxbow:
@hacdias added [here](https://github.com/ipfs-shipyard/ipfs-css/compare/main...juliaxbow:ipfs-css:patch-1?short_path=e22673d#diff-e22673d5849f2637ae9f3e1215a565b15cd7e685813e0a6d9ba3aee503c1adfb)

hacdias:
I added the new icon in #2008 @lidel @SgtPooki. Please review and merge that PR into this one.

hacdias:
@lidel @SgtPooki this PR is ready for review. It already includes the new icon.

src/bundles/pinning.js
```diff
@@ -55,23 +62,83 @@ const uniqueCidBatches = (arrayOfCids, size) => {
...
+
+  const { lastTry, tryAfter } = backoffs[service] || { lastTry: 0, tryAfter: PIN_CHECK_INTERVAL }
+  if (lastTry + tryAfter > new Date().getTime()) {
+    throw new Error('still within back-off period')
+  }
+
+  try {
+    return ipfs.pin.remote.ls(params)
+  } catch (e) {
+    if (e.toString().includes('429 Too Many Requests')) {
```
SgtPooki, line 77: Some logic in the pinning service compliance checker shows how to read the headers and automatically await the appropriate time (retry header, tokens available, etc.) I believe its in the middleware.

Instead of reaching the point of failure and catching it, we should be preventing the exhausting of the available calls from the get-go.

src/bundles/pinning.js
```diff
@@ -55,23 +62,83 @@ const uniqueCidBatches = (arrayOfCids, size) => {
...
+      backoffs[service] = {
+        lastTry: new Date().getTime(),
+        tryAfter: tryAfter * 3
```
SgtPooki, line 80: We should re-use the logic in pinning service compliance checker to prevent a 3x increase in retry times.

src/bundles/pinning.js
```diff
@@ -55,23 +62,83 @@ const uniqueCidBatches = (arrayOfCids, size) => {
...
+const resumePendingPins = (store) => {
+  const interval = setInterval(() => {
+    const isReady = store.selectIpfsReady()
+    if (isReady) {
+      clearTimeout(interval)
+      const pendingPins = store.selectPendingPins()
+
+      pendingPins.forEach(pin => {
+        const [service, cid] = pin.split(':')
+        console.log(service, cid)
```
SgtPooki, line 98: Remove if not using dev/preference logging.

src/bundles/pinning.js
```diff
@@ -55,23 +62,83 @@ const uniqueCidBatches = (arrayOfCids, size) => {
...
+    }
+  }, 1000)
+}
+
+const intervalFetchPins = (store) => {
+  setInterval(() => {
+    const pins = [
+      ...store.selectPendingPins(),
+      ...store.selectFailedPins()
+    ].map(serviceCid => ({ cid: serviceCid.split(':')[1] }))
```
SgtPooki, line 111: This service:cid is an implicit convention that could be better if defined.

```
/**
 * simple link between a cid and a pinning service with which the cid is intended to be pinned with.
 */
class PinningLink {

  constructor (service, cid){}
  get service(){}
  get cid(){}
  toString(){}
}

```

src/bundles/pinning.js
```diff
@@ -4,16 +4,23 @@ import memoize from 'p-memoize'
...
 const CID_PIN_CHECK_BATCH_SIZE = 10 // Pinata returns error when >10
 
+const PIN_CHECK_INTERVAL = 30000
+
 // id = `${serviceName}:${cid}`
 const cacheId2Cid = (id) => id.split(':').at(-1)
 const cacheId2ServiceName = (id) => id.split(':').at(0)
 
+const uniq = (arr) => [...new Set(arr)]
+const notIn = (...arrays) => p => arrays.every(array => !array.some(a => a === p))
```
SgtPooki, line 22: These should be in `src/hof/arrays.js` or similar

SgtPooki (changes requested):
Partial review

src/bundles/pinning.js
```diff
@@ -222,6 +317,25 @@ const pinningBundle = {
...
+
+  doDismissFailedPin: (...pins) => async ({ dispatch }) => {
+    dispatch({ type: 'DISMISS_REMOTE_PINS', payload: { failed: pins } })
+  },
+
+  doCancelPendingPin: (...pins) => async ({ getIpfs, dispatch }) => {
+    const ipfs = getIpfs()
+
+    for (const pin of pins) {
+      const [service, cid] = pin.split(':')
```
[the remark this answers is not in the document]
hacdias: Tracked on https://github.com/ipfs/ipfs-webui/issues/2042

src/bundles/pinning.js
hacdias: Tracked in #2043

hacdias:
@SgtPooki I created some follow up issues (#2043 #2042) according to your comments. Once CI is green, I will be merging this.

hacdias:
@SgtPooki it seems I need a new approval for this to go through 😄

SgtPooki (approved):

hacdias:
I do not understand. There are no pending changes requested, but I still can't merge.

hacdias:
Ah, @lidel requested changes at some point so we also need @lidel's review.

PR #1934, opened 2022-05-19
fix(ci): use the latest ipfs-cluster-ctl

lidel:
It seems that CI was using cached ipfs-cluster-ctl v0.14.5, and the latest one is v1.0.1 which had breaking change around pinsets.

Let's see if this solves the issue.


lidel:
ok, bumping cache fixed it, the github action is fetching the latest version (https://github.com/ipfs/ipfs-webui/runs/6515537924?check_suite_focus=true#step:6:42)

PR #1945, opened 2022-06-09
fix: use gateway links to download standalone files

hacdias:
In #1894 we tried to address #1887. However, doing that for all types of files is [blocked](https://github.com/ipfs/ipfs-webui/pull/1894#discussion_r786341423) by https://github.com/ipfs/go-ipfs/issues/7746. This PR does the same but only for standalone files. I think it's a good improvement and can mitigate the issues that we are trying to address.

I also don't understand why the CI is failing on "Installing Dependencies" as this PR changes nothing regarding dependencies.

### DIFF
--- a/src/files/FilesPage.js
+++ b/src/files/FilesPage.js
@@ -62,10 +62,17 @@ const FilesPage = ({
       return downloadAbort()
     }
 
-    const updater = (v) => setDownloadProgress(v)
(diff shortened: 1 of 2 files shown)


lidel (approved):
Good call! :+1: 

(CI got fixed in main branch, I updated this PR and all green now)

Issue #1951, opened 2022-06-20
UX Bug: Add folder action is hidden behind "import" dropdown button


SgtPooki:
**Describe the bug**
"New folder" behind "Import" doesn't make sense. You can't import something that doesn't exist yet, but you can create it. "Add" instead of "Import" makes the most sense.

**To Reproduce**
1. Go to Files screen
2. Click on 'Import'
3. Notice that 'New Folder' is listed as an option under "Import"

**Expected behavior**

**Screenshots**
<img width="262" alt="image" src="https://user-images.githubusercontent.com/1173416/174672920-f99e77b6-f378-439e-9e46-d1fe97c15192.png">

**Desktop (please complete the following information):**
IPFS Desktop on macOS Monterey 12.4 (21F79)

[UI v2.15.0](https://github.com/ipfs/ipfs-webui/releases/tag/v2.15.0)
[Revision fadd900](https://github.com/ipfs/ipfs-webui/commit/fadd900)
[See the code](https://github.com/ipfs/ipfs-webui)
[Report a bug](https://github.com/ipfs/ipfs-webui/issues)

**Additional context**
N/A

SgtPooki:
@juliaxbow, we discussed this in our GUI triage meeting, and @lidel mentioned that "Import" was added in the past because it made the most sense for all of the options other than "New Folder". What do you think makes the most sense for this screen?

juliaxbow:
@SgtPooki I'd recommend switching over to "Add," as suggested, or "New," which you'll see on some other file sharing platforms. I'm leaning towards "New" but defer to you on current user flow.  Let me know if you want me to conduct some quick tests.

I would also recommend:
- Modifying icons slightly to show upload vs new folder creation
- Rename actions for clarity (e.g. "File Upload" or "File Import" vs "File")
- Adding a line to differentiate between actions (upload vs create)

Screenshot attached for reference.
<img width="833" alt="Import Dropdown" src="https://user-images.githubusercontent.com/58958327/177883083-ac8911e9-f78a-4d13-b9d1-0b734f6f1f40.png">

SgtPooki:
@lidel what do you think about the proposals from Julia?

lidel:
I think both are :+1: thank you @juliaxbow!

-  The only nit around "Add" one is  that "IPFS Upload" should be changed to "IPFS Import"  (the data is already on IPFS, nothing to upload). If we fix that, it will look better than "Import" three times.
[...]

PR #1952, opened 2022-06-20
fix: change "import" button label to "add" on files screen

brainsaysno:
Changed "import" button label to "add" on files screen to better fit new folder option. As suggested by #1951.


SgtPooki:
@brainsaysno thanks so much for getting this out so quickly. We discussed this in our GUI triage meeting and there seems to be a use-case for leaving import, but I think we agreed on adding a "Add new folder" button or similar. I'll update #1951

lidel (changes requested):
@brainsaysno  do you mind updating this PR with labels proposed in  https://github.com/ipfs/ipfs-webui/issues/1951#issuecomment-1178331128 and https://github.com/ipfs/ipfs-webui/issues/1951#issuecomment-1209834082?   :pray:

PR #1957, opened 2022-06-28
feat: automate releases

SgtPooki:
* split github workflows so they're composable
* automatically update changelog on release
* Create .car file from dag export of CID
* publish .car file to release artifacts
* integrate with semantic-release, modifying changelog entries via `custom-release-notes-generator.js`, automatically adding CID


SgtPooki:
@lidel , I played around with this a bit longer than I planned yesterday, trying to get CID during changelog process, removing all manual steps. I think if we use a CAR with a single CID, we could remove the CID from the release notes, and then auto-generate them easier. what are your thoughts?

SgtPooki:
@lidel will do, but I think this update would be best decided after we discuss how to handle CIDs in the changelog. I moved this PR to a draft. 

1. If we have a single CID that is always the latest webui release, then we don't need to copy the CID into release notes each time and can simply put it at the top of the README (or hardcode into changelog template). 
2. If we don't, then adding the CID to release notes would require plugging into the existing workflows when generating the changelog (also do-able)

lidel:
We should continue documenting CID for every release (we've been doing that for past releases).

SgtPooki:
i've decided to go an entirely better direction and use semantic-release for this since aegir and everything else is using it.

I've got the CID automatically adding to the release-notes (which semantic-release should add to the changelog automatically) via `npx semantic-release -b feature/add-changelog --dry-run --no-ci  --debug`:

image

.github/workflows/publish-release-build.yml
```diff
@@ -51,3 +51,10 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           args: "./ipfs-webui.tar.gz"
+
+      - name: Attach .car file to Release Notes
+        uses: skx/github-action-publish-binaries@c881a3f8ffb80b684f367660178d38ceabc065c2 #release-0.15
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          args: "${{ github.workspace }}/.cache/ipfs-webui@${{ github.sha }}.car"
```
lidel, line 60: :warning:   you need to update `key` and `restore-key`  in `.github/workflows/publish-release-build.yml` to also  use `${{ runner.os }}-${{ github.sha }}`

.github/workflows/ci.yml
```diff
@@ -132,6 +76,9 @@ jobs:
 
       - run: echo ${{ steps.ipfs.outputs.cid }}
       - run: echo ${{ github.ref }}
+      - name: Create ipfs-webui.car file
+        run: |
+          ipfs dag export ${{ steps.ipfs.outputs.cid }} > ${{ github.workspace }}/.cache/ipfs-webui@${{ github.sha }}.car
```
lidel, line 81: nit: Is it still true that `@` is forbidden in filenames on Windows? `_` may be safer here:

```suggestion
          ipfs dag export ${{ steps.ipfs.outputs.cid }} > ${{ github.workspace }}/.cache/ipfs-webui_${{ github.sha }}.car
```

lidel (approved):
Thank you @SgtPooki! 

This approach looks sensible, ok to merge as long you address comments below.

.github/workflows/ci.yml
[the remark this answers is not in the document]
SgtPooki: good catch

PR #1958, opened 2022-06-29
fix: updated ipfs-geoip (2022-06-28)

lidel:
This PR updates ipfs-geoip dependency to the latest version, which brings the latest geoip dataset (`GeoLite2-City-CSV_20220628`)
The old one was 2 years old (!). – see https://github.com/ipfs-shipyard/ipfs-geoip/pull/88 for more details.

This should fix bugs  on Peers screen where it currently lists faster-than-light nonsense  like 8ms from Europe to USA ;-) 

>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
         "intl-messageformat": "^9.11.3",
         "ip": "^1.1.5",
         "ipfs-css": "^1.3.0",
-        "ipfs-geoip": "^7.0.0",
+        "ipfs-geoip": "^8.0.0",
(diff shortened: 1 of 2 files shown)


lidel:
@SgtPooki seems that `package-lock.json` changes break e2e tests both here and in #1957.
This PR  needs to be rebased after we figure it out.

lidel:
I've re-generated `package-lock.json` with Node v16.12.0, let's see if that helps.

SgtPooki (approved):

PR #1962, opened 2022-07-05
feat: test npm build&install against LTS node versions

SgtPooki:
this new workflow also publishes shields, which would then be published on our README.

### README update:
image

### workflow results (via https://github.com/SgtPooki/ipfs-webui/actions/runs/2605014389)

image


.github/workflows/node-test.yml
```diff
@@ -0,0 +1,92 @@
...
+    steps:
+      - name: "Reset ${{ inputs.gh-node-version }} badge"
+        uses: RubbaBoy/BYOB@v1.2.1
+        with:
+          ICON: https://raw.githubusercontent.com/devicons/devicon/master/icons/nodejs/nodejs-original.svg
+          NAME: "node-${{ inputs.gh-node-version }}"
+          LABEL: "${{ inputs.gh-node-version }}"
+          STATUS: "Building..."
+          COLOR: grey
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
SgtPooki, line 26: this may not be necessary, I'm open to removing it.

.github/workflows/node-versions.yml
```diff
@@ -0,0 +1,39 @@
...
+    uses: ./.github/workflows/node-test.yml
+    secrets: inherit
+    with:
+      gh-node-version: '16.x'
+
+  test-node-18:
+    uses: ./.github/workflows/node-test.yml
+    secrets: inherit
+    with:
+      gh-node-version: '18.x'
```
SgtPooki, line 39: A few questions here:

1. Do we want to test node 10.x? it's pretty old...
2. Do we know of a method to test `17.x`, or other common, but non-LTS, versions? I did a (very) quick search and didn't find anything.
3. Do we want to show node versions badge successf/fail in an aggregated manner?
   * i.e. if nodev14 only fails on windows, should the badge be red? If not, the work involved in that is significant, so maybe a separate issue could track that decoupling.

README.md
```diff
@@ -22,7 +22,7 @@ The app is built with [`create-react-app`](https://github.com/facebook/create-re
 
 ## Install
 
-With `node` >= 10 and `npm` >= 6.4.1 installed, run
+[![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-10.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml) [![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-12.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml) [![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-14.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml) [![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-16.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml) [![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-18.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml)
```
SgtPooki, line 25: I tried a few variations of displaying the results, and this seems to be the cleanest. I'm open to ideas.

package.json
```diff
@@ -138,6 +139,7 @@
     "jest-playwright-preset": "^1.7.0",
     "jest-process-manager": "^0.3.1",
     "multihashing-async": "^1.0.0",
+    "node-pre-gyp": "^0.17.0",
```
SgtPooki, line 142: node-pre-gyp is deprecated -- that's why `@mapbox/node-pre-gyp` was added -- but this repo wouldn't build on github-actions without it. not sure if you have a known fix for this.

.github/workflows/node-test.yml
```diff
@@ -0,0 +1,92 @@
...
+      id: cache
+      with:
+        path: ${{ github.workspace }}/.cache
+        key: ${{ runner.os }}-${{ inputs.gh-node-version }}-${{ hashFiles('package.json', 'package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-${{ inputs.gh-node-version }}-${{ hashFiles('package.json', 'package-lock.json') }}
+          ${{ runner.os }}-${{ inputs.gh-node-version }}
+
+    - name: Install dependencies
+      run: npm ci --prefer-offline --no-audit --progress=false --cache ${{ github.workspace }}/.cache/npm
```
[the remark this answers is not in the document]
lidel: `npm ci` provides more determinism and is pretty fast. Unless you find something faster, I'd keep it and update README to use it too.

.github/workflows/node-versions.yml
lidel: 1.  let's remove `10.x` 
   - when we make things build on  current LTS (`16.x`) we should also remove `12.x` and `14.x`
2. I think you can use  `lts/*` and `current`  to always run against latest versions from  https://nodejs.org
3.  No need to be that verbose, if job fails then one can go to action and see what failed.

README.md
lidel: Looks good, maybe add  context behind the badges:

```suggestion
With a supported Node.js version: [![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-10.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml) [![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-12.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml) [![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-14.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml) [![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-16.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml) [![](https://byob.yarr.is/SgtPooki/ipfs-webui/node-18.x)](https://github.com/SgtPooki/ipfs-webui/actions/workflows/node-versions.yml)
```

package.json
lidel: The proper fix is to update all dependencies so `node-pre-gyp` is no longer pulled-in. Unfortunately, for that we need to update js-ipfs dependencies to the latest version: https://github.com/ipfs/ipfs-webui/pull/1840

Using `@mapbox/node-pre-gyp` here sounds like a good band-aid until that happens :+1:

.github/workflows/node-test.yml
```diff
@@ -0,0 +1,92 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ inputs.gh-node-version }}
+  cancel-in-progress: true
+
+jobs:
+  reset-badge:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Reset ${{ inputs.gh-node-version }} badge"
+        uses: RubbaBoy/BYOB@v1.2.1
```
lidel, line 19: nit: since this is a third-party action, we should replace tag with git commit in all places it is used

lidel (approved):
I think one badge per nodejs version is more than enough. 

Small suggestions in comments inline, but feel free to merge either way.

PR #1932, opened 2022-05-19
fix: package.json & package-lock.json to reduce vulnerabilities

Vacryptogodfather45:
The following vulnerabilities are fixed with an upgrade:
- https://snyk.io/vuln/SNYK-JS-ANSIREGEX-1583908

PR #1969, opened 2022-07-09
chore(deps): update react-scripts

SgtPooki:
This is a replacement for #1932

github merge instructions are borked for CLI.. I didn't guess pushing to main would automatically update the PR.. and it didn't. it pushed to main, and closed the PR.

I reverted those changes and updated many things necessary for upgrading react-scripts. ~~Build finally succeeds, but tests are failing.~~ Build, tests, and CI checks are all passing


SgtPooki:
Kicked off a test of all node versions, https://github.com/ipfs/ipfs-webui/actions/runs/2709349231, so we can see how these changes impact builds across node versions and platforms.

lidel (approved):
lgtm,  i wonder if Node 18 build could work if we pass `--openssl-legacy-provider` (but that can be separate PR)

PR #1974, opened 2022-07-11
fix: resolve issues with automated releases 

SgtPooki:
After enabling automated releases for this repo, I noticed a number of issues:

1. The car file was not attached to the GitHub release notes
2. Releases are generated for every single commit, causing release notes to become useless. We should batch commits and release only after a minimum wait time 
   * I'm going to revert to creating a release manually, except all that will be required is running https://github.com/ipfs/ipfs-webui/actions/workflows/ci.yml (i.e. `workflow_dispatch`). This is still a significant improvement upon the previous manual process
3. Even though semantic-release doesn't publish to npm currently (we have it disabled), the files being added to the npm tarball were unnecessary (.cache/*, and many others. You can see this on old runs). This was making the logs extremely difficult even to scroll through.
   * Specifying `"files":` in the package.json allows us to limit which files are included.
5. PRs weren't testing the 'semantic-release' part of the `ci.yml` workflow, so everything in the ci.yml workflow could go great, and then fail during semantic-release job.
   * Now, semantic-release is run with a dry-run flag for PRs

Also, see https://github.com/ipfs/ipfs-webui/issues/1970

This PR attempts to resolve all of the above issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,7 +229,7 @@ jobs:
 
       # Update the version (npm version [major|minor|patch])
       - name: Run semantic release
-        if: github.ref == 'refs/heads/main'
+        if: github.ref == 'refs/heads/main' && github.event_name == 'workflow_dispatch'
         run: |
           git config user.name "ipfs-gui-bot"
           git config user.email "108953096+ipfs-gui-bot@users.noreply.github.com"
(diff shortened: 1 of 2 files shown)


SgtPooki:
@achingbrain I was told by @biglep that you've got a release setup where a PR is opened, and then added to as commits are merged into the main branch, and then you kick of a release just by merging that "release" PR. Is that correct? 

If so, I would love to get that set up here. Can you point me in the right direction?

hacdias (approved):
LGTM. I also agree that this is an improvement compared to what we had, and it's also better than releasing a version with every single commit.

Issue #1981, opened 2022-07-20
chore: install missing dependencies


SgtPooki:
```bash
╰─ ✘ 1 ❯ npx --yes npm-check | grep 'Not in the package.json.'
⠙ Checking for unused packages. --skip-unused if you don't want this.(node:57304) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at /Users/sgtpooki/code/work/protocol.ai/ipfs/webui/node_modules/@playwright/test/package.json.
Update this package.json to use a subpath pattern like "./*".
(Use `node --trace-deprecation ...` to show where the warning was created)
(node:57304) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at /Users/sgtpooki/code/work/protocol.ai/ipfs/webui/node_modules/playwright-chromium/package.json.
Update this package.json to use a subpath pattern like "./*".
eslint-config-react-app                     😟  PKG ERR!  Not in the package.json. Found in: /.eslintrc.js
@typescript-eslint/parser                   😟  PKG ERR!  Not in the package.json. Found in: /.eslintrc.js
@typescript-eslint/eslint-plugin            😟  PKG ERR!  Not in the package.json. Found in: /.eslintrc.js
expect-playwright                           😟  PKG ERR!  Not in the package.json. Found in: /test/e2e/setup/global-after-env.js, /test/e2e/setup/test-environment.js
```

SgtPooki:
```bash
❯ npm install -D eslint-config-react-app @typescript-eslint/parser @typescript-eslint/eslint-plugin expect-playwright
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR!
npm ERR! While resolving: ipfs-webui@2.17.3
npm ERR! Found: @typescript-eslint/eslint-plugin@4.15.2
npm ERR! node_modules/@typescript-eslint/eslint-plugin
npm ERR!   peer @typescript-eslint/eslint-plugin@"^4.0.0" from eslint-config-react-app@6.0.0
npm ERR!   node_modules/eslint-config-react-app
npm ERR!     eslint-config-react-app@"^6.0.0" from react-scripts@4.0.3
npm ERR!     node_modules/react-scripts
npm ERR!       react-scripts@"^4.0.3" from the root project
npm ERR!     dev eslint-config-react-app@"*" from the root project
[...]

PR #1983, opened 2022-07-20
refactor(test): switch e2e to @playwright/test

SgtPooki:
Potentially helps with  #1981
Unblocks https://github.com/ipfs/ipfs-webui/issues/1804


lidel (approved):
I took a stab at fixing E2E tests in this PR  and ended up removing jest-playwright and replacing it with @playwright/test (we did something similar in ipfs-desktop a while ago):

> 

It simplified things a lot, and restored CID to be green again.  
Updated the title to this PR to reflect  the change, and decouple it from #1981

Issue #1940, opened 2022-06-06
Can't build with nodejs v17 and npm v8


Jorropo:
**Version**
e77afcbdfbfd7110a9e9ca972e82afb114d273a0

**Describe the bug**
The README specify

> With `node` >= 10 and `npm` >= 6.4.1 installed, run

https://github.com/ipfs/ipfs-webui#install
But with :
```
$ npm --version
8.12.1
$ node --version
v17.3.1
```
I get:
```
$ npm i
npm WARN ERESOLVE overriding peer dependency
npm WARN While resolving: eslint-plugin-ava@12.0.0
npm WARN Found: eslint@7.21.0
npm WARN node_modules/eslint
npm WARN   peer eslint@"^5.0.0 || ^6.0.0 || ^7.0.0" from @typescript-eslint/eslint-plugin@4.15.2
npm WARN   node_modules/@typescript-eslint/eslint-plugin
npm WARN     peer @typescript-eslint/eslint-plugin@"^4.0.0" from eslint-config-react-app@6.0.0
npm WARN     node_modules/eslint-config-react-app
npm WARN     1 more (react-scripts)
npm WARN   18 more (@typescript-eslint/experimental-utils, ...)
npm WARN 
npm WARN Could not resolve dependency:
npm WARN peer eslint@">=7.22.0" from eslint-plugin-ava@12.0.0
npm WARN node_modules/eslint-plugin-ava
npm WARN   eslint-plugin-ava@"^12.0.0" from ipfs-repo@12.0.0
npm WARN   node_modules/ipfs-repo
npm WARN 
npm WARN Conflicting peer dependency: eslint@8.17.0
npm WARN node_modules/eslint
npm WARN   peer eslint@">=7.22.0" from eslint-plugin-ava@12.0.0
npm WARN   node_modules/eslint-plugin-ava
npm WARN     eslint-plugin-ava@"^12.0.0" from ipfs-repo@12.0.0
npm WARN     node_modules/ipfs-repo
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: ipld-explorer-components@2.0.4
npm ERR! Found: i18next@21.6.10
npm ERR! node_modules/i18next
npm ERR!   i18next@"^21.6.10" from the root project
npm ERR!   peer i18next@">= 19.0.0" from react-i18next@11.15.3
npm ERR!   node_modules/react-i18next
[...]

PR #1986, opened 2022-07-21
docs(readme): NodeJS version support info

SgtPooki:
See #1940 and https://github.com/ipfs/ipfs-webui/issues/1940#issuecomment-1179603709 for more information

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,13 +29,9 @@ With a supported Node.js version: [![](https://byob.yarr.is/ipfs/ipfs-webui/node
 ```
 
 #### A Note on NodeJS support
-We will only support *even numbered* nodejs versions that are classified in one of the following release statuses:
+We aim to support ipfs-webui development and build with  "Current" and "Active LTS"  Nodejs versions.
 
-* Current (Support = P1)
-* Active LTS (Support = P0)
-* Maintenance LTS (Support = P2)
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +28,15 @@ With a supported Node.js version: [![](https://byob.yarr.is/ipfs/ipfs-webui/node
...
 
+#### A Note on NodeJS support
+We will only support *even numbered* nodejs versions that are classified in one of the following release statuses:
+
+* Current (Support = P1)
+* Active LTS (Support = P0)
+* Maintenance LTS (Support = P2)
+
+See https://nodejs.org/en/about/releases/ for more information about which versions have which release statuses.
+
```
lidel, line 39: I don't think there is any value in supporting legacy LTS

```suggestion
#### A Note on NodeJS support
We aim to support ipfs-webui development and build with  "Current" and "Active LTS"  Nodejs versions.

See <https://nodejs.org/en/about/releases/> for more information about which versions have which release statuses.

```

lidel (approved):
lgtm, but let's simplify this – it is not a library that needs to support wide number of nodejs releases, and this is best-effort anyway.

SgtPooki:
@lidel this is intended to be linked to from ipfs-webui, ipfs-desktop, ipld-explorer-components, etc..

PR #1998, opened 2022-08-04
feat: ux improvements to publish modal

### LINKED ISSUE #1993: Improve UX of IPNS publishing
Now that we've added basic UI for publishing CIDs to IPNS (#1973) :partying_face: 
it may be a good time to go an extra mile and refine the UI.

Namely, I feel we could improve presentation of key names and CIDs (right now it is a bland monospace font).
Perhaps we could also add one-line explainer to the first modal (imagine user clicking "Publish to IPNS" and not knowing what it is):

Below is the current state:  

| Before Publishing |  Published | 
| ---- | ---- |
|   |  |

any ideas are welcome, cc @juliaxbow @hacdias

hacdias:
Closes #1993

When it finishes publishing:

image

After clicking:

image

Then the icon goes back to normal after 2 seconds.

### DIFF
--- a/public/locales/en/app.json
+++ b/public/locales/en/app.json
@@ -30,7 +30,8 @@
     "unselectAll": "Unselect all",
     "generate": "Generate",
     "publish": "Publish",
-    "downloadCar": "Download as CAR"
+    "downloadCar": "Download as CAR",
+    "done": "Done"
   },
   "cliModal": {
     "description": "Paste the following into your terminal to do this task in IPFS via the command line. Remember that you'll need to replace placeholders with your specific parameters."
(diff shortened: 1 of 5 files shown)


lidel:
cc @juliaxbow for :eyes:

juliaxbow:
Minor changes to the success state after clicking the copy icon: 
- Icon: padding to the right of the success icon should be increased by 3px to match copy icon formatting
- Link field: opacity of text or fill of #ffffff at 60% opacity should occur alongside the success message so that it is more legible
- Success message: I think the tone of "Link copied!" over "Link copied" fits better
Screen Shot 2022-08-04 at 1 20 23 PM

hacdias:
I addressed @juliaxbow's comments and removed the 'Cancel' button as it does not do anything different than 'Done' since it's already published.

https://user-images.githubusercontent.com/5447088/183389185-6d898e51-a98a-48b1-a84f-ef7783f00abc.mov

lidel (approved):
:ok_hand:

PR #2005, opened 2022-08-16
fix: korean detection

### LINKED ISSUE #1999: Attempted to resolve locale that does not exist
**Describe the bug**
The WebUI attempts to resolve a locale that does not exist. Eventually, WebUI rendering fails.

**To Reproduce**
This issue can be seen by accessing the WebUI from a non-English desktop.

**Expected behavior**
If the locale does not exist, it should be automatically resolved as `en`.

**Screenshots**
image

image

 - OS: Windows 10
 - Browser: Chrome
 - Version: 103

**Additional context**

Nothing

hacdias:
This should fix #1999. I just wonder if this is not something that happens in other languages too. For example pt -> PT-PT or pt -> PT-BR.

### DIFF
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -39,6 +39,7 @@ i18n
       'zh-Hans': ['zh-CN', 'en'],
       'zh-Hant': ['zh-TW', 'en'],
       zh: ['zh-CN', 'en'],
+      ko: ['ko-KR', 'en'],
       default: ['en']
     },
     debug: process.env.DEBUG,


lidel (approved):
I think this solves problem at hand. 
If we see it returning for other locale, we will replace `fallbackLng` with a fancy function described [here](https://www.i18next.com/principles/fallback#fallback-to-different-languages) to fix this in more generic fashion.

PR #2026, opened 2022-09-14
feat: add Web3.Storage and Estuary to default pinning provider list

SgtPooki:
- feat: add Web3.Storage default pinning provider
- feat: add Estuary default pinning provider

image

PR #2027, opened 2022-09-15
fix(settings): randomize pinning service templates

lidel:
Now that we have 4+ services (https://github.com/ipfs/ipfs-webui/pull/2026) we should always show them in random order, so we do not promote any specific one above others.

This PR:
-  adds a simple shuffle that ensures random order across app loads
- fixes `service endpoint should be provided without the /pins suffix` error for templates added in #2026

### DIFF
--- a/src/constants/pinning.js
+++ b/src/constants/pinning.js
@@ -2,6 +2,8 @@
 // This is a list of predefined templates for popular services from the IPFS
 // community.  We are open to reviewing PRs that add more entries here,
 // but only well-established and mission-aligned services will be accepted.
+// Services listed here are returned in a random order to ensure UI does not
+// promote any of them more than others.
 
 const complianceReportsHomepage = 'https://ipfs-shipyard.github.io/pinning-service-compliance'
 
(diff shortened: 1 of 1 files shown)


src/constants/pinning.js
```diff
@@ -33,21 +35,21 @@ const pinningServiceTemplates = [
   {
     name: 'Web3.Storage',
     icon: 'https://dweb.link/ipfs/bafybeiaqsdwuwemchbofzok4cq7cuvotfs6bgickxdqr6f7hdt7a64cwwa/Web3.Storage-logo.svg',
-    apiEndpoint: 'https://api.web3.storage/pins',
+    apiEndpoint: 'https://api.web3.storage',
```
SgtPooki, line 38: oops.

lidel:
> There is an argument to be made for not promoting one over the other, but I think UI consistency is more important.

This randomization happens on page load, and not every time on modal open.
This means the user will see the same order when they add multiple services and open the modal multiple times. 

Lmk if you feel this is not enough. Alternatively, I could sort them in a deterministic way, e.g. based on PeerID.
This way every user will always get the same order, but it will be randomized across users, so no service will be promoted as the default (first).

> [..] removing services that you've already added. For example, I already have Pinata added to my pinning services, with API_TOKEN, but it still shows in the list of providers to add.

This is bad UX: user should be able to use multiple tokens against the same endpoint.

SgtPooki (approved):

PR #2032, opened 2022-09-27
ci: made the kubo version param CI workflow uses configurable

galargh:
I propose to make kubo version used during CI workflow run configurable.

This would enable running the CI workflow for a specific kubo release. In particular, we could include a testing step in the release checklist that asks the release lead to kick off CI workflow in `ipfs-webui` to test if the new kubo release is OK.

###### Testing
- https://github.com/ipfs/ipfs-webui/actions/runs/3135588411

_This is related to https://github.com/ipfs/kubo/issues/9237_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -226,7 +226,7 @@ jobs:
           name: ipfs-webui_${{ github.sha }}.car
 
       - name: Dry-run semantic release
-        if: github.ref != 'refs/heads/main' || github.event.inputs.kubo-version
+        if: github.ref != 'refs/heads/main' && !github.event.inputs.kubo-version
         run: |
           git config user.name "ipfs-gui-bot"
           git config user.email "108953096+ipfs-gui-bot@users.noreply.github.com"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,6 @@ jobs:
       - uses: ipfs/download-ipfs-distribution-action@v1
         with:
           name: ipfs-cluster-ctl
-          version: ${{ github.event.inputs.ipfs-cluster-ctl-version }}
       - name: Fix DNS resolver
         run: |
           # fix resolv - DNS provided by Github is unreliable for DNSLik/dnsaddr
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,6 @@ on:
       kubo-version:
         description: Kubo version to use during the run
         required: false
-      ipfs-cluster-ctl-version:
-        description: IPFS Cluster CTL version to use during the run
-        required: false
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -221,7 +226,7 @@ jobs:
           name: ipfs-webui_${{ github.sha }}.car
 
       - name: Dry-run semantic release
-        if: github.ref != 'refs/heads/main'
+        if: github.ref != 'refs/heads/main' || github.event.inputs.kubo-version
         run: |
           git config user.name "ipfs-gui-bot"
           git config user.email "108953096+ipfs-gui-bot@users.noreply.github.com"
```
galargh, line 232: Should we be skipping this step as well? It failed during my test run: https://github.com/ipfs/ipfs-webui/actions/runs/3135588411

.github/workflows/ci.yml
```diff
@@ -221,7 +226,7 @@ jobs:
           name: ipfs-webui_${{ github.sha }}.car
 
       - name: Dry-run semantic release
-        if: github.ref != 'refs/heads/main'
+        if: github.ref != 'refs/heads/main' || github.event.inputs.kubo-version
```
galargh, line 229: Should we be skipping this step as well if `kubo-version` is provided? It failed during my test run: https://github.com/ipfs/ipfs-webui/actions/runs/3135588411

```suggestion
        if: github.ref != 'refs/heads/main' && !github.event.inputs.kubo-version
```

SgtPooki: The error was 

```
SemanticReleaseError: The release branches are invalid in the `branches` configuration.
```
I don't think we should skip the dry-run of semantic release as this helps determines whether the PR is going to break the real semantic release. 

We could update https://github.com/ipfs/ipfs-webui/blob/d3a652400902e564ebae5f257acfd254ee52472a/.github/workflows/ci.yml#L228 if we need to. It seems like maybe `github.ref` was empty?

SgtPooki: we could probably just change the `if: ` to check if it's a workflow_dispatch or not

SgtPooki: considering this change is specifically for running workflow_dispatch with a requested kubo version, I'm fine with the `&& !` change.

SgtPooki (approved):

hacdias (approved):

lidel:
@galargh this is not enough for E2E tests to run against new Kubo version. 
`ipfs/download-ipfs-distribution-action` is used only for publishing  ipfs-webui to IPFS and pinning to cluster, not testing it.

You either need to update `go-ipfs` dependency in `package.json`, or set `IPFS_GO_EXEC` env variable to the location of alternative `ipfs` binary

galargh:
> @galargh this is not enough for E2E tests to run against new Kubo version. `ipfs/download-ipfs-distribution-action` is used only for publishing ipfs-webui to IPFS and pinning to cluster, not testing it.
> 
> You either need to update `go-ipfs` dependency in `package.json`, or set `IPFS_GO_EXEC` env variable to the location of alternative `ipfs` binary

Good callout. We might not need this change at all then. I'll come up with a better way to test it during the next Kubo release.

PR #2035, opened 2022-09-28
chore: replace topojson with topojson-client

SgtPooki:
`topojson` has been deprecated in favor of other packages
image

`topojson-client` is recommended for the two methods we were using: https://www.npmjs.com/package/topojson#manipulation-topojson-client

works 
image


hacdias (approved):
Tried it, it works. Thanks!

PR #2052, opened 2022-10-13
fix: add missing devDependencies to package.json

Funarp:
Dep eslint-plugin-n is required for yarn start or build, missing this will lead errors. and also eslint-plugin-node should be removed if all test passed, cause eslint-plugin-n is updated eslint-plugin-node

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,6 +157,7 @@
     "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-jsx-a11y": "^6.6.0",
     "eslint-plugin-node": "^11.1.0",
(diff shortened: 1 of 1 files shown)


SgtPooki:
i'm not getting the same errors when using `npm build` after `npm i`. though I do agree that we should replace the plugin with the currently maintained one.

related: 

- https://github.com/mysticatea/eslint-plugin-node/issues/341
- https://github.com/mysticatea/eslint-plugin-node/issues/300

SgtPooki:
this is being handled by #2092

Issue #2029, opened 2022-09-19
set up tx-sync action to auto-merge


SgtPooki:
tx-sync action has been working successfully for a while (https://github.com/ipfs/ipfs-webui/actions/runs/3075313689). We should update the action so that PRs are auto-merged. https://github.com/ipfs/ipfs-webui/pull/2028

PR #2053, opened 2022-10-17
feat: Auto-Pulling translations

whizzzkid:
Issue: #2029 

In this PR:
- Removed the PR creation step as we can instead directly commit on the `main`.
- This approach was chosen because otherwise, we need to create a PR and then merge it which is not as straightforward as this.
- Also, the workflow broke recently: https://github.com/ipfs/ipfs-webui/actions/runs/3257699721
  - Looks like there is an issue reported: https://github.com/transifex/cli/pull/107
  - However this conflicts with their documentation.
  - Adding that flag explicitly.

### DIFF
--- a/.github/workflows/tx-pull.yml
+++ b/.github/workflows/tx-pull.yml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
     - name: Install Transifex client
       run: |
         curl -o- https://raw.githubusercontent.com/transifex/cli/master/install.sh | bash
@@ -25,7 +27,7 @@ jobs:
         # Optional. Local and remote branch name where commit is going to be pushed
         #  to. Defaults to the current branch.
         #  You might need to set `create_branch: true` if the branch does not exist.
-        branch: i18n-sync
+        branch: main
 
         # Optional. Options used by `git-commit`.
         # See https://git-scm.com/docs/git-commit#_options
@@ -46,33 +48,10 @@ jobs:
         # See https://git-scm.com/docs/git-add#_options
         add_options: '-A'
 
-        # Optional. Options used by `git-push`.
-        # See https://git-scm.com/docs/git-push#_options
-        push_options: '--force'
-
         # Optional. Disable dirty check and always try to create a commit and push
-        skip_dirty_check: true
+        skip_dirty_check: false
 
-        # Optional. Skip internal call to `git fetch`
-        skip_fetch: true
-
-        # Optional. Skip internal call to `git checkout`
-        skip_checkout: true
 
         # Optional. Prevents the shell from expanding filenames.
         # Details: https://www.gnu.org/software/bash/manual/html_node/Filename-Expansion.html
         disable_globbing: true
-
-        # Optional. Create given branch name in local and remote repository.
-        create_branch: true
-    - name: pull-request
-      uses: repo-sync/pull-request@65785d95a5a466e46a9d0708933a3bd51bbf9dde
-      with:
-        source_branch: "i18n-sync"
-        destination_branch: "main"
-        pr_title: "chore: pull new translations"
-        pr_body: "Automated PR created by .github/workflows/tx-pull.yml"
-        pr_label: "area/i18n/translations"
-        pr_draft: false
-        pr_allow_empty: false
-        github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/tx-pull.yml
+++ b/.github/workflows/tx-pull.yml
@@ -17,7 +17,7 @@ jobs:
         chmod +x ./tx
     - name: Pull translations from Transifex
       run: |
-        ./tx -t ${{ secrets.TX_TOKEN }} pull -a -f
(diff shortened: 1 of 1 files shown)


.github/workflows/tx-pull.yml
```diff
@@ -9,13 +9,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
```
whizzzkid, line 13: https://github.com/stefanzweifel/git-auto-commit-action#commits-made-by-this-action-do-not-trigger-new-workflow-runs

☝🏽 implies passing the token here would allow the action to inherit the write authority to `main` (unless it's protected in another way I did not account for)

.github/workflows/tx-pull.yml
```diff
@@ -9,13 +9,15 @@ jobs:
...
     - name: Install Transifex client
       run: |
         curl -o- https://raw.githubusercontent.com/transifex/cli/master/install.sh | bash
         chmod +x ./tx
     - name: Pull translations from Transifex
       run: |
-        ./tx -t ${{ secrets.TX_TOKEN }} pull -a -f
+        ./tx -t ${{ secrets.TX_TOKEN }} pull -a -t -f
```
whizzzkid, line 20: To fix the broken run: https://github.com/ipfs/ipfs-webui/actions/runs/3257699721

.github/workflows/tx-pull.yml
```diff
@@ -46,33 +48,10 @@ jobs:
         # See https://git-scm.com/docs/git-add#_options
         add_options: '-A'
 
-        # Optional. Options used by `git-push`.
-        # See https://git-scm.com/docs/git-push#_options
-        push_options: '--force'
```
whizzzkid, line 51: no longer force pushing for sanity.

.github/workflows/tx-pull.yml
```diff
@@ -46,33 +48,10 @@ jobs:
...
-
         # Optional. Disable dirty check and always try to create a commit and push
-        skip_dirty_check: true
-
-        # Optional. Skip internal call to `git fetch`
-        skip_fetch: true
+        skip_dirty_check: false
 
-        # Optional. Skip internal call to `git checkout`
-        skip_checkout: true
```
whizzzkid, line 60: keeping default checks

whizzzkid (commented):
explainers

.github/workflows/tx-pull.yml
SgtPooki: I think github already uses the token and we only need to pass token if we need to use a specific PAT

.github/workflows/tx-pull.yml
SgtPooki: when tx-pull updates the tracked files, are there dirty files we don't want to commit?

.github/workflows/tx-pull.yml
SgtPooki: I'm more okay with the removal of this one than the others.

SgtPooki (commented):
just a few concerns about the changes made to the workflow. Are we confident these changes function as we expect? Is there an example of this new workflow running successfully on the fork at whizzzkid:feature/issue-2029-auto-merge-tx-pull ?

.github/workflows/tx-pull.yml
whizzzkid: shouldn't we be checking for `dirty` always `fetch` before creating commit?

.github/workflows/tx-pull.yml
whizzzkid: I didn't find any, I think those were enabled to have dirty files on the branch that gets the translations.

.github/workflows/tx-pull.yml
whizzzkid: Token will be needed to create commits on the protected branch: https://github.com/ipfs/ipfs-webui/pull/2053/files#diff-62bcbd77df88ba2eaae2e825b94fe6e1cdbbfb09690a17b094c9a628a2d64b31L78

lidel (commented):
https://github.com/transifex/cli/pull/107 seems to be resolved upstream, do we still want to do this?

.github/workflows/tx-pull.yml
SgtPooki: i'm not sure I understand your response. IIRC, When I set up this workflow, there is a chance of dirty files that were not related to translations, and we shouldn't care about anything except the non `en` language files that are updated by transifex, and we should accept those as transifex is the source of truth for non `en` language files.

SgtPooki:
> [transifex/cli#107](https://github.com/transifex/cli/pull/107) seems to be resolved upstream, do we still want to do this?

My main concern with these changes is that I'm not as intimately familiar with them as I was when I created this workflow, and that multiple changes are being made in this single PR. 

1. tx flag changes (no longer needed)
2. pr removal
3. git commit changes (dirty checks, etc..)


I'd be much more comfortable if we kept the changes small and isolated to a single purpose. And to that end, I'd recommend the following order of updates:

1. Any changes to the commit being made after tx updates (leave PR creation so we can confirm things are still working as we expect)
2. Remove PR creation process after confirming commits are still what we want

SgtPooki:
@whizzzkid closing this since it hasn't had activity for a while

PR #2058, opened 2022-10-26
Add ipfs-search.com pinning service

dokterbob:
Allows queue'ing items for indexing on ipfs-search.com.

Reference: https://ipfs-search.readthedocs.io/en/latest/queue-pinservice.html


SgtPooki:
@dokterbob 

Hey Frido! I'm finally getting to this PR you mentioned in-person during IPFS Camp!!

As we discussed, we need a better story or UI explainer for our customers for the "fake pinning service" that ipfs-search offers.  I don't think it makes sense to enable ipfs-search-auto-index type functionality into the "default pinning services" UX because they're very different things with different goals. We should hash out where and how we think this fits into the ipfs-desktop, webui, and kubo ecosystem. 

Maybe it would make more sense if this were done as a [kubo plugin](https://github.com/ipfs/kubo/blob/master/docs/plugins.md) that we could configure via the ipfs configuration file. 

What do you think of that approach? We should get together with @juliaxbow, and @tinytb to discuss where we think this fits best.

SgtPooki (changes requested):
see previous comment.

PR #2073, opened 2022-12-09
fix: support /quic-v1

lidel:
Kubo 0.18  (https://github.com/ipfs/kubo/issues/9417) will automatically listen on both `/quic` and new `/quic-v1`
Without this fix Status page would fail like https://github.com/ipfs/ipfs-webui/issues/2033

Ref. https://github.com/ipfs/kubo/issues/9410


lidel:
I'm speed running this one: will merge, make a new release of webui, and switch Kubo to it, so 0.18.0-rc1 includes this fix.

SgtPooki (commented):
Makes sense

PR #2075, opened 2022-12-14
fix: import status window UX

rafaelramalho19:
Closes https://github.com/ipfs/ipfs-webui/issues/2030

## Demo

(1st part of the demo is with mouse clicks, 2nd is with the keyboard navigation for accessibility)

https://user-images.githubusercontent.com/24696635/207647031-0ff21567-2c7e-4bfe-9cbd-5b6e7c90fe01.mov

## Potential side-effects

- I changed `src/reset.css` to use `:focus-visible` instead of `:focus`, which makes the outline box around "tabbable" elements only appear with **KEYBOARD** and not **MOUSE** interactions.

### DIFF
--- a/src/files/file-import-status/FileImportStatus.css
+++ b/src/files/file-import-status/FileImportStatus.css
@@ -5,25 +5,19 @@
 
 .fileImportStatusButton {
     max-height: 3rem;
+    padding-right: 0;
 }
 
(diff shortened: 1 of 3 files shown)


SgtPooki:
The demo looks great! Thanks so much for throwing this together. I'm going to check out the code now.

SgtPooki (approved):
LGTM

PR #2089, opened 2023-01-30
Fix spaces between words in UK translation files

shelegdmitriy:
There is no spaces between some words what makes it more difficult to read.


lidel:
@shelegdmitriy thank you for the fix, but these files are not the source.
Translations are automatically fetched from https://www.transifex.com/ipfs/ipfs-webui/translate/#uk 

I've applied your suggestions there.
If you want to apply similar  fix in the future, please submit it at https://www.transifex.com/ipfs/ipfs-webui/translate/#uk

PR #2092, opened 2023-02-07
feat!: migrate to ESM

SgtPooki:
- feat!: migrate to kubo-rpc-client
- tmp: revert me.. just showing work
- chore: set type=module
- chore: use react-app-rewired-esm
- chore: use esm import update rule
- chore: run `npm run eslint -- --fix`
- chore: lint
- chore: getting tests working
- chore: fix pqueue constructor error
- fix: peer-locations.test.js
- tmp: trying to fix identity.test.js
- fix: AsyncRequestLoader.test.js
- chore: fix e2e/setup/ipfs-backend.js
- fix: npm start loads the webui
- fix: build,start,test succeeding
- fix: some e2e tests passing
- fix: e2e/files.test.js
- fix: e2e/ipns.test.js
- fix: e2e/peers.test.js
- fix: e2e/remote-api.test.js
- fix: e2e/settings.test.js
- fix: npm run build-storybook
- fix: storybook build & test
- fix(storybook): icons.stories.js
- fix: storybook and e2e tests both passing


SgtPooki:
@whizzzkid confirmed working in ipfs-desktop with:

```bash
# ~/code/work/protocol.ai/ipfs/webui
npm run build
cp -r build/ ../ipfs-desktop/assets/webui/

# ~/code/work/protocol.ai/ipfs/ipfs-desktop

npm run start

```

patches/eslint-plugin-import+2.26.0.patch
```diff
@@ -0,0 +1,265 @@
+diff --git a/node_modules/eslint-plugin-import/config/recommended-esm.js b/node_modules/eslint-plugin-import/config/recommended-esm.js
```
SgtPooki, line 1: will be able to remove when https://github.com/import-js/eslint-plugin-import/pull/2701 is merged

src/bundles/identity.test.js
```diff
@@ -8,14 +8,33 @@
  * Run the tests with
  *    KUBO_PORT_2033_TEST=5001 npm run test:unit -- --runTestsByPath "test/kubo-webtransport.test.js" --env=./custom-jest-env.js
  */
-import ipfsHttpModule from 'ipfs-http-client'
-import { createController } from 'ipfsd-ctl'
+describe.skip('identity.js', function () {
```
SgtPooki, line 11: currently skipping this due to some issues, but this was only added to address concerns with https://github.com/ipfs/ipfs-webui/issues/2033 and isn't really a valid test.

src/bundles/peer-locations.js
```diff
@@ -168,7 +168,12 @@ const getPublicIP = memoize((identity) => {
       if ((ip.isV4Format(addr.address) || ip.isV6Format(addr.address)) && !ip.isPrivate(addr.address)) {
         return addr.address
       }
-    } catch (_) {}
+    } catch (e) {
+      // TODO: We should provide a way to log these errors when debugging
+      // if (['development', 'test'].includes(process.env.REACT_APP_ENV)) {
+      //   console.error(e)
+      // }
```
SgtPooki, line 175: will create an issue to address this

src/bundles/peer-locations.test.js
```diff
@@ -1,33 +1,102 @@
...
-    get: jest.fn((arg) => (arg === '5.5.5.5' || arg === '123.123.123.123') ? false : 'location-cached'),
-    set: jest.fn()
-  })
-}))
-
-jest.mock('ipfs-geoip', () => ({
-  lookup: (_, address) => {
-    if (address === '5.5.5.5') throw new Error()
-    return `${address}+looked-up`
+import createPeersLocationBundle from './peer-locations.js'
```
SgtPooki, line 1: had to change this test to handle the removal of module mocks

src/icons/Icons.stories.js
```diff
@@ -1,42 +1,76 @@
...
+  name: m,
+  Icon: iconImports[m]
 }))
 
-const filterByTextQuery = (icon) => {
-  const searchQuery = text('Search', '')
-  return icon.name.includes(searchQuery)
+const filterByTextQuery = (icon, searchQuery) => {
+  return icon.name.toLowerCase().includes(searchQuery.toLowerCase())
+}
```
SgtPooki, line 11: had to change how we were loading icons because require.context isn't available. we export all icons from index.js, so importing them all from there.

SgtPooki (commented):
self review

config-overrides.js
```diff
@@ -67,14 +68,13 @@ function modifyBabelLoaderRuleForTest (rules) {
...
+function webpackOverride (config) {
   const fallback = config.resolve.fallback || {}
 
   Object.assign(fallback, {
-    assert: require.resolve('./src/webpack-fallbacks/assert'),
-    stream: require.resolve('./src/webpack-fallbacks/stream'),
-    os: require.resolve('./src/webpack-fallbacks/os'),
-    path: require.resolve('./src/webpack-fallbacks/path')
+    stream: 'stream-browserify',
+    os: 'os-browserify/browser',
```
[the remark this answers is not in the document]
SgtPooki: in all honesty, yes it does, doesn't it? We don't really have any instance of a backend web-server for webui. Even in ipfs-desktop, it's "serving" the content from `assets/webui` as a basic webpage using https://www.electronjs.org/docs/latest/api/browser-window and https://www.electronjs.org/docs/latest/api/web-contents

So in all honesty, this is a very lazy way to not split up the build deps from the runtime deps, but we're not quite there yet.

config-overrides.js
```diff
@@ -97,7 +111,22 @@ function webpackOverride(config) {
...
+    /**
+     * @type {import('jest').Config}
+     */
+    return ({
+      ...config,
+      setupFiles: [...config.setupFiles, 'fake-indexeddb/auto'],
+      moduleNameMapper: {
+        ...config.moduleNameMapper,
+        'multiformats/basics': '<rootDir>/node_modules/multiformats/cjs/src/basics.js',
+        'multiformats/bases/(.*)$': '<rootDir>/node_modules/multiformats/cjs/src/bases/$1.js'
```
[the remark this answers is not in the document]
SgtPooki: There really shouldn't be any performance difference, and if anything, it should be more performant with the greedy match. we already know the size is > 0 because of the forward-slash in front of `X` in imports of `multiformats/bases/X`, and we want to replace everything, so the regex engine doesn't have to double/triple check endings/beginnings/counts at all.. just.. get everything.

I put together a quick hyperfine test by doing 
```javascript
  jest: (config) => {
    let mfBases = 'multiformats/bases/(.*)$'
    if (process.env.MFBASE === '1') {
      mfBases = 'multiformats/bases/(.+)$'
    } else if (process.env.MFBASE === '2') {
      mfBases = 'multiformats/bases/([^$]+)$'
    }
    /**
     * @type {import('jest').Config}
     */
    return ({
      ...config,
      setupFiles: [...config.setupFiles, 'fake-indexeddb/auto'],
      moduleNameMapper: {
        ...config.moduleNameMapper,
        'multiformats/basics': '<rootDir>/node_modules/multiformats/cjs/src/basics.js',
        [mfBases]: '<rootDir>/node_modules/multiformats/cjs/src/bases/$1.js'
      }
    })
  }
```

and running 
```bash
╰─ ✔ ❯ hyperfine 'MFBASE=0 npm run test:unit' 'MFBASE=1 npm run test:unit' 'MFBASE=2 npm run test:unit'
Benchmark 1: MFBASE=0 npm run test:unit
  Time (mean ± σ):      5.429 s ±  0.104 s    [User: 3.829 s, System: 1.008 s]
  Range (min … max):    5.232 s …  5.587 s    10 runs

Benchmark 2: MFBASE=1 npm run test:unit
  Time (mean ± σ):     17.498 s ± 24.296 s    [User: 4.158 s, System: 1.064 s]
  Range (min … max):    5.483 s … 63.602 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 3: MFBASE=2 npm run test:unit
  Time (mean ± σ):      5.675 s ±  0.114 s    [User: 3.911 s, System: 1.042 s]
  Range (min … max):    5.548 s …  5.919 s    10 runs

Summary
  'MFBASE=0 npm run test:unit' ran
    1.05 ± 0.03 times faster than 'MFBASE=2 npm run test:unit'
    3.22 ± 4.48 times faster than 'MFBASE=1 npm run test:unit'
```

npm/jest actually stalls out on MFBASE=1, below is the output when running `MFBASE=1 npm run test:unit` directly.

```bash
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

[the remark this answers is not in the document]
SgtPooki: MFBASE=0 actually has two additional checks for the process.env.MFBASE whereas the other two do not, so it's saying something that it's still faster.

Either way, optimizing regex here is.. tedious =P

config-overrides.js
```diff
@@ -87,6 +87,20 @@ function webpackOverride(config) {
...
   config.module.rules = modifyBabelLoaderRuleForBuild(config.module.rules)
+  config.module.rules.push({
+    test: /\.(js|jsx)$/,
+    exclude: /(node_modules|bower_components)/,
+    loader: 'babel-loader',
+    options: { presets: ['@babel/env', '@babel/preset-react'] }
+  })
+
+  config.module.rules.push({
+    test: /\.(m?js)$/,
```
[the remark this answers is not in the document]
SgtPooki: wanted to test locally if this caused any issues with webpack build because the build config was super wonky (we need to react-scripts eject). this is fine

PR #2095, opened 2023-02-09
test(storybook): add PeerBandwidthTable.stories.js

SgtPooki:
Similar to https://github.com/ipfs/ipfs-webui/pull/2094

adding storybook coverage to file with low coverage at https://app.codecov.io/gh/ipfs/ipfs-webui/tree/main/src/status

@alanshaw @hacdias @lidel @olizilla Do any of you know if this component was ever used? If so, was it removed purposefully? 

It looks like this table is supposed to be surfaced inside the currently used peers table as a single column (see screenshot from https://github.com/ipfs/ipfs-webui/pull/1616)

other related issues:

- https://github.com/ipfs/ipfs-webui/issues/1037
- https://github.com/ipfs/ipfs-gui/issues/38
- https://github.com/ipfs/ipfs-webui/issues/1003


SgtPooki:
> Why call it a chore?

I guess it's technically a test but wasnt required, isnt a fix, isnt a feat, and not docs.. 

Im fine with either.


> - https://icon.now.sh/triangleUp or down is not reachable, can we also fix that as part of this?
> 
> 
> 
> Not blocking this but that broken icon looks bad.
> 

Yea as mentioned in the description, this component doesn't seem to be used, so i think we could address later if the component is used? But i think the component has value either way as an example.. we could add a version of this to the status or peers page later and could address it then.

PR #2179, opened 2023-11-28
fix: error loading scaleway template URL

SgtPooki:
- fixes an error with setting complianceURL for scaleway when using "template" URL
- optimizes pinning services sorting (less mapping)

When running `npm start` I got the following error:

image


SgtPooki:
this was pushed directly to main cause it was blocking other things: https://github.com/ipfs/ipfs-webui/commit/7a3388e3d17560d3e376e57361dd78b8c7fdf8e9

Issue #2190, opened 2023-12-15
bug: webui not loading in firefox with cookies and localStorage disabled


ivan386:
**Describe the bug**
I try to open webui and browser show me nothing.

My browser Firefox in private mod. All cookies disabled in settings. It is block access to local storage.

In console i see errors:

First error:
```
17:08:30.946 Uncaught DOMException: The operation is insecure.
    value i18nextLocalStorageBackend.js:50
    e i18nextLocalStorageBackend.js:60
    value i18nextChainedBackend.js:19
    value i18nextChainedBackend.js:49
    n i18next.js:1925
    value i18next.js:2361
    61455 i18n.js:19
    Webpack 4
        __webpack_require__
        <anonymous>
        <anonymous>
        <anonymous>
main.20056d59.js:2
    value i18nextLocalStorageBackend.js:50
    e i18nextLocalStorageBackend.js:60
    value i18nextChainedBackend.js:19
    forEach self-hosted:203
    value i18nextChainedBackend.js:49
    n i18next.js:1925
    value i18next.js:2361
    61455 i18n.js:19
    Webpack 4
        __webpack_require__
        <анонимный>
        <анонимный>
        <анонимный>
``` 

Error happen here in code (i18nextLocalStorageBackend.js:50):
```javascript
function getDefaults() {
  return {
    prefix: 'i18next_res_',
    expirationTime: 7 * 24 * 60 * 60 * 1000,
    defaultVersion: undefined,
    versions: {},
    store: typeof window !== 'undefined' ? window.localStorage : null /* error on this line */
  };
}
```

Second error:
```
17:08:30.988 Uncaught (in promise) DOMException: The operation is insecure.
    n money-clip.m.js:32
2 main.20056d59.js:2
```

Error happen here in code (money-clip.m.js:32):
```javascript
function createStore(dbName, storeName) {
    const dbp = idbReady().then(() => {
        const request = indexedDB.open(dbName); /* error on this line */
        request.onupgradeneeded = () => request.result.createObjectStore(storeName);
        return promisifyRequest(request);
    });
    return (txMode, callback) => dbp.then((db) => callback(db.transaction(storeName, txMode).objectStore(storeName)));
}
```

**To Reproduce**
1. Disable all cookies in Firefox 
2. Open webui in private mod

**Expected behavior**
Web UI is shown

**Desktop (please complete the following information):**
 - OS: Windows 7
 - Browser Firefox
[...]

PR #2192, opened 2024-01-04
fix(i18n): 🩹 Fixing store for localStorageBackend plugin

whizzzkid:
Closes: #2190


SgtPooki (changes requested):
I don't see anything inherently wrong with the code here, but I also don't see that this fixes the problems in the referenced issue.

I posted some comments on that issue, but I think there are higher priority items. Folks can allow localStorage/cookie access explicitly for the URL they're using to access the webui as a workaround.

PR #2216, opened 2024-04-05
chore: disable countly analytics and hide unused UI

### LINKED ISSUE #2198: Remove/update countly.ipfs.tech telemetry
This is a placeholder issue: we won't have paid Countly after Feb 18th, 2024
and iiuc should remove / disable code that sends requests to countly.ipts.tech.

There is a proposal in https://github.com/ipfs-shipyard/ignite-metrics/issues/133 to update ignite-metrics to use Plausible instead and then have this repo update to use the new ignite-metrics.

Regardless of the chosen path, the Countly shutdown will happen first, independent of this work.  This means that for a time, metrics will be POSTed to countly but not stored.

cc https://github.com/ipfs/ipfs-companion/issues/1315 https://github.com/ipfs/ipfs-desktop/issues/2743

lidel:
This PR applies minimal changes to Close #2198 but without waiting for https://github.com/ipfs-shipyard/ignite-metrics/issues/133

- hide analytics UI components
- stop sending broken opt-out metrics to instance which no longer works

It does not remove any code, and allows restoring analytics once we have time to do https://github.com/ipfs-shipyard/ignite-metrics/issues/133.

### DIFF
--- a/src/bundles/analytics.js
+++ b/src/bundles/analytics.js
@@ -65,11 +65,17 @@ import { onlyOnceAfter } from '../lib/hofs/functions.js'
  * @property {number} lastDisabledAt
  * @property {string[]} consent
  * @property {boolean?} showAnalyticsBanner
+ * @property {boolean?} showAnalyticsComponents
  * @property {boolean?} optedOut
  *
  * @typedef {Object} State
  * @property {Model} analytics
  */
+
+// 2024-Q2:
+// All analytics are disabled since we no longer use Countly instance.
+// See https://github.com/ipfs/ipfs-webui/issues/2198
+const DISABLE_ALL_ANALYTICS = true
 
 // Unknown actions (can't seem to see anything
 // dispatching those).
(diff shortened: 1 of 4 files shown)


lidel:
@SgtPooki does this look ok to ship as a patch release?

src/bundles/analytics.js
```diff
@@ -442,8 +453,9 @@ const createAnalyticsBundle = ({
       state = state || {
         lastEnabledAt: 0,
         lastDisabledAt: 0,
+        showAnalyticsComponents: DISABLE_ALL_ANALYTICS, // hide related UI  for now, see https://github.com/ipfs/ipfs-webui/issues/2198
         showAnalyticsBanner: false,
-        optedOut: false,
+        optedOut: !DISABLE_ALL_ANALYTICS, // disable analytics by default for now, see https://github.com/ipfs/ipfs-webui/issues/2198
```
SgtPooki, line 458:
```suggestion
        optedOut: DISABLE_ALL_ANALYTICS, // disable analytics by default for now, see https://github.com/ipfs/ipfs-webui/issues/2198
```

SgtPooki: don't we want to set this opted out to true for now? if not, we should probably leave as `false`

SgtPooki (approved):
approved, but we're currently setting `optedOut` to false

src/bundles/analytics.js
lidel: Yeah my bad, I negated wrong line. Fixed in 6f36572b12ca89db88260d85d3f9c85346f067db + updated tests to confirm we hide UI and do no tracking by default.

PR #2237, opened 2024-05-29
fix: loading empty content

oldshensheep:
content may be an empty string but`!"" === true` so it will use` ComponentLoader`, but we should load it as it has content(empty content)


lidel (approved):
Thank you!

PR #2253, opened 2024-08-30
fix(file-preview): safeSubresourceGwUrl

lidel:
This aims to sanitize localhost `src` url for embedded image/audio/video to avoid mixed-content warning in latest chrome-based browsers.

Rationale: https://github.com/ipfs/ipfs-webui/issues/2246#issuecomment-2322192398

### DIFF
--- a/src/files/file-preview/FilePreview.js
+++ b/src/files/file-preview/FilePreview.js
@@ -60,14 +60,14 @@ const Preview = (props) => {
         <Drag {...props}>
           {/* eslint-disable-next-line jsx-a11y/media-has-caption */}
           <audio width='100%' controls>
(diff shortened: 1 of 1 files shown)


lidel (commented):
The challenge of testing this, is that it works on localhost subdomain, but it may be a bug in chrome that manifests only when top level document URL is `https://` and not just a secure context (in Chromium codebase, these are not the same, and we reported bugs before).

I'm going to merge this and deploy to `dev.webui.ipfs.io`:
- If problem is solved, we will cut a patch release next week.
- If not, I'll fill a bug in Chromium issue tracker and as Igalia for help.

PR #2254, opened 2024-08-31
fix(files): avoid duplicated fetch during preview

### LINKED ISSUE #2217: Files: video/image bytes are loaded twice on preview page
Tested on `webui.ipfs.io` and `http://127.0.0.1:5001/webui/`, Firefox 124.0.2 and Google Chrome 124.0.6367.18.

Opening image in Files triggers fetch twice
- once via Kubo RPC and `/api/v0/cat` 
- once via subdomain gateway at `cid.ipfs.localhost:port`

> 2024-04-09-004847_1475x68_scrot

I guess we had code somewhere which should try one, and then fallback to other, but instead does both?

There is also unrelated(?) bug in Chrome where the gateway GET listed here fails (https://github.com/ipfs/ipfs-webui/issues/2218)

lidel:
- Closes #2217 by 
 - [x] not using `ipfs.cat` if file is embedded as subresource  (image, video, pdf, audio)
 - [x] limiting `ipfs.cat` to 10KiB for other files types 
 - [x] removing buggy pagination (low utility, it never worked, it always fetched entire file)
 
 ## Preview
 
 Can be tested with big txt file, like `bafybeicbevt3gr2jhs23f2sikwcqtirrhrdyscriutfg5mtlxoud55htii` 
 
 >

### DIFF
--- a/src/bundles/files/actions.js
+++ b/src/bundles/files/actions.js
@@ -214,6 +214,12 @@ const actions = () => ({
           ...fileFromStats({ ...stats, path }),
           fetched: time,
           type: 'file',
+          /**
+          * Reads a portion of data from IPFS.
+          * @param {number} offset - The starting point to read from.
+          * @param {number} length - The number of bytes to read.
+          * @returns {AsyncIterable<Uint8Array>} An async generator that yields the data read from IPFS.
+          */
           read: (offset, length) => ipfs.cat(stats.cid, { offset, length }),
           name: path.split('/').pop(),
           size: stats.size,
--- a/src/bundles/files/actions.js
+++ b/src/bundles/files/actions.js
@@ -214,7 +214,7 @@ const actions = () => ({
           ...fileFromStats({ ...stats, path }),
           fetched: time,
           type: 'file',
-          read: () => ipfs.cat(stats.cid),
+          read: (offset, length) => ipfs.cat(stats.cid, { offset, length }),
           name: path.split('/').pop(),
           size: stats.size,
           cid: stats.cid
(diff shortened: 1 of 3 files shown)


src/bundles/files/actions.js
```diff
@@ -214,7 +214,7 @@ const actions = () => ({
           ...fileFromStats({ ...stats, path }),
           fetched: time,
           type: 'file',
-          read: () => ipfs.cat(stats.cid),
+          read: (offset, length) => ipfs.cat(stats.cid, { offset, length }),
```
SgtPooki, line 217: nit: this CI action is failing but this is not a TS file.. we could add jsdoc.. 

do we want to disable typecheck CI while we are not full typescript?

src/files/file-preview/FilePreview.js
```diff
@@ -29,7 +31,11 @@ const Preview = (props) => {
   const type = typeFromExt(name)
 
   const loadContent = useCallback(async () => {
-    const readBuffer = buffer || await read()
+    if (['audio', 'video', 'pdf', 'image'].includes(type)) {
+      // noop, we dont need to read() preview for these because we embedd them on page
```
SgtPooki, line 35:
```suggestion
      // noop, we dont need to read() preview for these because we embed them on page
```

SgtPooki (approved):

src/bundles/files/actions.js
lidel: Let's keep it, we get contributions from people, and better if they include types (easier for us to review).
I've added jsdoc.

PR #2259, opened 2024-09-12
chore: use lts node version in CI

SgtPooki:
Updates all github workflows selections of node-version to be "lts/*"


SgtPooki (commented):
self review

lidel (commented):
building with 20 lgtm, but it fails to build for me with `v22.4.1`, so the CI will break when 24 becomes LTS and someone will have to  search for last working version and hardcode it again :crying_cat_face: 

is `lts/*` useful in CI? i remember the same breackage occuring 12→14→16→18 (nearly every time a  change of LTS image broke CI)

maybe hardcode to `20.x` to avoid headache? (not feeling strongly, i'd just rather not deal with problem once LTS breaks again)

SgtPooki:
> is `lts/*` useful in CI of rarely-maintained apps? i remember the same breackage occuring 12→14→16→18 (nearly every time a change of LTS image broke CI in one of our legacy JS apps)

most likely not, but we probably want to see a failure when it occurs right?

> maybe hardcode to `20.x` to avoid headache? (not feeling strongly, i'd just rather not deal with problem once LTS breaks again)

That is good with me, but we should probably target 22 now since it's going active very soon

image

SgtPooki:
I think once I get https://github.com/ipfs/ipld-explorer-components/pull/448 done, we will be closer to updating everything here.

SgtPooki:
> but it fails to build for me with `v22.4.1`

It does build for me with `v22.9.0`

SgtPooki:
done in https://github.com/ipfs/ipfs-webui/pull/2283

PR #2268, opened 2024-09-23
fix: set Explore defaults on app load

SgtPooki:
fix: set Explore defaults on app load

* sets localStorage settings required by ipld-explorer-components on app load
    * disable remote gateways
    * enable local gateways

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -21,6 +21,7 @@ import FilesExploreForm from './files/explore-form/FilesExploreForm.js'
 
 export class App extends Component {
   static propTypes = {
+    doSetupLocalStorage: PropTypes.func.isRequired,
     doTryInitIpfs: PropTypes.func.isRequired,
     doInitHelia: PropTypes.func.isRequired,
     doUpdateUrl: PropTypes.func.isRequired,
@@ -30,6 +31,11 @@ export class App extends Component {
     filesPathInfo: PropTypes.object,
     // Injected by DropTarget
     isOver: PropTypes.bool.isRequired
+  }
+
+  constructor (props) {
+    super(props)
+    props.doSetupLocalStorage()
   }
 
   componentDidMount () {
@@ -131,6 +137,7 @@ export default connect(
   'doExploreUserProvidedPath',
   'doUpdateUrl',
   'doUpdateHash',
+  'doSetupLocalStorage',
   'doTryInitIpfs',
   'doInitHelia',
   'doFilesWrite',
--- a/src/bundles/ipfs-provider.js
+++ b/src/bundles/ipfs-provider.js
@@ -315,6 +315,25 @@ const selectors = {
  */
 
 const actions = {
+
+  doSetupLocalStorage: () => async () => {
+    /** For the Explore page (i.e. ipld-explorer-components) */
+    const useRemoteGatewaysToExplore = localStorage.getItem('explore.ipld.gatewayEnabled')
+    if (useRemoteGatewaysToExplore === null) {
+      // by default, disable remote gateways for the Explore page (i.e. ipld-explorer-components)
+      await writeSetting('explore.ipld.gatewayEnabled', false)
+    }
+
+    const kuboGateway = readSetting('kuboGateway')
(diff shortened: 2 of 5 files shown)


SgtPooki:
note that we need to update ipld-explorer-components to use @helia/block-brokers@3.0.4 so we do not need to set `allowInsecure: true`


```console
╰─ ✘ 1 ❯ npm why @helia/block-brokers
@helia/block-brokers@3.0.3
node_modules/@helia/block-brokers
  @helia/block-brokers@"^3.0.3" from @helia/http@1.0.10
  node_modules/@helia/http
    @helia/http@"^1.0.10" from ipld-explorer-components@7.0.2
    node_modules/ipld-explorer-components
      ipld-explorer-components@"^7.0.2" from the root project
  @helia/block-brokers@"^3.0.3" from ipld-explorer-components@7.0.2
  node_modules/ipld-explorer-components
    ipld-explorer-components@"^7.0.2" from the root project
```

src/bundles/ipfs-provider.js
```diff
@@ -315,6 +315,25 @@ const selectors = {
...
+    const useRemoteGatewaysToExplore = localStorage.getItem('explore.ipld.gatewayEnabled')
+    if (useRemoteGatewaysToExplore === null) {
+      // by default, disable remote gateways for the Explore page (i.e. ipld-explorer-components)
+      await writeSetting('explore.ipld.gatewayEnabled', false)
+    }
+
+    const kuboGateway = readSetting('kuboGateway')
+    if (kuboGateway === null || typeof kuboGateway === 'string' || typeof kuboGateway === 'boolean' || typeof kuboGateway === 'number') {
+      // empty or invalid, set defaults
+      await writeSetting('kuboGateway', { trustlessBlockBrokerConfig: { init: { allowLocal: true, allowInsecure: true } } })
```
lidel, line 330: nit: shouldn't we store port and host? user may have it on something other than `127.0.0.1:8080`, for example ex-brave user could have `127.0.0.1:48080`

src/App.js
```diff
@@ -32,6 +33,11 @@ export class App extends Component {
     isOver: PropTypes.bool.isRequired
   }
 
+  constructor (props) {
+    super(props)
+    props.doSetupLocalStorage()
```
lidel, line 38: This does not seem to work on initial page load.
If you use empty browser profile and load webui, and then go to ipld explorer and click on wikipedia example you will see spinner forever, and no request:

>  
>  

If you then force-reload the page (F5) the helia will initialize with local gateway and work fine:

>

lidel (changes requested):

src/bundles/ipfs-provider.js
SgtPooki: That happens elsewhere so i was planning to leave that alone

src/App.js
SgtPooki: I think landing directly on explore page might cause the issue? 

Issue doesn't happen if you land on status page first

SgtPooki: Might need to call the localStorage fn in router / component loader

src/bundles/ipfs-provider.js
SgtPooki: I guess that would work for now but it doesn't feel great

SgtPooki:
@lidel try now, it should work now that we're explicitly using ipld-explorer-components@7.0.3

src/App.js
SgtPooki: landing directly on explore page works for me with the latest.

https://github.com/user-attachments/assets/c7953852-56af-48e2-840b-3b41c76f83d3

SgtPooki (commented):
self review

lidel (approved):
Thanks, it now works :partying_face: feel free to ship a patch release.

(I also confirmed running on a custom RPC port also works)

src/App.js
lidel: Yep, that was it. New version works fine.

ps. TIL [Firefox does not support HEVC while it is encumbered by patents](https://www.reddit.com/r/firefox/comments/zeg2qy/firefox_doesnt_support_h265_hevc_for_patent/), so inline player does not work (but i played fine with external player): 

> image

PR #2273, opened 2024-10-10
fix: remove iltorb dep, build for node v22.9.0

SgtPooki:
bundlesize->ilortb->node-gyp was throwing errors when trying to build using nodev22

this pr fixes build for node v22 by removing dependency on bundlesize, which was only a dev-dep.

spawned due to https://github.com/ipfs/ipfs-webui/pull/2259#pullrequestreview-2333846193

This PR also fixes a bug with the npm script `analyze` not functioning properly

We can now see that the bundle size is ~23MB:

image


SgtPooki (commented):
self review

PR #2276, opened 2024-10-13
fix: PDF viewer refreshes when interacting with web UI

### LINKED ISSUE #2145: PDF Viewer Refreshes when interacting with webui
**Describe the bug**
While interacting with other elements on page, the PDF viewer re-renders.

**To Reproduce**
1. Import a sample PDF in either webui/desktop
2. Try interacting with ui elements around the pdf
3. PDF viewer re-renders.
4. ReRenders can be expensive and based on the size of file it can take longer.

**Expected behavior**
It should not re-render.

**Screenshots**

https://github.com/ipfs/ipfs-webui/assets/1895906/e573131f-c233-48d7-9f52-0558101a2f95

 - OS: MacOS Ventura 13.5 (22G74)
 - Browser Chrome, IPFS - Desktop
 - Version Version 115.0.5790.114 (Official Build) (arm64), webui: https://github.com/ipfs/ipfs-webui/commit/a43dd23

**Additional context**
It looks like it should be fairly straightforward to remove pdf-view updates from the rendering cycle.

MattWong-ca:
In Files, the PDF viewer was re-rendering when interacting with web UI.

In [FilesPage.js](https://github.com/ipfs/ipfs-webui/blob/main/src/files/FilesPage.js), `MainView` is defined as a new component inside the `FilesPage` component, so clicking UI elements around the PDF re-renders `MainView` (and since `MainView` contains the logic that shows the file preview component, the preview re-renders). Issue seemed to only affect PDFs.

Fix: creating and using a new file preview component defined outside of `FilePages`. Not sure if this is the cleanest approach, so I'm open to suggestions!

Closes #2145 

https://github.com/user-attachments/assets/853c578d-d843-474a-9548-5314ac607bb9

### DIFF
--- a/src/files/FilesPage.js
+++ b/src/files/FilesPage.js
@@ -116,7 +116,7 @@ const FilesPage = ({
   }
 
   const MainView = ({ t, files, remotePins, pendingPins, failedPins, doExploreUserProvidedPath }) => {
-    if (!files) return (<div/>)
+    if (!files || files.type === 'file') return (<div/>)
 
     if (files.type === 'unknown') {
       const path = files.path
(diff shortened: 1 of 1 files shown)


lidel (approved):
Thank you @MattWong-ca, this looks like a sensible fix.

I confirmed this not only fixes PDF preview, but also Video (it no longer get interrupted).

SgtPooki (commented):
awesome.. thanks a ton for this @MattWong-ca

PR #2287, opened 2024-10-31
fix: normalize catch-all listeners to loopback

### LINKED ISSUE #1821: Download Broken - Tries To Use IPFS gateway at 0.0.0.0:8080
#### Version information:

go-ipfs version: 0.8.0-ce693d7
Repo version: 11
System version: amd64/linux
Golang version: go1.14.4

The above is from a container running docker tag ipfs/go-ipfs:v0.8.0, running on Ubuntu 20.04 LTS.

#### Description:

Browse to the admin web site (yourserver:5001), click on Files, click on the elipsis, and click Download. The download will fail.

If you use F12 in FireFox to open the networking tab and try again, you'll see a GET failure to 0.0.0.0:8080. Presumably, the web site is assuming that the IP address the gateway HTTP server is bound on is also an IP address it can use from the browser.

Since the IPFS gateway is running in a Docker container, it binds to 0.0.0.0, and does not know what the external IP address of its own IPFS gateway is. This would also apply if the IPFS gateway is behind a reverse proxy or NAT.

#### Suggested changes

* The web site should detect obvious incorrect information for the gateway (e.g. when it is 0.0.0.0) and alert the user. The web site already does this for detection of incorrect CORS configuration
* There should be an additional setting to specify the public address of its own gateway. If this setting already exists, the alert should describe how to configure it or refer the user to documentation on how to do it

A harsher option could be to require the new field to be populated when the gateway is bound to 0.0.0.0, but that wouldn't be backward compatible.

lidel:
Closes #1821


SgtPooki:
@lidel FYI the fleek build is caused by different deps not compatible with node 22, that should be fixed by https://github.com/ipfs/ipfs-webui/pull/2283

src/bundles/config.js
```diff
@@ -91,9 +91,11 @@ function getURLFromAddress (name, config) {
   if (!config) return null
 
   try {
-    const address = Array.isArray(config.Addresses[name])
+    const address = (Array.isArray(config.Addresses[name])
       ? config.Addresses[name][0]
-      : config.Addresses[name]
+      : config.Addresses[name])
+      .replace(/\/0\.0\.0\.0\//, '/127.0.0.1/') // fix for https://github.com/ipfs/ipfs-webui/issues/1821
+      .replace(/\/::\//, '/::1/')
```
SgtPooki, line 98: should we take this chance to more gracefully handle the array of addresses if it exists? do we always want the first one, or do we want the local one if one is found?

Vt2049:
k

SgtPooki (approved):

PR #2292, opened 2024-11-12
chore(ci): udpate artifact actions to v4

lidel:
See https://github.com/ipfs/distributions/pull/1112


lidel:
Merging to avoid brownout issues, the unit test failure is not related, it was happening in main branch already.

PR #2295, opened 2024-11-14
docs: size-related labels in Files screen

lidel:
Closes https://github.com/ipfs/ipfs-webui/issues/1248 without changing UI.

We just clarify that it is normal if the first number is bigger than the second one.

When at the root of MFS (all files):
> 

We now have more friendly label when in a subdirectory:

> 

While at it, also added this disclaimer to "Import IPFS Path" :

>


SgtPooki (approved):
lgtm, awesome

PR #2302, opened 2024-11-28
Do not specify replication factor when pinning - use cluster's default

hsanjuan:
This was not a good idea when pinning to collab clusters...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,8 +137,6 @@ jobs:
               --basic-auth "${CLUSTER_USER}:${CLUSTER_PASSWORD}" \
               pin add \
               --name "${PIN_NAME}" \
-              --replication-min 1 \
-              --replication-max 3 \
               --wait \
               $PIN_CID
         env:


lidel (approved):
I vaguely remember this was either workaround to not wait for all cluster nodes  to confirm pinning, or break when not all nodes are online.


Either way, let's merge and figure out if still an issue once we make the next release (cc @SgtPooki  for visibility)

hsanjuan:
ahh, yes could be! That cluster has a 3rd party peer that is broken to all effects. Direct CAR upload with --local is definitely the best.

dist.ipfs.io pins without --wait and then waits itself for things to be pinned in at least one peer:

```
echo "::group::pin add"
ipfs-cluster-ctl --enc=json \
    --host "/dnsaddr/ipfs-websites.collab.ipfscluster.io" \
    --basic-auth "${CLUSTER_USER}:${CLUSTER_PASSWORD}" \
    pin add \
    --name "${PIN_NAME}" \
    --no-status $PIN_ADD_EXTRA_ARGS \
    "$PIN_CID"
echo "::endgroup::"

# TODO: below can now be replaced with ipfs-cluster-ctl --wait now (https://github.com/ipfs/ipfs-cluster/blob/v0.14.1/CHANGELOG.md#v0141---2021-08-18)
echo "::group::waiting until pinned"
    while true; do
    ipfs-cluster-ctl --enc=json \
        --host "/dnsaddr/ipfs-websites.collab.ipfscluster.io" \
        --basic-auth "${CLUSTER_USER}:${CLUSTER_PASSWORD}" \
        status "$PIN_CID" | tee cluster-pin-status
    if [[ $(jq '.peer_map[].status' cluster-pin-status | grep '"pinned"' | wc -l) -ge 1 ]]; then
        echo "Got first pin confirmation, finishing the workflow"
        break
    else
        echo "Still waiting for at least one pin confirmation, sleeping for 1 minute.."
        sleep 60
    fi
    done
echo "::endgroup::"
```

PR #2304, opened 2024-11-29
ci workflow: Add CAR file directly to cluster

hsanjuan:
Since we already have the CAR, add directly rather than waiting for bitswap upload.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,25 +125,15 @@ jobs:
         continue-on-error: true
         run: |
           ipfs-cluster-ctl --enc=json \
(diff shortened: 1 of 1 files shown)


lidel (approved):
Thanks! I'll test today since we need fix release with from https://github.com/ipfs/ipfs-webui/pull/2305 for next Kubo.

lidel:
Looks good. Old [pinning took 30m](https://github.com/ipfs/ipfs-webui/actions/runs/11893294770/job/33138080999), new [CAR-based took <8m](https://github.com/ipfs/ipfs-webui/actions/runs/12087618957/job/33709446700).

PR #2318, opened 2025-01-20
feat(settings): public gateway inputs allow ports

SgtPooki:
- **fix: subdomain gateway share url supports ports**
- **test: add port tests for public gateway input**

FYI: still in progress because the tests are not working for some reason even though it works with local testing.


SgtPooki:
i'm hoping something is up with my local build/tests and that the tests actually pass in CI because things *are* working when I manually test.

PR #2321, opened 2025-01-28
fix(ci): cluster pinning with timeout and retry

lidel:
This PR is backport of https://github.com/protocol/badbits.dwebops.pub/pull/48053

- Upload of <50MiB CAR with webui should take only a few seconds, everything more means something is really, really wrong
- It runs pinning to cluster in a loop, assuming pin should not take longer than 1 minute. 
- If the first attempt fails, follow-up ones will run with debug enabled to see which cluster node is picked and fails
- We retry up to 5 minutes
- We disable collab-cluster-am6-1.cluster.dwebops.net for now

cc @SgtPooki  @2color  for visibility

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,18 +123,40 @@ jobs:
         if: github.ref == 'refs/heads/main'
         continue-on-error: true
         run: |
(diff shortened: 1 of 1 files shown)


SgtPooki (approved):
lgtm

PR #2322, opened 2025-01-30
feat: add theme provider

### LINKED ISSUE #1702: Feature: dark mode
My eyes, I go blind at night when I see the screen, this window is very bright.

They could place a selection of themes to change the colors to suit one.

use the colors used by gtk / kde apps and use those colors / themes

**Additional context**
try to change the dependency of the use of npm for yarn, I use yarn instead of npm but well it is not so important but I must mention it since npm is not the only one.

kaf-lamed-beyt:
fixes #1702 

Hi @lidel, This is what I have so far. I'm currently using an approach involving a Context Provider to access the theme value across the entire application. 

I'm also adding the styles individually in index.css by targeting the `data-theme` attribute. This is nice, but it can be quite tasking, hence my move towards a custom context hook. This would allow us to dynamically change colors across components in the application.

2025-01-3015-52-55-ezgif com-video-to-gif-converter

I ran into a slight blocker though. `App.js` uses a legacy React pattern &mdash; a class-based component which makes it a bit difficult to use the hook in this component, so I resorted to using the value from `localStorage` which is outside of the component/app life-cycle. So even if the value of `theme` in `localStorage` changes, the color &mdash; based on the ternary operation below &mdash; of the header won't change, because we do not trigger a re-render. Which is what the hook helps us achieve. 

```jsx
<div
     className="flex items-center ph3 ph4-l"
      style={{
         WebkitAppRegion: "drag",
          height: 75,
          background: currentTheme === "dark" ? "red" : "#F0F6FA",
          paddingTop: "20px",
          paddingBottom: "15px",
        }}
>
```

So, my question now is if it is possible to refactor `App.js` into a function-based component

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -18,6 +18,7 @@ import Notify from './components/notify/Notify.js'
 import Connected from './components/connected/Connected.js'
 import TourHelper from './components/tour/TourHelper.js'
 import FilesExploreForm from './files/explore-form/files-explore-form.tsx'
+import { ThemeProvider, ThemeContext } from './context/theme-provider'
 
 export class App extends Component {
   static propTypes = {
@@ -31,6 +32,8 @@ export class App extends Component {
     // Injected by DropTarget
     isOver: PropTypes.bool.isRequired
   }
+
+  static contextType = ThemeContext
 
   constructor (props) {
     super(props)
@@ -62,45 +65,48 @@ export class App extends Component {
 
   render () {
     const { t, route: Page, ipfsReady, doFilesNavigateTo, routeInfo: { url }, connectDropTarget, canDrop, isOver, showTooltip } = this.props
+    const currentTheme = typeof window !== 'undefined' && localStorage.getItem('theme')
     return connectDropTarget(
-      // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
-      <div className='sans-serif h-100 relative' onClick={getNavHelper(this.props.doUpdateUrl)}>
-        {/* Tinted overlay that appears when dragging and dropping an item */}
-        { canDrop && isOver && <div className='h-100 top-0 right-0 fixed appOverlay' style={{ background: 'rgba(99, 202, 210, 0.2)' }} /> }
-        <div className='flex flex-row-reverse-l flex-column-reverse justify-end justify-start-l' style={{ minHeight: '100vh' }}>
-          <div className='flex-auto-l'>
-            <div className='flex items-center ph3 ph4-l' style={{ WebkitAppRegion: 'drag', height: 75, background: '#F0F6FA', paddingTop: '20px', paddingBottom: '15px' }}>
-              <div className='joyride-app-explore' style={{ width: 560 }}>
-                <FilesExploreForm onBrowse={doFilesNavigateTo} />
+      <div>
+        <ThemeProvider>
+          {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
+          <div className='sans-serif h-100 relative' onClick={getNavHelper(this.props.doUpdateUrl)}>
+            {/* Tinted overlay that appears when dragging and dropping an item */}
+            { canDrop && isOver && <div className='h-100 top-0 right-0 fixed appOverlay' style={{ background: 'rgba(99, 202, 210, 0.2)' }} /> }
+            <div className='flex flex-row-reverse-l flex-column-reverse justify-end justify-start-l' style={{ minHeight: '100vh' }}>
+              <div className='flex-auto-l'>
+                <div className='flex items-center ph3 ph4-l' style={{ WebkitAppRegion: 'drag', height: 75, background: currentTheme === 'dark' ? 'red' : '#F0F6FA', paddingTop: '20px', paddingBottom: '15px' }}>
+                  <div className='joyride-app-explore' style={{ width: 560 }}>
+                    <FilesExploreForm onBrowse={doFilesNavigateTo} />
+                  </div>
+                  <div className='dn flex-ns flex-auto items-center justify-end'>
+                    <TourHelper />
+                    <Connected className='joyride-app-status' />
+                  </div>
+                </div>
+                <main className='bg-white pv3 pa3 pa4-l'>
+                  { (ipfsReady || url === '/welcome' || url.startsWith('/settings'))
+                    ? <Page />
+                    : <ComponentLoader />
+                  }
+                </main>
               </div>
-              <div className='dn flex-ns flex-auto items-center justify-end'>
-                <TourHelper />
-                <Connected className='joyride-app-status' />
+              <div className='navbar-container flex-none-l bg-navy'>
+                <NavBar />
               </div>
             </div>
-            <main className='bg-white pv3 pa3 pa4-l'>
-              { (ipfsReady || url === '/welcome' || url.startsWith('/settings'))
-                ? <Page />
-                : <ComponentLoader />
-              }
-            </main>
+            <ReactJoyride
+              run={showTooltip}
+              steps={appTour.getSteps({ t })}
+              styles={appTour.styles}
+              callback={this.handleJoyrideCb}
+              scrollToFirstStep
+              disableOverlay
+              locale={getJoyrideLocales(t)}
+            />
+            <Notify />
           </div>
-          <div className='navbar-container flex-none-l bg-navy'>
-            <NavBar />
-          </div>
-        </div>
-
-        <ReactJoyride
-          run={showTooltip}
-          steps={appTour.getSteps({ t })}
-          styles={appTour.styles}
-          callback={this.handleJoyrideCb}
-          scrollToFirstStep
-          disableOverlay
-          locale={getJoyrideLocales(t)}
-        />
-
-        <Notify />
+        </ThemeProvider>
       </div>
     )
   }
--- a/src/context/theme-provider.tsx
+++ b/src/context/theme-provider.tsx
@@ -1,0 +1,49 @@
+import React from 'react'
+
+export interface ThemeProviderProps {
+  children: React.ReactNode
+}
+
+export type Theme = 'light' | 'dark'
+
+export type ThemeContextValues = {
+  currentTheme: Theme,
+  toggleTheme: () => void;
+  toggleThemeWithKey: (event: React.KeyboardEvent<HTMLButtonElement>) => void;
+}
+
+const createThemeContext = () => React.createContext<ThemeContextValues | null>(null)
+export const ThemeContext = createThemeContext()
+
+export const ThemeProvider = ({ children }: ThemeProviderProps) => {
+  const [theme, setTheme] = React.useState<boolean>(() => {
+    const savedTheme =
+      typeof window !== 'undefined' && localStorage.getItem('theme')
+    if (savedTheme) return savedTheme === 'dark'
+    return window.matchMedia('prefers-color-scheme: dark').matches
+  })
+  React.useEffect(() => {
+    const htmlElem = document.documentElement
+    const currentTheme = theme ? 'dark' : 'light'
+    htmlElem.setAttribute('data-theme', currentTheme)
+    localStorage.setItem('theme', currentTheme)
+    htmlElem.setAttribute('aria-label', `Current theme: ${currentTheme}`)
+  }, [theme])
+  const toggleTheme = () => setTheme((prevTheme) => !prevTheme)
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLButtonElement>) => {
+    if (event.key === 'Enter' || event.key === ' ') {
+      event.preventDefault()
+      toggleTheme()
+    }
+  }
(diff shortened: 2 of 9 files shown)


kaf-lamed-beyt:
Hi @lidel, kindly take a look at this when you get the chance. Thanks.

SgtPooki:
> So, my question now is if it is possible to refactor App.js into a function-based component

@kaf-lamed-beyt I would love for all components in ipfs-webui to be functional components. That is the direction we're moving. However, since that will likely be a significant change, we should probably do that as a separate PR.

You can see an example of a simple function component that is also using redux-bundler-react at https://github.com/ipfs/ipfs-webui/blob/7fbb51bf23192dda3527e2944a6b7b14e810a869/src/explore/explore-page-renderer.jsx#L32

kaf-lamed-beyt:
Oh! Thanks for the feedback, @SgtPooki! I'll take a look at the file you pointed me to now. 

So, from the looks of things now, are you suggesting that i use the manual approach of targeting each element based on the styles in `[data-theme="dark|light"]` in index.css? 

We can always keep the hook.

src/navigation/NavBar.js
```diff
@@ -73,6 +74,9 @@ export const NavBar = ({ t }) => {
         </div>
       </div>
       <div className='dn db-l navbar-footer mb2 tc center f7 o-80 glow'>
+        <div className='mb4'>
+          <ThemeToggle />
+        </div>
```
lidel, line 79: :thought_balloon:  spale in left mavbar is limited especially in ipfs-desktop where UI acts as native app.  

perhaps it is  better to move this toggle to top right corner, where we already have two items:

> image

kaf-lamed-beyt: oh, yeah! that sounds goood, @lidel! 

i originally thought of placing it there, but couldn't locate the icons in the code. I'll try looking for them and update the theme toggle's position

src/context/theme-provider.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const ThemeProvider = ({ children }: ThemeProviderProps) => {
+  const [theme, setTheme] = React.useState<boolean>(() => {
+    const savedTheme =
+      typeof window !== 'undefined' && localStorage.getItem('theme')
+    if (savedTheme) return savedTheme === 'dark'
+    return window.matchMedia('prefers-color-scheme: dark').matches
+  })
+  React.useEffect(() => {
+    const htmlElem = document.documentElement
+    const currentTheme = theme ? 'dark' : 'light'
```
SgtPooki, line 27: seems like we can use better names for `theme` and `setTheme` since it seems to be a boolean indicating whether `dark` theme is selected or not

src/App.js
```diff
@@ -62,45 +65,48 @@ export class App extends Component {
...
-                <FilesExploreForm onBrowse={doFilesNavigateTo} />
+      <div>
+        <ThemeProvider>
+          {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
+          <div className='sans-serif h-100 relative' onClick={getNavHelper(this.props.doUpdateUrl)}>
+            {/* Tinted overlay that appears when dragging and dropping an item */}
+            { canDrop && isOver && <div className='h-100 top-0 right-0 fixed appOverlay' style={{ background: 'rgba(99, 202, 210, 0.2)' }} /> }
+            <div className='flex flex-row-reverse-l flex-column-reverse justify-end justify-start-l' style={{ minHeight: '100vh' }}>
+              <div className='flex-auto-l'>
+                <div className='flex items-center ph3 ph4-l' style={{ WebkitAppRegion: 'drag', height: 75, background: currentTheme === 'dark' ? 'red' : '#F0F6FA', paddingTop: '20px', paddingBottom: '15px' }}>
```
SgtPooki, line 78: Do we really want the background to be red here? I haven't ran the code to see for myself yet but this seems wrong?

src/context/theme-provider.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }, [theme])
+  const toggleTheme = () => setTheme((prevTheme) => !prevTheme)
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLButtonElement>) => {
+    if (event.key === 'Enter' || event.key === ' ') {
+      event.preventDefault()
+      toggleTheme()
+    }
+  }
+  const values: ThemeContextValues = {
+    currentTheme: theme as unknown as Theme,
```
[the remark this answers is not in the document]
SgtPooki: ideally, use of `as` (typescript casting) should be limited as much as possible.

src/context/theme-provider.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+
+export interface ThemeProviderProps {
+  children: React.ReactNode
+}
```
SgtPooki, line 5: instead of explicitly setting children for props, you can `extend React.FC<PropsWithChildren>` or similar.. 

or, my preference since there are no custom props, remove this and do it below

```suggestion
```

src/context/theme-provider.tsx
```diff
@@ -0,0 +1,49 @@
...
+export type Theme = 'light' | 'dark'
+
+export type ThemeContextValues = {
+  currentTheme: Theme,
+  toggleTheme: () => void;
+  toggleThemeWithKey: (event: React.KeyboardEvent<HTMLButtonElement>) => void;
+}
+
+const createThemeContext = () => React.createContext<ThemeContextValues | null>(null)
+export const ThemeContext = createThemeContext()
```
SgtPooki, line 16:
```suggestion
export const ThemeContext = React.createContext<ThemeContextValues | null>(null)
```

src/context/theme-provider.tsx
```diff
@@ -0,0 +1,49 @@
...
+export interface ThemeProviderProps {
+  children: React.ReactNode
+}
+
+export type Theme = 'light' | 'dark'
+
+export type ThemeContextValues = {
+  currentTheme: Theme,
+  toggleTheme: () => void;
+  toggleThemeWithKey: (event: React.KeyboardEvent<HTMLButtonElement>) => void;
```
SgtPooki, line 12: I'm not convinced we need both of these. Can `toggleTheme` accept an optional event and handle expected keys?

src/index.css
```diff
@@ -36,5 +80,5 @@ html, body, #root {
 
 .react-joyride__tooltip button {
   font-size: 14px !important;
-  font-family: 'Montserrat';
+  font-family: "Montserrat";
```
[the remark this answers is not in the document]
kaf-lamed-beyt: ah! my bad. i must have explicitly triggered 'format on save"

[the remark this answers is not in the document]
SgtPooki: it happens! no worries, thanks a ton for throwing this together, i know folks have been wanting this one for a while :D (over 4 years!)

src/hooks/theme.ts
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import { ThemeContext, ThemeContextValues } from '../context/theme-provider'
+
+export const useTheme = () => {
+  const context = React.useContext(ThemeContext)
+  if(context === null) {
```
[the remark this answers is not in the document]
kaf-lamed-beyt: oh. so the strict equal caused that?

src/context/theme-provider.tsx
SgtPooki: perfect

src/App.js
kaf-lamed-beyt: oh! do not mind that. i was using it to test

src/context/theme-provider.tsx
kaf-lamed-beyt: I think there's this minor issue with using `React.FC` in newer versions of React and I understand that it sits well with this component since it expects children which `React.FC<PropsWithChildren>` provides implicitly. 

why shouldn't we set the `children` prop explicitly?

src/context/theme-provider.tsx
kaf-lamed-beyt: i'll try refactoring this.

src/context/theme-provider.tsx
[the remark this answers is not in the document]
kaf-lamed-beyt: thank you for catching this. I'll update it.

src/context/theme-provider.tsx
kaf-lamed-beyt: @SgtPooki

src/index.css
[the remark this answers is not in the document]
kaf-lamed-beyt: over four years?! dang!

kaf-lamed-beyt:
@lidel, I've moved the toggle from the navbar.  Here's what it looks like now. 

ipfs-webui-dark-theme-ezgif com-video-to-gif-converter

I still have to wrap up with the other routes. When I do, I'll push the changes. 

cc: @SgtPooki

src/context/theme-provider.tsx
SgtPooki: @kaf-lamed-beyt no huge reason, but here are a few: 

* prevent issues if React changes the type of the children prop
* less code to maintain
* using prop types from react directly prevents needing to add other props if we need to extend them (adding className or other common props in the future)

SgtPooki: > I think there's this minor issue with using React.FC in newer versions of React

I haven't ran into this, can you link to examples?

kaf-lamed-beyt: huh. makes sense. 

for this 👇🏼 
> using prop types from react directly prevents needing to add other props if we need to extend them (adding className or other common props in the future)

I may be wrong... but, can't we do `...props`?

> I haven't ran into this, can you link to examples?

This [issue](https://github.com/facebook/create-react-app/pull/8177) from facebook/create-react-app sheds more light on why they deprecated `React.FC` @SgtPooki

kaf-lamed-beyt:
done with this. @SgtPooki, @lidel,  kindly review. 
2025-02-1216-36-22-ezgif com-video-to-gif-converter

the explore route was quite tricky to work on. I couldn't locate the elements/components that are used to render data in the UI, that's why the one on the right still retained a white background. 

perhaps, if anyone can point me to where they are exactly.

kaf-lamed-beyt:
@SgtPooki, @lidel, using this to bump your notification(s) respectively

SgtPooki (changes requested):
FYI I set brave browser to prefer dark theme, reloaded the page, and it didn't take. 

Also, I don't see an event listener for `matchMedia` so i'll probably add one..

there are a few minor nits i'm looking through and will probably update and then merge this.

kaf-lamed-beyt:
> FYI I set brave browser to prefer dark theme, reloaded the page, and it didn't take.
> 
> Also, I don't see an event listener for `matchMedia` so i'll probably add one..
> 
> there are a few minor nits i'm looking through and will probably update and then merge this.

Oh okay, @SgtPooki, I'll wait for the changes.

kaf-lamed-beyt:
> We should add file extensions to all imports, and also prefer `useEffect` as a direct import instead of using `React.blah`

oh! I'll take note of that. i also understand the preference for `useEffect` instead of `React.useEffect`.  Old habits, haha.

I also can't see the video you attached BTW

image

SgtPooki:
> I also can't see the video you attached BTW

try this GIF:

webui dark theme demo

SgtPooki (commented):
it seems like there is a bit more work to do cleaning up old create-react-app cruft before we can fully fix the import paths.. there are issues with importing extensions.. 

i want to move the build for ipfs-webui to vite.. i think we will need to do that before addressing more build issues or import pathings.

src/components/api-address-form/ApiAddressForm.js
```diff
@@ -3,12 +3,14 @@ import { connect } from 'redux-bundler-react'
...
 import Button from '../button/button.tsx'
 import { checkValidAPIAddress } from '../../bundles/ipfs-provider.js'
+import { useTheme } from '../../hooks/theme'
 
 const ApiAddressForm = ({ t, doUpdateIpfsApiAddress, ipfsApiAddress, ipfsInitFailed }) => {
   const [value, setValue] = useState(asAPIString(ipfsApiAddress))
   const initialIsValidApiAddress = !checkValidAPIAddress(value)
   const [showFailState, setShowFailState] = useState(initialIsValidApiAddress || ipfsInitFailed)
   const [isValidApiAddress, setIsValidApiAddress] = useState(initialIsValidApiAddress)
+  const { darkTheme: isDarkTheme } = useTheme()
```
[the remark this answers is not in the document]
SgtPooki: fixed this in a later commit

SgtPooki (approved):
lgtm.. i pushed up some changes. @kaf-lamed-beyt please take a look, if you're good with it, I will merge this today

kaf-lamed-beyt:
> lgtm.. i pushed up some changes. @kaf-lamed-beyt please take a look, if you're good with it, I will merge this today
> 
> cc @lidel if you want to take a look but I won't block on you

oh alright. I'll do this now.

kaf-lamed-beyt:
This looks good to me, @SgtPooki. I saw how you listened to `prefers-color-scheme` with `matchMedia` too. Fine work! :wine:

PR #2307, opened 2024-11-30
feat: bulk CIDs import 

MattWong-ca:
Introduces bulk import of CIDs using a text file, as requested in #2173 

Users can upload a text file with a list of CIDs (and an optional name right after), as shown below:
```
bafkreibm6jg3ux5qumhcn2b3flc3tyu6dmlb4xa7u5bf44yegnrjhc4yeq
QmawceGscqN4o8Y8Fv26UUmB454kn2bnkXV5tEQYc4jBd6 barrel.png
QmbvrHYWXAU1BuxMPNRtfeF4DS2oPmo5hat7ocqAkNPr74 pi equals.png
```
- Click `Bulk import` from Import dropdown and a modal to select a text file will pop up
- Frontend validation is done on each line of the text file using `isIPFS.cid()`, if any line fails then user won't be able to import it. Otherwise the name of the selected file will be shown
- Added new function `doFilesBulkCidImport` in actions.js responsible for going through text file and adding each CID
- Added stories.js file 
- Didn't include a CLI tutor mode command since it seemed more complex than the other ones already present, but happy to revisit if needed!

Testing
- Manually tested to ensure it works in both the root folder and subfolders
- Tried using a text file with improper formatting, which didn't let me import
- **NOTE:** the most number of CIDs I tested with in a file was 91, at which there was a slight lag before all the files were imported

Demo

https://github.com/user-attachments/assets/0eaee9ca-6791-40a9-8eda-80768c5e52fb

PR #2323, opened 2025-02-03
feat(files): CAR Import

badgooooor:
### Description

- Close  https://github.com/ipfs/ipfs-webui/issues/1798
- Add "Import CAR File", including
  - Dropdown menu in import
     image
  - CAR File import modal with file input, optional name input and validation on file name by extension
  - Add local definition of DAG from kubo to `IPFSService`
Screenshot 2568-02-03 at 21 29 52
Screenshot 2568-02-03 at 21 30 01
Screenshot 2568-02-03 at 21 30 10

### DIFF
--- a/@types/ipfs/index.d.ts
+++ b/@types/ipfs/index.d.ts
@@ -2,6 +2,7 @@ declare module 'ipfs' {
   import type { CID } from 'multiformats/cid'
   import type { Multiaddr } from '@multiformats/multiaddr'
   import type { Buffer } from 'buffer'
+  import type { KuboRPCClient } from 'kubo-rpc-client'
 
   declare export interface IPFSService extends CoreService {
     pin: PinService;
@@ -9,7 +10,7 @@ declare module 'ipfs' {
     name: NameService;
     object: ObjectService;
     config: ConfigService;
-    dag: DagService;
+    dag: KuboRPCClient['dag'];
 
     stop(options?: TimeoutOptions): Promise<void>
   }
@@ -51,21 +52,6 @@ declare module 'ipfs' {
     replace(config: Object, options?: TimeoutOptions): Promise<void>;
 
     profiles: ConfigProfiles;
-  }
-
-  declare export interface DAGImportOptions {
-    pinRoots: boolean;
-  }
-
-  declare export interface DAGImportResult {
-    root: {
-      cid: CID;
-      pinErrorMsg?: string;
-    };
-  }
-
-  declare export interface DAGService {
-    import(sources: Iterable<Uint8Array> | AsyncIterable<Uint8Array> | AsyncIterable<AsyncIterable<Uint8Array>> | Iterable<AsyncIterable<Uint8Array>>, options?: DAGImportOptions): AsyncIterable<DAGImportResult>
   }
 
   declare export interface ConfigProfiles {
--- a/@types/ipfs/index.d.ts
+++ b/@types/ipfs/index.d.ts
@@ -1,5 +1,3 @@
-import { DAGImportOptions, DAGImportResult } from 'kubo-rpc-client';
-
 declare module 'ipfs' {
   import type { CID } from 'multiformats/cid'
   import type { Multiaddr } from '@multiformats/multiaddr'
@@ -53,6 +51,17 @@ declare module 'ipfs' {
     replace(config: Object, options?: TimeoutOptions): Promise<void>;
 
     profiles: ConfigProfiles;
+  }
+
+  declare export interface DAGImportOptions {
+    pinRoots: boolean;
+  }
+
+  declare export interface DAGImportResult {
+    root: {
+      cid: CID;
+      pinErrorMsg?: string;
+    };
   }
 
   declare export interface DAGService {
--- a/@types/ipfs/index.d.ts
+++ b/@types/ipfs/index.d.ts
@@ -1,3 +1,5 @@
+import { DAGImportOptions, DAGImportResult } from 'kubo-rpc-client';
+
 declare module 'ipfs' {
   import type { CID } from 'multiformats/cid'
   import type { Multiaddr } from '@multiformats/multiaddr'
@@ -9,6 +11,7 @@ declare module 'ipfs' {
     name: NameService;
     object: ObjectService;
     config: ConfigService;
+    dag: DagService;
 
     stop(options?: TimeoutOptions): Promise<void>
   }
@@ -50,6 +53,10 @@ declare module 'ipfs' {
     replace(config: Object, options?: TimeoutOptions): Promise<void>;
 
     profiles: ConfigProfiles;
+  }
+
+  declare export interface DAGService {
+    import(sources: Iterable<Uint8Array> | AsyncIterable<Uint8Array> | AsyncIterable<AsyncIterable<Uint8Array>> | Iterable<AsyncIterable<Uint8Array>>, options?: DAGImportOptions): AsyncIterable<DAGImportResult>
   }
 
   declare export interface ConfigProfiles {
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -63,6 +63,7 @@
   "addByCarModal": {
     "title": "Import from CAR",
     "description": "Insert a CAR file to import",
+    "namePlaceholder": "Name (optional)",
     "examples": "Examples:"
   },
   "newFolderModal": {
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -63,6 +63,7 @@
   "addByCarModal": {
     "title": "Import from CAR",
     "description": "Select a CAR file to import",
+    "selectCARButtonText": "Select CAR…",
     "namePlaceholder": "Name (optional)",
     "examples": "Examples:"
   },
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -65,7 +65,8 @@
     "title": "CAR Import",
     "description": "Choose a CAR file to import a DAG from and specify a name for it to be imported into the current directory.",
     "selectCARButtonText": "Select CAR…",
-    "namePlaceholder": "Name"
+    "namePlaceholder": "Name",
+    "renameImportPath": "Rename import"
   },
   "bulkImportModal": {
     "title": "Bulk import with text file",
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -14,6 +14,7 @@
   "openWithLocalAndPublicGateway": "Try opening it instead with your <1>local gateway</1> or <3>public gateway</3>.",
   "cantBePreviewed": "Sorry, this file can’t be previewed",
   "addByPath": "From IPFS",
+  "addByCar": "From CAR",
   "newFolder": "New folder",
   "generating": "Generating…",
   "actions": {
@@ -57,6 +58,11 @@
     "preloadNote": "NOTE: Imported path is lazy-loaded by default, and will be retrieved on-demand as you browse it. To proactively prefetch imported path to the local Datastore, choose ”Download” from the context menu.",
     "importPathPlaceholder": "Path or CID (required)",
     "namePlaceholder": "Name (optional)",
+    "examples": "Examples:"
+  },
+  "addByCarModal": {
+    "title": "Import from CAR",
+    "description": "Insert a CAR file to import",
     "examples": "Examples:"
   },
   "newFolderModal": {
--- a/public/locales/en/files.json
+++ b/public/locales/en/files.json
@@ -62,11 +62,10 @@
     "examples": "Examples:"
   },
   "addByCarModal": {
-    "title": "Import from CAR",
-    "description": "Select a CAR file to import",
+    "title": "CAR Import",
+    "description": "Choose a CAR file to import a DAG from and specify a name for it to be imported into the current directory.",
     "selectCARButtonText": "Select CAR…",
-    "namePlaceholder": "Name (optional)",
-    "examples": "Examples:"
+    "namePlaceholder": "Name"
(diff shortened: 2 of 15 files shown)


src/bundles/files/actions.js
```diff
@@ -398,6 +399,47 @@ const actions = () => ({
...
+      const { done, value } = await reader.read()
+      if (done) break
+      fileResultStr += value
+    }
+    const encoder = new TextEncoder()
+    const fileResult = encoder.encode(fileResultStr)
+    try {
+      const result = await all(ipfs.dag.import([fileResult], {
+        pinRoots: true
+      }))
```
SgtPooki, line 424: I think the problem here might be how we're reading the actual file bytes and passing it to dag.import.

We want something like https://github.com/ipfs/helia/blob/f7155d8bece43dd91d19060881294f61df9d222c/packages/interop/src/car.spec.ts#L60-L66

Basically we need to get the bytes of the carFile we get (which should be of type `File`) to bytes, and then pass those bytes (without encoding/decoding) to ipfs.dag.import.

An example of a car import process directly into the helia blockstore can be seen at https://github.com/ipfs/ipld-explorer-components/blob/57d625dadb934556506030fc532bf5ad24f36dc6/src/lib/import-car.ts#L12

src/bundles/files/actions.js
```diff
@@ -620,7 +652,7 @@ const dirStats = async (ipfs, cid, { path, isRoot, sorting }) => {
     const absPath = join(path, f.name)
     let file = null
 
-    if (dirCount < 1000 && (f.type === 'directory' || f.type === 'dir')) {
```
lidel, line 623: why we change this?

https://github.com/ipfs/ipfs-webui/blob/d234a6012a899e52fdac1c44730c0e7f7af6412e/src/bundles/files/actions.js#L47 still refers to `directory` 


avoid touching things unrelated to CAR import :pray:

src/bundles/files/actions.js
```diff
@@ -398,6 +398,36 @@ const actions = () => ({
...
+        // to avoid leaking private information via Countly?
+        throw Object.assign(new Error('ipfs.files.cp call failed'), {
+          code: 'ERR_FILES_CP_FAILED'
+        })
+      }
+      return file
+    } finally {
+      await store.doFilesFetch()
+    }
+  }),
```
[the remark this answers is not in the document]
badgooooor: Just found out that `ipfs`is actually `KuboRPCClient` but JSDocs annotation is `IPFSService`. This also happens in couple of methods in actions. Not sure if there's way to fix in order to pass Typecheck in CI?

@types/ipfs/index.d.ts
```diff
@@ -1,3 +1,5 @@
+import { DAGImportOptions, DAGImportResult } from 'kubo-rpc-client';
```
lidel, line 1: iiuc we dont use types from kubo-rcp-client for all other commands,  but have local definitions in @types/ipfs/index.d.ts.
maybe follow existing pattern and do the same?

src/bundles/files/utils.js
```diff
@@ -315,3 +316,25 @@ export const ensureMFS = (store) => {
     throw new Error('Unable to perform task if not in MFS')
   }
 }
+
+/**
+ * Get bytes from file
+ * @param {FileStream} file
+ * @returns {Promise<Uint8Array>}
+ */
+export const fileToByteArray = (file) => {
```
2color, line 325: You can simplify this by using the [Blob.arrayBuffer()](https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer) function which already returns a promise with an ArrayBuffer which you can convert to a Uint8Array.

Though we probably want to stream the upload, assuming that the kubo-rpc-client supports it.

src/bundles/files/actions.js
```diff
@@ -398,6 +398,37 @@ const actions = () => ({
...
+   * Adds CAR file. On completion will trigger `doFilesFetch` to update the state.
+   * @param {FileStream} carFile
+   * @param {string} name
+   */
+  doAddCarFile: (carFile, name = '') => perform(ACTIONS.ADD_CAR_FILE, async (ipfs, { store }) => {
+    ensureMFS(store)
+
+    const fileResult = await fileToByteArray(carFile)
+    try {
+      const result = await all(ipfs.dag.import([fileResult], {
```
2color, line 411: You should be able to pass a stream of the file to dag.import in order to avoid loading the whole CAR into memory.

src/bundles/files/actions.js
```diff
@@ -398,6 +398,37 @@ const actions = () => ({
...
+    ensureMFS(store)
+
+    const stream = carFile.content.stream()
+    try {
+      const result = await all(ipfs.dag.import(stream, {
+        pinRoots: true
+      }))
+      const cid = result[0].root.cid
+      const src = `/ipfs/${cid}`
+      const dst = join(realMfsPath('/files'), name)
```
lidel, line 416: Found a bug: relative path within MFS (Files) is not respected. 

If I am in a subdirectory `/a/b` and I try to import `c.car` then the CAR data is added at `/c` rather than `/a/b/c`

lidel (changes requested):
Thank you @badgooooor, found some issues while testing current interation. 
Details inline.

lidel (approved):
Thank you @badgooooor!

I've pushed small cosmetic changes to follow similar visual language as in "Bulk import" feature added in #2307, added CLI tutor docs and unified CAR Import/Export language. 

Looks good to merge on my end, but would appreciate second set of eyes on UI and language.

@SgtPooki  / @2color  mind taking a final look, and if looks ok, merge?

2color:
I discovered a bug when you try to import and the name already exists. I've pushed https://github.com/ipfs/ipfs-webui/pull/2323/commits/a229aa9cfb8ea2304a9bd7f9ba6aad527e33c107 to fix this, in lieu of a better way to show errors, I just throw a more helpful error. 

Besides that, LGTM.

Thanks @badgooooor 🎉

SgtPooki:
i'm getting an error when importing:

```
core.ts:90 Uncaught (in promise) HTTPError: file does not exist
    at Object.errorHandler [as handleError] (core.ts:90:22)
    at async Client.fetch (http.ts:165:11)
    at async Object.stat (stat.ts:9:17)
    at async actions.js:421:7
    at async task.js:100:12
    at async task.js:199:20
    at async utils.js:139:12
```

src/bundles/files/actions.js
```diff
@@ -399,10 +399,57 @@ const actions = () => ({
...
+      } catch (/** @type {any} */ err) {
+        // the file does not exist yet.. swallow error for checking if it exists
+      }
+
+      try {
+        await ipfs.files.cp(src, dst)
+      } catch (/** @type {any} */ err) {
+        // TODO: Not sure why we do this. Perhaps a generic error is used
+        // to avoid leaking private information via Countly?
+        throw Object.assign(new Error('ipfs.files.cp call failed'), {
```
SgtPooki, line 436: FYI we no longer use countly.. we need to strip it from this repo

src/files/FilesPage.js
```diff
@@ -15,13 +15,13 @@ import FilesList from './files-list/FilesList.js'
 import { getJoyrideLocales } from '../helpers/i8n.js'
 
 // Icons
-import Modals, { DELETE, NEW_FOLDER, SHARE, RENAME, ADD_BY_PATH, BULK_CID_IMPORT, CLI_TUTOR_MODE, PINNING, PUBLISH } from './modals/Modals.js'
+import Modals, { DELETE, NEW_FOLDER, ADD_BY_CAR, SHARE, RENAME, BULK_CID_IMPORT, ADD_BY_PATH, CLI_TUTOR_MODE, PINNING, PUBLISH } from './modals/Modals.js'
```
SgtPooki, line 18:
```suggestion
import Modals, { DELETE, NEW_FOLDER, ADD_BY_CAR, SHARE, RENAME, ADD_BY_PATH, BULK_CID_IMPORT, CLI_TUTOR_MODE, PINNING, PUBLISH } from './modals/Modals.js'
```

remove unnecessary change

2color:
@SgtPooki sorry that was oversight on my end. I used the `then` notation (avoiding try catch) because we swallow the error. I just pushed a small fix.

SgtPooki (commented):
a few small nits.. 

I think we should update the modal to provide a little more information.. here is a quick edit I made that I think improves clarity for users a bit:

image

@types/ipfs/index.d.ts
```diff
@@ -2,13 +2,15 @@ declare module 'ipfs' {
...
   import type { Buffer } from 'buffer'
+  import type { KuboRPCClient } from 'kubo-rpc-client'
 
   declare export interface IPFSService extends CoreService {
     pin: PinService;
     files: FileService;
     name: NameService;
     object: ObjectService;
     config: ConfigService;
+    dag: KuboRPCClient['dag'];
```
SgtPooki, line 13: note that I removed the custom defined types here so we don't have to manage them later. Ideally this whole IPFSService will be going away and we'll be using types from kubo-rpc-client going forward.

public/locales/en/files.json
```diff
@@ -60,6 +61,13 @@
     "namePlaceholder": "Name (optional)",
     "examples": "Examples:"
   },
+  "addByCarModal": {
+    "title": "CAR Import",
+    "description": "Choose a CAR file to import a DAG from and specify a name for it to be imported into the current directory.",
+    "selectCARButtonText": "Select CAR…",
+    "namePlaceholder": "Name",
+    "renameImportPath": "Rename import"
```
SgtPooki, line 69: new label for rename input

src/bundles/files/actions.js
```diff
@@ -399,10 +399,56 @@ const actions = () => ({
...
+   * @param {FileStream} carFile
+   * @param {string} name
+   */
+  doAddCarFile: (root, carFile, name = '') => perform(ACTIONS.ADD_CAR_FILE, async (/** @type {IPFSService} */ ipfs, { store }) => {
+    ensureMFS(store)
+
+    const stream = carFile.content.stream()
+    try {
+      // @ts-expect-error - https://github.com/ipfs/js-kubo-rpc-client/issues/278
+      const result = await all(ipfs.dag.import(stream, {
```
SgtPooki, line 413: kubo-rpc-client types need updated.

src/bundles/files/actions.js
```diff
@@ -399,10 +399,56 @@ const actions = () => ({
...
+      // Check if destination path already exists
+      await ipfs.files.stat(dst).then(() => {
+        dstExists = true
+      }).catch(() => {
+        // Swallow error. We can add the file to the dst path
+      })
+
+      if (dstExists) {
+        throw new Error(`The name "${name}" already exists in the current directory. Try importing with a different name.`)
+      }
```
SgtPooki: I don't know where my try block went... will add it back

src/files/modals/add-by-car-modal/AddByCarModal.js
```diff
@@ -0,0 +1,100 @@
...
+          placeholder='File'
+          accept='.car'
+          onChange={onFileInputChange}
+        />
+        <Button className='ma2 tc' bg='bg-teal' onClick={onAddFileButtonClick}>{t('addByCarModal.selectCARButtonText')}</Button>
+        {/* Gray text showing original car file name */}
+        { file != null
+          ? <div className='mt1 f7 center charcoal-muted tl w-90'>{file.path}</div>
+          : null
+        }
```
SgtPooki, line 89: new display of original car file

src/files/modals/add-by-car-modal/AddByCarModal.js
```diff
@@ -0,0 +1,100 @@
...
+  }, [isFileNameValid, file])
+
+  return (
+    <Modal className={className} onCancel={onCancel}>
+      <ModalBody title={t('addByCarModal.title')} Icon={Icon}>
+        <div className='mb3 flex flex-row items-center'>
+          <p className='mt0 center charcoal tl w-90'>{t('addByCarModal.description')}</p>
+        </div>
+
+        <label htmlFor='import-car-file-name' className='mt0 f7 center charcoal-muted tl w-90'>{t('addByCarModal.renameImportPath')}</label>
```
SgtPooki, line 64: new label for rename input

SgtPooki (approved):

src/bundles/files/actions.js
```diff
@@ -418,15 +418,19 @@ const actions = () => ({
...
+          // Swallow error. We can add the file to the dst path
+        })
 
-      if (dstExists) {
-        throw new Error(`The name "${name}" already exists in the current directory. Try importing with a different name.`)
+        if (dstExists) {
+          throw new Error(`The name "${name}" already exists in the current directory. Try importing with a different name.`)
+        }
+      } catch (/** @type {any} */ err) {
+        // Swallow error. We can add the file to the dst path
```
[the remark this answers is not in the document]
2color: CC @SgtPooki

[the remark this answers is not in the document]
SgtPooki: https://github.com/ipfs/ipfs-webui/pull/2323#issuecomment-2700827724

we can't just remove it, but we do need to move one thing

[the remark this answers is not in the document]
SgtPooki: nevermind, you're right.. my editor magically converted the previous try catch block to a .catch() and i didn't notice.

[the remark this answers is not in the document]
SgtPooki: fixed with [`8c5ff77` (#2323)](https://github.com/ipfs/ipfs-webui/pull/2323/commits/8c5ff77ce262d337a82a79f8535a983e45d8989c)

PR #2333, opened 2025-02-19
fix(peers): chrome regression and update ipfs-geoip to v9.2

lidel:
https://github.com/ipfs-shipyard/ipfs-geoip/releases/tag/v9.2.0 updates Peers screen to use dataset from 2025-02-18.

This fixes most of invalid country flags that suggest libp2p is breaking the speed of light ;-)

- [x] CI green 
- [x] confirm manually that flags load ok on Peers screen (in empty browser profile, to avoid cached results)
  - fixed Chrome bug on Peers screen (forcing `https://` on localhost redirect) in https://github.com/ipfs/ipfs-webui/pull/2333/commits/0310932cce5c8f038c5f1d2c29b0a69207070a16

### DIFF
--- a/src/bundles/peer-locations.js
+++ b/src/bundles/peer-locations.js
@@ -234,6 +234,10 @@ class PeerLocationResolver {
   async findLocations (gatewayUrls, peers) {
     const res = {}
 
+    // Normalize Gateway URLS:
+    // switch localhost to raw IP to avoid subdomain redirect AND avoid Chrome forcing https:// on such redirect
+    gatewayUrls = (Array.isArray(gatewayUrls) ? gatewayUrls : [gatewayUrls]).map(url => url.replace(/localhost:(\d+)/, '127.0.0.1:$1'))
+
     for (const p of this.optimizedPeerSet(peers)) {
       const peerId = p.peer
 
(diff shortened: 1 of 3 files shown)


src/bundles/peer-locations.js
```diff
@@ -234,6 +234,10 @@ class PeerLocationResolver {
   async findLocations (gatewayUrls, peers) {
     const res = {}
 
+    // Normalize Gateway URLS:
+    // switch localhost to raw IP to avoid subdomain redirect AND avoid Chrome forcing https:// on such redirect
+    gatewayUrls = (Array.isArray(gatewayUrls) ? gatewayUrls : [gatewayUrls]).map(url => url.replace(/localhost:(\d+)/, '127.0.0.1:$1'))
```
lidel, line 239: :information_source: At some point Chrome started normalizing HTTP 301 to use `https://` instead of `http://`, even if the origin is `localhost`.  This broke Flags on Peers screen, but we did not notice because most of flags were locally cached in IndexDB and was not immediately obvious it stopped working.

This is a localized fix for flags on Peers screen but may be useful to know if this bug surfaces elsewhere.

PR #2356, opened 2025-03-30
Update cors-config.sh

Yardapestr:


SgtPooki (changes requested):
is this fixing an existing issue you can link, or can you explain why we need this change?

SgtPooki (changes requested):
what is this trying to do?

Issue #2349, opened 2025-03-17
epic: typescript migration


SgtPooki:
# TypeScript Migration Guidelines

**Context:**  
We are migrating our codebase to TypeScript. This issue will remain open until the migration is fully complete. Please ensure that all new contributions follow these guidelines to maintain consistency and code quality.

## File Types & Naming Conventions

- **File Types:**  
  - Use TypeScript files only: use `.tsx` for React components and `.ts` for non-React components.
- **Filename Convention:**  
  - Use kebab-case for filenames (lowercase only, words separated by hyphens).

## Code Standards

- **Component Style:**  
  - All components must be functional.
  - Do not use `prop-types` or `.defaultProps` as these are unnecessary in a TypeScript codebase.
- **Imports:**  
  - All imports must include the `.js` extension.
- **State Management & Bundles:**  
  - Do not use `redux-bundler` or `redux-bundler-react`.  
  - Do not add new bundles; if possible, create a hook or use context instead.
- **Internationalization:**  
  - Do not use `withTransation`; use `useTranslation` instead.

## Project Configuration

- **TypeScript Setup:**  
  - Add new TypeScript files to `tsconfig.json`.
  - Include any dependent files that require type checking.  
  - If an *existing* file is too large or requires extensive changes, it can be temporarily ignored until someone can convert it properly.

## Additional Guidelines for Contributors

- **Linting & Formatting:**  
  - Follow the project’s ESLint and Prettier configurations to maintain a consistent code style.
- **Testing:**  
  - Include unit tests & e2e tests with 100% coverage for all new code. This ensures that our code remains robust during the migration.
- **Documentation:**  
  - Update or add documentation as necessary to reflect changes or new patterns introduced during the migration.
- **Dependencies:**  
  - Verify that any new third-party libraries include proper TypeScript type definitions.
- **Backwards Compatibility:**  
  - Avoid breaking changes. When updating legacy code, strive to maintain compatibility until the entire file can be migrated.
- **Questions & Support:**  
  - If you have any questions or need clarification on migration specifics, please ping @SgtPooki.

---

Your cooperation in following these guidelines is essential for a smooth transition to a fully TypeScript codebase. Thank you for your contributions!

badgooooor:
Hey, would love to help on in this issue. Is there any suggestion on tackling tasks and breakdown this epic to smaller task in order to make this issue easier to contribute?

SgtPooki:
[...]

PR #2370, opened 2025-04-14
Feat/migrate box component

tesol2y090:
# Migrate Box Component to TypeScript

## Changes
- Converted `Box` component from JavaScript to TypeScript
- Added proper type definitions and interfaces
- Improved type safety and developer experience
- Apply box component

## Related Issues
- #2349


src/components/box/Box.tsx
```diff
@@ -0,0 +1,28 @@
...
+  className = 'pa4',
+  style,
+  children,
+  ...props
+}) => {
+  return (
+    <section
+      className={className}
+      style={{ background: '#fbfbfb', ...style }}
+      {...props}
```
SgtPooki, line 21: adding `...props` here is probably the right thing to do that we weren't previously doing, but: are we receiving extra props that will change the display unexpectedly?

src/components/about-ipfs/AboutIpfs.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import { withTranslation, Trans } from 'react-i18next'
-import Box from '../box/Box.js'
+import Box from '../box/Box'
```
SgtPooki, line 3: we should include the file extension on the import

src/components/box/Box.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { ReactNode } from 'react'
+import ErrorBoundary from '../error/ErrorBoundary'
+
+interface BoxProps {
+  className?: string;
+  style?: React.CSSProperties;
+  children: ReactNode;
+  [key: string]: any; // For any additional props
```
SgtPooki, line 8: if available, 

```suggestion
interface BoxProps extends ComponentProps<'section'> {
  className?: string;
  style?: React.CSSProperties;
  children: ReactNode;
```

or use explicit types

SgtPooki (changes requested):
thanks for this!! just a few minor changes requested

PR #2385, opened 2025-07-15
fix: refresh stale multiaddrs on the Status page

### LINKED ISSUE #1964: Status page doesn't refresh to display new multiaddresses
**Describe the bug**
I had originally opened https://github.com/ipfs/go-ipfs/issues/9043 since it seemed `RelayClient` took a while to get public multiaddresses. However, as I've just realised the public multiaddresses are probably found fairly early, but, the WebUI has to be refreshed to show the public multiaddresses that are found later on.

**To Reproduce**
1. Run `ipfs daemon` (`RelayClient` enabled).
2. Look under **Advanced** and see local multiaddresses only.
3. Wait until public multiaddresses have evidently been found (other peers can find you).
4. Observe that the information under **Advanced** hasn't updated.
5. Refresh the page and see the public multiaddresses show up under **Advanced**.

**Expected behavior**
When new multiaddresses are detected, the **Status** page should refresh by itself, or just update the information under **Advanced**, to reflect the current status.

**Desktop**
 - OS: Arch Linux
 - Browser: Firefox v102.0.1
 - Version: v2.15.1 (originally found: v2.15.0)

lidel:
This PR closes #1964 and ensures addresses added by AutoTLS/RelayV2 are updated and visible in WebUI correctly:

- one-time update when Status Page is opened
- update addresses every 5s  if addrs section is expanded

This way we call `/ipfs/v0/id` only when UI actually needs it:
>

### DIFF
--- a/src/bundles/identity.js
+++ b/src/bundles/identity.js
@@ -1,12 +1,15 @@
 import { createAsyncResourceBundle, createSelector } from 'redux-bundler'
 
+// Matches APP_IDLE and peer bandwidth update intervals
+export const IDENTITY_REFRESH_INTERVAL = 5000
+
 const bundle = createAsyncResourceBundle({
   name: 'identity',
   actionBaseType: 'IDENTITY',
   getPromise: ({ getIpfs }) => getIpfs().id().catch((err) => {
     console.error('Failed to get identity', err)
   }),
-  staleAfter: Infinity,
+  staleAfter: IDENTITY_REFRESH_INTERVAL,
   persist: false,
   checkIfOnline: false
 })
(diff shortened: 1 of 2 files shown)


src/bundles/identity.js
```diff
@@ -1,12 +1,15 @@
 import { createAsyncResourceBundle, createSelector } from 'redux-bundler'
 
+// Matches APP_IDLE and peer bandwidth update intervals
+export const IDENTITY_REFRESH_INTERVAL = 5000
```
2color, line 4:
```suggestion
export const IDENTITY_REFRESH_INTERVAL_MS = 5000
```

PR #2387, opened 2025-07-21
feat: add async provide on share/copy/pin/ipns

lidel:
This PR dispatches one-time extra announce of the root CID when sharing, copying CIDs, pinning, or publishing to IPNS.

## Rationale

Providing this CID costs us virtually nothing, but will improve performance when content is accessed by others shortly after these operations by ensuring we are as close to the desired number of Amino DHT peers with provider records as possible, and also that there ARE provider records for the root CID in the DHT in the first place.

The latter especially helps desktop users, who may share a CID that has not been provided yet due to the node being offline for a while and starting IPFS Desktop recently, with Kubo having many CIDs to announce.

### DIFF
--- a/src/bundles/files/utils.js
+++ b/src/bundles/files/utils.js
@@ -1,12 +1,14 @@
 import { sortByName, sortBySize } from '../../lib/sort.js'
 import { IS_MAC, SORTING } from './consts.js'
 import * as Task from '../task.js'
+import { debouncedProvide } from '../../lib/files.js'
 
 /**
  * @typedef {import('ipfs').IPFSService} IPFSService
  * @typedef {import('../../lib/files').FileStream} FileStream
  * @typedef {import('./actions').Ext} Ext
  * @typedef {import('./actions').Extra} Extra
+ * @typedef {import('multiformats/cid').CID} CID
  */
 
 /**
@@ -314,5 +316,21 @@ export const ensureMFS = (store) => {
   const info = store.selectFilesPathInfo()
   if (!info || !info.isMfs) {
     throw new Error('Unable to perform task if not in MFS')
+  }
+}
+
+/**
+ * Dispatches an async provide operation for a CID.
+ *
+ * @param {CID|null|undefined} cid - The CID to provide
+ * @param {IPFSService} ipfs - The IPFS service instance
+ * @param {string} context - Context for logging
+ */
+export const dispatchAsyncProvide = (cid, ipfs, context) => {
+  if (cid) {
+    console.debug(`[${context}] Dispatching one-time ad-hoc provide for root CID ${cid.toString()} (non-recursive) for improved performance when sharing today`)
+    debouncedProvide(cid, ipfs).catch((error) => {
+      console.error(`[${context}] debouncedProvide failed:`, error)
+    })
   }
 }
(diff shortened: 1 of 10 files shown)


src/bundles/files/utils.js
```diff
@@ -316,3 +318,19 @@ export const ensureMFS = (store) => {
...
+
+/**
+ * Dispatches an async provide operation for a CID.
+ *
+ * @param {CID|null|undefined} cid - The CID to provide
+ * @param {IPFSService} ipfs - The IPFS service instance
+ * @param {string} context - Context for logging
+ */
+export const dispatchAsyncProvide = (cid, ipfs, context) => {
+  if (cid) {
```
SgtPooki, line 330:
```suggestion
  if (cid != null) {
```

src/bundles/files/utils.js
```diff
@@ -316,3 +318,19 @@ export const ensureMFS = (store) => {
...
+    console.debug(`[${context}] Dispatching one-time ad-hoc provide for root CID ${cid.toString()} (non-recursive) for improved performance when sharing today`)
+    debouncedProvide(cid, ipfs).catch((error) => {
```
SgtPooki, line 332:
```suggestion
    void debouncedProvide(cid, ipfs).catch((error) => {
```

src/lib/files.js
```diff
@@ -152,6 +152,46 @@ export async function getCarLink (files, gatewayUrl, ipfs) {
...
+  const cidStr = cid.toString()
+  const now = Date.now()
+  const lastProvideTime = provideCache.get(cidStr)
+
+  if (lastProvideTime && (now - lastProvideTime) < PROVIDE_DEBOUNCE_TIME) {
+    return
+  }
+
+  try {
+    // @ts-ignore - ipfs is actually a KuboRPCClient with routing API
```
SgtPooki, line 175: yeaaaa... we need to remove @types/ipfs and all use of `'ipfs'` for types and imports.

```suggestion
    // @ts-expect-error - ipfs is actually a KuboRPCClient with routing API
```

src/lib/files.js
```diff
@@ -152,6 +152,46 @@ export async function getCarLink (files, gatewayUrl, ipfs) {
...
+ *
+ * @param {CID} cid - The CID to provide to the network
+ * @param {IPFSService} ipfs - The IPFS service instance
+ */
+export async function debouncedProvide (cid, ipfs) {
+  const cidStr = cid.toString()
+  const now = Date.now()
+  const lastProvideTime = provideCache.get(cidStr)
+
+  if (lastProvideTime && (now - lastProvideTime) < PROVIDE_DEBOUNCE_TIME) {
```
SgtPooki, line 170:
```suggestion
  if (lastProvideTime != null && (now - lastProvideTime) < PROVIDE_DEBOUNCE_TIME) {
```

src/lib/files.js
```diff
@@ -152,6 +152,46 @@ export async function getCarLink (files, gatewayUrl, ipfs) {
...
+
+  try {
+    // @ts-ignore - ipfs is actually a KuboRPCClient with routing API
+    const provideEvents = ipfs.routing.provide(cid, { recursive: false })
+
+    for await (const event of provideEvents) {
+      console.debug(`[PROVIDE] ${cidStr}:`, event)
+    }
+
+    provideCache.set(cidStr, now)
```
SgtPooki, line 182: we should set the new entry after we clean old ones just to prevent the extra check

SgtPooki (changes requested):
lgtm but a few minor changes requested

SgtPooki (approved):

PR #2389, opened 2025-07-23
fix: eslint doesnt check build folder

SgtPooki:
FYI that eslint was passing in CI because it runs before build, but when running locally i was getting hundreds of thousands of errors.. debugging with `DEBUG=eslint:cli-engine npm run eslint` showed that the build folder was being checked. when I removed that, it started succeeding again.

Also removed the ignoring of .ts files in the `.eslintrc.cjs` file

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 !.storybook
+build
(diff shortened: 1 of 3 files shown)


SgtPooki (commented):
self review

PR #2390, opened 2025-07-24
Migrating icons to typescript

Pfed-prog:
Removed unused react imports and migrated to tsx
Inspired by https://github.com/ipfs/ipfs-webui/issues/2349


Pfed-prog:
I apologize for making a lot of commits and being all over the place. I definitely would need help @SgtPooki

SgtPooki:
FYI icons are generated by `lol` and we should either update that library or use a similar one that generates typescript. I think we can ignore the icons because they are generated.

PR #2392, opened 2025-07-28
feat: add diagnostics screen

SgtPooki:
Adds diagnostics screen to the webui.

The page has tabs support, but only the logs tab is implemented currently.

Related https://github.com/ipfs/go-log/pull/160
Related https://github.com/ipfs/kubo/pull/10885


lidel:
This is very good, thank you @SgtPooki ❤️ 
I think we will polish UI/UX but performance is more than 

I've pushed some changes for small papercuts I noticed, one of them impacts fonts/responsive layout, would be good if you took a look if they look ok on macOS:

1. 0e566ef1 - small perf. fix by reducing unnecessary redraws (reuses IDs created by IndexedDB + memo)
2. f77d0933 - puts all stats on one line to save vertical space
3. 7867d572 - moves trash icon to bottom-right, and the scroll button up top, below play/pause
4. 06e7cd80 - makes gray text darker for readability (improves contrast)
5. 4554d7fe - blurs background when logs are expanded fullscreen
6. cb96b0a3 - unified monospace font everywhere (looks better on my linux box, but lmk)
7. b2f5008c - adjusts layout heights for mobile vs desktop menus


I don't know what to do with the top, it feels we use a lot of empty space around the filter UI, but I think we will tackle it once we have the second tab with CID Checker. 

Do you want to add it here, or is the plan to merge this (as log-only) and open a second PR to add Checker?

lidel:
Pushed some cosmetic changes improving UX, cleaning up some labels

## Preview

Moved Diagnostics to the end of navbar to not disturb muscle memory people have after half of decade :) 

### Opening log screen...

..now has visual indicator what to click + button to restore default error level, and more user friendly header

> 
> 
> 

## Retrieval check integrated with Files screen

Clicking on context menu now has "Check Retrieval" option which will open ipfs-check with CID prefilled:

> 
>
> 
> 
>

lidel (approved):
Lgtm, thank you @SgtPooki  I think this is as good as we could make it given the time allocated.

I'm going to merge it and cut a release later this week, so we can dogfood it bit more in Kubo 0.38 RC1

PR #2399, opened 2025-08-15
refactor: migrate versionLink to typescript

SgtPooki:
Part of https://github.com/ipfs/ipfs-webui/pull/2392 where I needed to re-use the agent parsing logic but it was all unexportable and done in the VersionLink component.


.eslintignore
```diff
@@ -1,2 +1,3 @@
 !.storybook
 build
+storybook-static
```
SgtPooki, line 3: i keep running into this when i infrequently run storybook and don't clean it up..

SgtPooki (commented):
self review

PR #2404, opened 2025-08-22
refactor: typescript api-address-form

SgtPooki:
part of #2392.. migrating files to typescript that will eventually be typechecked


SgtPooki (commented):
self review

PR #2408, opened 2025-08-24
fix: reference to `global` in browser environment

ozrg:
replaced `global.location.pathname` with `window.location.pathname` since `global` is only available in Node.js.
this resolves the runtime error in the browser and ensures correct path handling in the React app.


SgtPooki (approved):
oops. I did that.. thanks for the quick fix.

PR #2421, opened 2025-09-06
feat: add pinned-first sorting option to files section

alikhere:
## Summary
Implements a new sorting feature for the Files section that allows users to sort files with pinned items appearing first, while maintaining default sorting behaviour.

## Features Added
- **Sort Dropdown**: Clean dropdown UI with two options:
  - "Name (A–Z)" - Default(current) behavior (folders first, then files, alphabetically)
  - "Pinned first" - Pinned items at top,
- **Persistence**: Both sort preference and view mode (list/grid) are saved to localStorage
- **Auto-sorting**: When "Pinned first" is selected, newly pinned items automatically move to top
- **Backwards Compatible**  For those who prefer current behavior

## Implementation
- Enhanced `sortFiles` function with pinned-first logic
- Added `BY_PINNED` sort option to existing sorting system
- Implemented localStorage persistence for user preferences
- Added TypeScript type support for new sort option
- Created accessible dropdown component following existing UI patterns

## Before 
pinned-before

## After
pinned-after

Fixes: #2407


alikhere:
@SgtPooki  @lidel  Could you please review the PR

lidel (approved):
Thank you @alikhere, applied some changes, but lgtm.

Realized we already had partially implemented sorting UI in the list mode, when you clicked on column names. 

What I did I've wired up sorting by pin there:  

> 


..and then added option to sort by size to the dropdown, and then adding asc/desc version of each.  I also added option to disable sorting and show original order as it is in the UnixFS DAG:

>

alikhere:
> Thank you @alikhere, applied some changes, but lgtm.
> 
> Realized we already had partially implemented sorting UI in the list mode, when you clicked on column names.
> 
> What I did I've wired up sorting by pin there:
> 
> > 
> 
> ..and then added option to sort by size to the dropdown, and then adding asc/desc version of each. I also added option to disable sorting and show original order as it is in the UnixFS DAG:
> 
> > 
> 
> Once i get CI to be green, I'm merging as-is but feel free to refine / improve in a follow-up PR.

Looks great now!

PR #2425, opened 2025-09-10
fix(navbar):  highlighting on browser back

### LINKED ISSUE #2423: bug: navbar highlighting doesn't update properly on browser back
**Describe the bug**
When you are on a particular page, the navbar is highlighted behind the current item, with a border.

If you click back, then two items are highlighted (change in background color to a muted color), but the border bottom is only on the correct one.

Found when working on diagnostics page, but happens on main:

https://github.com/user-attachments/assets/13204f63-0d6f-4562-b999-e5fb64f1415b

SuchitraSwain:
Closes #2423 

https://github.com/user-attachments/assets/b967e625-32dc-4b70-a2fc-d580c06df93a


lidel (approved):
Thank you @SuchitraSwain, works as expected.

Small nits inline for the future 

Also, it is useful to use "Closes #issue-number" phrasing, this way merging PR automatically resolves the issue it fixes:

- https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

(I made change here alread, juts fyi in the future)

PR #2426, opened 2025-09-10
fix: shared max for bandwidth chart

### LINKED ISSUE #2417: feat: bandwidth chart radial charts should use same reference point
Currently, the radial charts on the right are based on MAX and CURRENT of their particular channel (input/output)

however, I feel like this is much less valuable than if they were based on the same bandwith scale. Whether that be a moving average or a min/max. 

#### Current:

Image

this makes it look (at first glance) like outgoing bandwidth is slightly higher than incoming bandwidth, but the reality is that outgoing is significantly less than incoming (8.17% of incoming bandwidth max `(251×1024/3×1024×1024)×100=(257,024/3,145,728)×100≈8.17%`. We should have a MAX bandwidth over the datapoints of input AND output, and base the radial charts on that.

#### Potential:

https://github.com/user-attachments/assets/3dc07f50-347d-4b90-aafd-ef56ad1e7679

with this video, bandwidth guages are based on `currentRateIn + currentRateOut` which makes the meaning much easier to determine.

SuchitraSwain:
- Fixes #2417 

image


lidel (approved):
Thank you, pushed cosmetic changes, but overall lgtm. 

Some tips for  future contributions in comments inline.

PR #2428, opened 2025-09-10
fix: overlay click does not close modals

### LINKED ISSUE #2414: bug: overlay click does not close modals
We have a click handler in our `react-overlay` called onBackdropClick but it's not properly handling click, because the visible modal is capturing all of the clicks.

I attempted a quick fix but some style issues were blocking so I left it..

Tested on main and on https://github.com/ipfs/ipfs-webui/pull/2413, and the issue is in both.

SuchitraSwain:
- Closes #2414 

[488204972-43487ba6-dc29-4923-a257-5a2cd717ca64.webm](https://github.com/user-attachments/assets/1dbb39ef-0372-4200-a755-cf947cb1d516)

### DIFF
--- a/test/e2e/settings.test.js
+++ b/test/e2e/settings.test.js
@@ -45,8 +45,6 @@ test.describe('Settings screen', () => {
   })
 
   test('Submit/Reset Public Subdomain Gateway', async ({ page }) => {
-    // Increase timeout for this test as validation can be slow
-    test.setTimeout(45000)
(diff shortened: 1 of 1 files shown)


lidel (approved):
Thank you @SuchitraSwain, works as expected. 

- Pushed cosmetic changes making tests easier to maintain. 
- Removed timeouts -- iiuc you were actually lowering it, because the default for CI is 90 seconds.

PR #2435, opened 2025-09-15
feat/diagnostics screen check tab

SgtPooki:
- **feat: add ipfs-check tab to diagnostics screen**
- **chore: use relative import**
- **fix: useDebouncedCallback doesnt lose function call on delay change**
- **fix: ipfs check iframe uses query params**

### DIFF
--- a/src/diagnostics/check-screen/check-screen.tsx
+++ b/src/diagnostics/check-screen/check-screen.tsx
@@ -1,8 +1,20 @@
-import React, { useRef, useEffect } from 'react'
+import React, { useRef, useEffect, useMemo } from 'react'
 import { useDebouncedCallback } from '../../lib/hooks/use-debounced-callback'
+import { useBridgeSelector } from '../../helpers/context-bridge'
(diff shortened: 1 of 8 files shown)


src/diagnostics/check-screen/check-screen.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const routeInfo = useBridgeSelector<RouteInfo>('selectRouteInfo')
+  // pass through IPFS check query params.
+  const checkIframeUrl = useMemo(() => {
+    if (routeInfo == null) {
+      return 'https://check.ipfs.network/'
+    }
+    const url = new URL(routeInfo.url.replace('/diagnostics/check', ''), 'https://check.ipfs.network/')
+
+    return url.toString()
+  }, [routeInfo])
```
SgtPooki, line 17: @lidel I added this as i was updating the check screen.. i see you merged that PR and added just a CID pass through. will close this PR

PR #2482, opened 2026-03-10
feat: choose share link type, add native uris

lidel:
## Problem

Share Link always built a URL on a third-party public gateway (dweb.link, with ipfs.io fallback) and Publish to IPNS hardcoded the public path gateway. There was no way to copy a native `ipfs://` address, opt out of public gateways, or get a link pointing at your own node for other apps on the same machine (ipfs/ipfs-desktop#2500). The webui also probed gateways with image loads on startup and in Settings.

## Fix

- Settings gains a "Sharing IPFS Links" section used by both Share Link and Publish to IPNS: native `ipfs://` addresses, local gateway links, or public gateway links (path or subdomain)
- defaults are unchanged: `ipfs.io` and `dweb.link` stay prefilled and share links are built on `dweb.link` as before; flipping the default to native URIs is a follow-up where `DEFAULT_SHARE_LINK_TYPE` and the `DEFAULT_*_GATEWAY` constants are the only knobs
- clearing a public gateway in Settings now opts out of it entirely; a link type whose gateway is missing falls back to a native address, so copied links are never empty or broken
- Share modal offers a local HTTP link checkbox, with optional localhost subdomains for web apps that need origin isolation; links labeled local draw from the local gateway only, never a public fallback
- gateway reachability probes removed; URL format is validated and the user's choice trusted

This is part of wider URI support to move away from copying public gateway links by default:

- ipfs/kubo#11375
- ipfs/ipfs-companion#1371

## Demo

New "Sharing IPFS Links" section on the Settings screen:

> 

The Share modal looks the same as before, with a new opt-in checkbox for a local link:

> 

Once clicked, we update UI to indicate this is no longer for sharing with friends, and only other apps on the same machine. We use IP-based Path gateway that is known to work everywhere:

> 

Subdomain gateways do not work on all platforms, and only benefit web contexts (browser web apps, websites with relative paths etc), so are opt-in:

>

### DIFF
--- a/public/locales/en/app.json
+++ b/public/locales/en/app.json
@@ -46,13 +46,13 @@
     "placeholder": "Enter a URL (http://user:password@127.0.0.1:5001) or a Multiaddr (/ip4/127.0.0.1/tcp/5001)"
   },
   "publicGatewayForm": {
-    "placeholder": "Enter a URL (https://ipfs.io)"
+    "placeholder": "Enter an HTTP URL (https://path-gw.example.com)"
   },
   "localGatewayForm": {
-    "placeholder": "Enter a URL (http://localhost:8080)"
+    "placeholder": "Enter an HTTP URL (http://localhost:8080)"
   },
   "publicSubdomainGatewayForm": {
-    "placeholder": "Enter a URL (https://dweb.link)"
+    "placeholder": "Enter an HTTP URL (https://subdomain-gw.example.net)"
   },
   "ipfsCheckForm": {
     "label": "Retrieval Check Service URL",
@@ -90,7 +90,8 @@
     "pinStatus": "Pin Status",
     "publicKey": "Public key",
     "publicGateway": "Public Gateway",
-    "localGateway": "Local Gateway",
+    "publicSubdomainGateway": "Public Subdomain Gateway",
+    "localGateway": "Local HTTP Gateway",
     "rateIn": "Rate in",
     "rateOut": "Rate out",
     "repo": "Repo",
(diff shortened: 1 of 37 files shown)


guillaumemichel (approved):
Tested & approved

lidel:
Pushed some related tests / changes since review:

- Share Link and Publish to IPNS now default to native `ipfs://cid` (and `ipns://name`). Public gateways are a conscious opt-in: nothing goes through `ipfs.io` or any other third party unless you configure a public gateway yourself.
- Added a "Sharing IPFS Links" settings section to choose the link type, with e2e tests for the default and each option.

So the UI still makes sense the day `ipfs.io` and other public gateways go away:

This is part of wider URI support to move away from copying public gateways by default:
- https://github.com/ipfs/kubo/pull/11375
- ipfs-companion will have similar one

## Preview of Settings page

Screenshot 2026-07-02 at 00-39-36 Settings IPFS

lidel:
This PR no longer changes any defaults, so it can merge without touching existing behavior.

Changes since my last update:

- c33c8a7c keeps today's defaults: `ipfs.io` path and `dweb.link` subdomain gateways prefilled, share links built on `dweb.link` as before. Native-by-default moves to a follow-up PR; `DEFAULT_SHARE_LINK_TYPE` and the `DEFAULT_*_GATEWAY` constants are the only knobs to flip.
- Opting out still works: a gateway cleared in Settings stays cleared, and links fall back to native addresses.
- Picking a public link type persists the gateway it points at, so an explicit choice survives a future change of defaults.
- Merged latest main.

## Preview of Settings page

>

PR #2493, opened 2026-06-30
fix(ipns): surface publish errors to user

### LINKED ISSUE #2492: Publishing a File to IPNS Hangs after Throwing Internal Server Error 500
**Describe the bug**

After pressing "Publish" in the IPNS publishing modal of the webui the process never completes despite the finishing progress bar.

**To Reproduce**
1. Import some file. Optionally pin it.
2. Open the file's three dots menu, select "Publish to IPNS", choose the default IPNS key ("self"), open devtools, press "Publish", see the 500 internal server error in the console.

**Expected behavior**

IPNS publishing completes and reports success or failure.

**Actual behavior**

1) The progress bar finishes but the process actually hangs because of the 500 error.
2) Nothing reported to the user.

**Screenshots**

Image

Image

**Desktop:**

 - OS: Alpine Linux
 - Browsers: FireFox 151.0.3 and Chromium 149.0.7827.155.
 - WebUI Version: v4.11.1 - https://github.com/ipfs-shipyard/ipfs-webui/commit/d1e5cd2

lidel:
Publishing to IPNS failed silently: the modal stayed on a finished-looking progress bar and the RPC error was never shown, so the operation looked stuck with no way forward.

- ipns.js: catch name.publish failures, emit a notify toast carrying the original error, and re-throw so the modal can react
- notify: render IPNS_PUBLISH_FAILED as an error toast using the ipnsPublishFailReason message
- PublishModal: reset the progress state on failure so the inline error renders and Publish re-enables for a retry

Closes #2492 reported by @ilyaigpetrov

### DIFF
--- a/public/locales/en/notify.json
+++ b/public/locales/en/notify.json
@@ -7,6 +7,7 @@
   "ipfsPinFailReason": "Unable to set pinning at {serviceName}: {errorMsg}",
   "ipfsPinSucceedReason": "Successfully pinned at {serviceName}",
   "ipfsUnpinSucceedReason": "Successfully unpinned from {serviceName}",
+  "ipnsPublishFailReason": "Unable to publish to IPNS under key {keyName}: {errorMsg}",
   "ipfsIsBack": "Normal IPFS service has resumed. Enjoy!",
   "folderExists": "An item with that name already exists. Please choose another.",
   "filesFetchFailed": "Failed to get those files. Please check the path and try again.",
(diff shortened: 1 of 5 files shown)


hiSandog:
This fix matches the reported UX problem well. One edge case worth covering in the test is the retry path after a failed publish: fail `name.publish`, assert the inline error/toast appears and the progress state resets, then trigger Publish again and verify the modal can proceed instead of staying in a terminal failed/progress state. That would protect the "Publish re-enables for a retry" behavior described in the PR.

PR #2496, opened 2026-07-12
feat: add Auto/Light/Dark theme support (fixes #1702)

### LINKED ISSUE #1702: Feature: dark mode

kushmunjal:
Fixes #1702

## Summary

This adds a three-way theme setting — Auto, Light, and Dark — to ipfs-webui.
Auto follows the OS-level `prefers-color-scheme` and updates live if the OS
setting changes; Light/Dark are manual overrides, persisted across sessions.

The toggle is available in two places: a compact icon in the top bar for
quick switching, and a full selector on the Settings page.

## Why I approached it this way

I know dark mode was attempted once before in #2322 and reverted in #2339
after several screens/modals were found unreadable post-merge. I read through
that history before starting, and tried to design around the specific gap
that caused the revert — rather than hand-editing individual components, I
introduced CSS custom properties for color (`--element-bg`, `--text`,
`--border-color`, etc., swapped per theme via a `[data-theme="dark"]`
selector) so shared utility classes and components get themed centrally, and
went through every screen and every modal/dropdown/menu I could find, not
just the main page layouts.

## What changed

- `ThemeContext`/`ThemeProvider` (`src/contexts/theme-context.js`) — tracks
  the active theme, listens for OS-level scheme changes when set to Auto,
  persists the user's explicit choice to localStorage.
- `TopBarThemeToggle` and `ThemeSelector` components for switching the theme
  from the top bar and from Settings respectively.
- Removed hardcoded inline colors across [Box, ProgressBar, SelectedActions,
  FileImportStatus, Breadcrumbs, PublishModal, modal.tsx, App.js header,
  NodeBandwidthChart, Speedometer, WorldMap] in favor of theme-aware CSS
  classes/custom properties.
- Fixed a pre-existing bug where the Settings page rendered a raw
  `themeDescription` i18n key instead of translated text.
- World map on the Peers page uses a CSS filter to invert/desaturate the
  static map image for dark mode rather than shipping a second image asset.

## Explicitly out of scope for this PR

The Explore/DAG page's card styling is rendered by the separate
`ipld-explorer-components` package and needs its own companion PR there
(same pattern as the original #2322 effort) — I'm treating that as a
follow-up rather than bundling it in here, since it's a different repo with
its own release cycle.

## Testing

- [ ✓ ] `npm run lint` — passing as of <83d9e2b>
- [ ✓ ] `npm run test:unit` — passing as of <83d9e2b>
- [ ✓ ] `npm run test:e2e` — passing as of <83d9e2b>
- [ ✓ ] Manually verified every screen (Status, Files, Peers, Settings,
      Diagnostics) plus modals/dropdowns/selection states in both Light and
      Dark, including toggling OS-level dark mode live for Auto
- [ ✓ ] Confirmed `package.json`/`package-lock.json` have no unrelated
      dependency changes

## Open questions / things I'd appreciate a second pair of eyes on

- [Fill in honestly — e.g. "I'm not 100% sure the `background` prop I'm
  passing into `@tableflip/react-dropdown` is officially supported by that
  library vs. just being ignored — I confirmed via devtools that it renders
  correctly, but wanted to flag the assumption in case there's a cleaner way."]
- [Any other genuine uncertainty — better to name it than have a reviewer
  find it, especially given this feature's history.]

Given the history on this issue, I'd rather get feedback early and iterate
than have this merge and need another revert — happy to adjust scope,
approach, or split this into smaller PRs if that's easier to review.


src/components/button/button.tsx
```diff
@@ -15,7 +15,7 @@ export interface ButtonProps extends Omit<React.ButtonHTMLAttributes<HTMLButtonE
 
 const getButtonClassName = ({ fill, bg, color, danger, disabled }: Pick<ButtonProps, 'fill' | 'bg' | 'color' | 'danger' | 'disabled'>, type: ButtonProps['type']) => {
   if (danger) return 'bg-red fill-white white'
-  if (disabled) return 'bg-gray-muted fill-snow light-gray'
+  if (disabled) return 'bg-black-10 fill-charcoal-muted charcoal-muted'
```
lidel: iiuc this changes disabled buttons in light mode too, and `bg-black-10` is close to invisible on a dark background. Consider keeping the old classes and adding a dark override for `.bg-gray-muted` in index.css instead.

src/components/is-not-connected/is-not-connected.tsx
```diff
@@ -42,7 +42,7 @@ const IsNotConnected = () => {
         { !isSameOrigin && (
           <div>
             <Trans i18nKey='notConnected.paragraph3' t={t}>
-              <li className='mb3 mt4'>Is your Kubo RPC API configured to allow <a className='link blue' href='https://github.com/ipfs/ipfs-webui#configure-kubo-rpc-api-cors-headers'>cross-origin (CORS) requests</a>? If not, run these commands and then start your daemon from the terminal:</li>
+              <li className='mb3'>Is your Kubo RPC API configured to allow <a className='link blue' href='https://github.com/ipfs/ipfs-webui#configure-kubo-rpc-api-cors-headers'>cross-origin (CORS) requests</a>? If not, run these commands and then start your daemon from the terminal:</li>
```
lidel, line 45: The margin changes in this file look unrelated to the theme work. Mind reverting them, or explaining why they are needed?

src/components/theme-selector/ThemeSelector.js
```diff
@@ -0,0 +1,34 @@
...
+const ThemeSelector = ({ t }) => {
+  const { theme, setTheme } = useTheme()
+
+  const onChange = (e) => {
+    setTheme(e.target.value)
+  }
+
+  return (
+    <div className='joyride-settings-theme'>
+      <Title>{t('theme') || 'Theme'}</Title>
```
lidel, line 15: There is no `theme` key in settings.json, and i18next returns the key itself when one is missing, so this title renders as literal "theme" (the `|| 'Theme'` fallback never runs; same for the `themeDescription` fallback below). Please add the key, and use `t()` for the Auto/Light/Dark labels too so they can be translated.

src/components/theme-selector/TopBarThemeToggle.js
```diff
@@ -0,0 +1,52 @@
...
+  let Icon = StrokeMonitor
+  if (theme === 'light') Icon = SunIcon
+  if (theme === 'dark') Icon = MoonIcon
+
+  return (
+    <button
+      className="button-reset bg-transparent bn p0 m0 mr3 flex items-center justify-center pointer charcoal-muted hover-navy transition-all glow"
+      onClick={cycleTheme}
+      title={t('themeDescription')}
+      style={{ width: 28, height: 28, outline: 'none' }}
```
lidel: ux nit: `outline: 'none'` hides the focus ring for keyboard users, and the button has no text for screen readers.

Please use the existing `focus-outline` classes instead, and add an `aria-label` naming the current theme.

src/contexts/theme-context.js
```diff
@@ -0,0 +1,58 @@
...
+// Default theme is auto
+const DEFAULT_THEME = 'auto'
+
+/**
+ * @param {Object} props
+ * @param {import('react').ReactNode} props.children
+ */
+export const ThemeProvider = ({ children }) => {
+  const [theme, setThemeState] = useState(() => {
+    return localStorage.getItem('ipfs-webui-theme') || DEFAULT_THEME
```
lidel: Any string in localStorage becomes `theme` unchecked, and the Settings select renders blank for unknown values. localStorage access can also throw in private browsing / tor browser under certain modes

Please validate against `['auto', 'light', 'dark']` and wrap localStorage access in try/catch that defaults to auto on unknown value

src/contexts/theme-context.js
```diff
@@ -0,0 +1,58 @@
...
+  })
+
+  useEffect(() => {
+    const mediaQuery = window.matchMedia
+      ? window.matchMedia('(prefers-color-scheme: dark)')
+      : { matches: false, addEventListener: () => {}, removeEventListener: () => {} }
+
+    const applyTheme = () => {
```
lidel, line 67: nit: when the theme is `auto` and the OS scheme changes, only the DOM attribute updates; the context value stays `'auto'`, so consumers do not re-render. Speedometer.js and NodeBandwidthChart.js read the attribute during render, so their chart colors go stale until the next data update.

Perhaps expose the resolved light/dark value through the ontext; that fixes both?

src/contexts/theme-context.js
```diff
@@ -0,0 +1,58 @@
...
+      const isDark = theme === 'dark' || (theme === 'auto' && mediaQuery.matches)
+      document.documentElement.setAttribute('data-theme', isDark ? 'dark' : 'light')
+    }
+
+    applyTheme()
```
lidel, line 74: The theme is applied only after React mounts, so dark-mode users see a white flash on every load.

One way to fix it is a tiny script in public/index.html, ahead of the bundle, can set `data-theme` from localStorage and matchMedia (to avoid that white flash on slow machines).

src/files/dropdown/Dropdown.js
```diff
@@ -2,17 +2,17 @@ import React, { forwardRef } from 'react'
 import { Dropdown as Drop, DropdownMenu as Menu } from '@tableflip/react-dropdown'
 import StrokeCode from '../../icons/StrokeCode.js'
 
-export const Option = ({ children, onClick, className = '', isCliTutorModeEnabled, onCliTutorMode, ...props }) => (
+export const Option = ({ children, onClick, className = '', isCliTutorModeEnabled, onCliTutorMode, disabled, ...props }) => (
```
lidel: ⚠️  Nothing in this PR passes `disabled`, so I think this slipped in from other work (the two branches also differ: real `disabled` vs `aria-disabled`, and `not-allowed` is not a tachyons/ipfs-css class).

Would you mind moving it to its own PR and explain there why you do this? Keeps this one focused on theming.

src/files/sort-dropdown/SortDropdown.js
```diff
@@ -93,7 +93,7 @@ const SortDropdown = ({ currentSort, onSortChange, t }) => {
       {isOpen && (
         <DropdownMenu
           open={isOpen}
-          className="sans-serif br2 charcoal"
+          className="sans-serif br2 charcoal dropdown-menu"
```
lidel, line 96: Same story as Dropdown.js: this file imports `DropdownMenu` straight from the library, so the arrow stays white in dark mode. `src/components/context-menu/ContextMenu.js` (the "..." menus for pinning services and IPNS keys on Settings) does the same, though it is not part of this diff.

Adding `background='var(--element-bg)'` in both places should cover it.

src/files/dropdown/Dropdown.js
```diff
@@ -22,7 +22,7 @@ export const DropdownMenu = forwardRef((props, ref) => {
 
   return (
     <Menu
-      className='sans-serif br2 charcoal'
+      className='sans-serif br2 charcoal dropdown-menu'
```
lidel, line 25: The library fills the menu arrow with inline `background: 'white'` by default, so in dark mode menus show a white triangle above the dark panel.

> Image

Passing `background='var(--element-bg)'` here should theme both the panel and the arrow for everything using this wrapper, and then the `!important` rule in index.css can go too.

src/index.css
```diff
@@ -61,10 +61,159 @@ button,
...
+  --charcoal-muted: #7f8491;
+  --green-muted: #eaf5e6;
+
+  --overlay-bg: rgba(255, 255, 255, 0.5);
+  --publish-modal-gradient: linear-gradient(90deg, rgba(255, 255, 255, 0.35), #ffffff 10%, #ffffff 100%);
+  --breadcrumb-gradient: linear-gradient(to right, #ffffff 0%, transparent 100%);
+  --tooltip-bg: #ffffff;
+
+  /* File selection */
+  --file-row-selected-bg: #daedf5;
```
lidel: Small one: this changes the light-mode selection color, rows were `#F0F6FA` before (see File.js).

If that was not intentional, keeping `#F0F6FA` here restores the old look.
If was intentional, why we changed them?

src/index.css
```diff
@@ -61,10 +61,159 @@ button,
...
+  --action-bar-border: #CFE0E2;
+  --action-bar-text: #59595A;
+  --action-bar-muted: #A4BFCC;
+}
+
+[data-theme="dark"] {
+  --background: #181a1b;
+  --text: #e8e6e3;
+  --element-bg: #1a1c1e;
+  --navy-dark: #0b3a53;
```
lidel, line 109: Looks like `--navy-dark`, `--navy-text-color` and `--gray-muted` ended up unused, no rule references them?

Maybe add comment noting they are part of pallette so refactor does not drop them (or wire them up if they were meant for something).

src/files/file-import-status/FileImportStatus.js
```diff
@@ -160,9 +160,8 @@ export const FileImportStatus = ({ filesFinished, filesPending, filesErrors, doF
           onClick={() => setExpanded(!expanded)}
           onKeyPress={handleExpandByKeyboard}
           role="button"
-          className="fileImportStatusButton pv2 ph3 relative flex items-center no-select pointer charcoal w-100 justify-between"
+          className="fileImportStatusButton pv2 ph3 relative flex items-center no-select pointer charcoal w-100 justify-between bg-snow"
```
lidel, line 163: Same here, white triangle should be fixed too:

> Image

src/index.css
```diff
@@ -61,10 +61,159 @@ button,
...
+[data-theme="dark"] .bg-white,
+[data-theme="dark"] .bg-white-80 { background-color: var(--element-bg) !important; }
+[data-theme="dark"] .bg-snow,
+[data-theme="dark"] .bg-snow-muted { background-color: var(--snow-muted) !important; }
+[data-theme="dark"] .bg-near-white { background-color: var(--snow-muted) !important; }
+[data-theme="dark"] .bg-light-gray { background-color: var(--snow-muted) !important; }
+[data-theme="dark"] .bg-white-70 { background-color: var(--element-bg) !important; opacity: 1 !important; }
+[data-theme="dark"] .charcoal { color: var(--charcoal) !important; }
+[data-theme="dark"] .charcoal-muted { color: var(--charcoal-muted) !important; }
+[data-theme="dark"] .navy { color: var(--text) !important; }
```
lidel: Tiny one: `.navy` is overridden twice in this file (here and in the heading block below).

Could we simplify (e.g. one of the two can go) or is this intentional?

src/index.css
```diff
@@ -61,10 +61,159 @@ button,
...
+[data-theme="dark"] .dark-gray { color: var(--text) !important; }
+
+/* Map theming */
+[data-theme="dark"] .world-map {
+  filter: invert(0.85) hue-rotate(180deg) opacity(0.6);
+}
+
+[data-theme="dark"] .webui-header { background-color: var(--snow-muted) !important; }
```
lidel, line 160: This one is the most visible issue I found: `.webui-header` has no light-mode rule, and App.js no longer sets the inline background, so the top bar lost its `#F0F6FA` background in light mode (the default for most users).

A base rule here, or a `--header-bg` variable with a dark override, keeps light mode looking as before.

> Image

src/index.css
```diff
@@ -61,10 +61,159 @@ button,
...
+
+/* Tachyons borders */
+[data-theme="dark"] .b--black-20 { border-color: var(--border-color) !important; }
+[data-theme="dark"] .b--black-10 { border-color: var(--border-color) !important; }
+[data-theme="dark"] .b--near-white { border-color: var(--border-color) !important; }
+
+/* Dropdown / Modal specifics */
+[data-theme="dark"] .shadow-4 { box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.8) !important; }
+/* The background prop on @tableflip/react-dropdown injects var(--element-bg) as an inline
```
lidel, line 187: I could not find any code passing a `background` prop to the dropdown, so this comment describes something the PR does not do.. yet?

Assuming you will add it, once the prop is added in Dropdown.js (see my comment there), this rule and the comment can both go.

src/settings/SettingsPage.js
```diff
@@ -125,6 +126,10 @@ export const SettingsPage = ({
         <LanguageSelector t={t} />
       </div>
 
+      <div className='mt4 joyride-settings-theme'>
```
lidel: Tiny one: `joyride-settings-theme` is set both here and on ThemeSelector's root div.

One of the two is enough. Also, the white triangle in Joyride popups on every page:

> Image

src/index.css
```diff
@@ -61,10 +61,159 @@ button,
...
+  --navy-text-color: rgb(202, 198, 191);
+  --snow-muted: rgb(28, 30, 31);
+  --border-color: rgba(140, 130, 115, 0.5);
+  --green-muted: #1e3a1e;
+
+  --overlay-bg: rgba(0, 0, 0, 0.5);
+  --publish-modal-gradient: linear-gradient(90deg, rgba(31, 38, 46, 0.35), var(--element-bg) 10%, var(--element-bg) 100%);
+  --breadcrumb-gradient: linear-gradient(to right, var(--element-bg) 0%, transparent 100%);
+  --gray-muted: #d9dbe2;
+  --charcoal: #c8c3bc;
```
lidel, line 120: From testing the branch in dark mode: on the Status page welcome box, the body text and the `link blue` links are hard to read (see screenshot). Contrast this low is an accessibility problem, WCAG AA expects 4.5:1 for text.

The text comes from `charcoal` and inherited colors in IsConnected.js, AboutWebUI.js and AboutIpfs.js, so the fix belongs here in the dark palette: brighter `--charcoal` / `--charcoal-muted` values, plus a dark override for `.link.blue`. Running the final values through a contrast checker would confirm AA.


> Image

src/files/file/File.js
```diff
@@ -82,7 +82,7 @@ const File = ({
   const styles = { height: 55, overflow: 'visible' }
 
   if (focused || (selected && !translucent) || coloured || (isOver && canDrop)) {
-    styles.backgroundColor = '#F0F6FA'
+    className += ' file-row-selected'
```
lidel, line 85: From testing in dark mode: the separators between file rows come out almost white and very bright. They should be dark grey, like the other borders.

Two sources: the row root above uses `b--light-gray` (which has no dark override), and further down this file sets an inline `borderTop: '1px solid #eee'`, which no CSS override can reach. Suggest dropping the inline border in favor of the class, and adding `[data-theme="dark"] .b--light-gray { border-color: var(--border-color) !important; }` in index.css.

src/index.css
```diff
@@ -61,10 +61,159 @@ button,
...
+
+/* Override Tachyons/IPFS-CSS utility classes */
+[data-theme="dark"] .bg-white,
+[data-theme="dark"] .bg-white-80 { background-color: var(--element-bg) !important; }
+[data-theme="dark"] .bg-snow,
+[data-theme="dark"] .bg-snow-muted { background-color: var(--snow-muted) !important; }
+[data-theme="dark"] .bg-near-white { background-color: var(--snow-muted) !important; }
+[data-theme="dark"] .bg-light-gray { background-color: var(--snow-muted) !important; }
+[data-theme="dark"] .bg-white-70 { background-color: var(--element-bg) !important; opacity: 1 !important; }
+[data-theme="dark"] .charcoal { color: var(--charcoal) !important; }
```
lidel, line 146: From testing in dark mode: the view-toggle and search icons next to the sort dropdown on the Files page are nearly invisible, navy on near-black (see screenshot). Many people will simply not see them, so this is an accessibility problem too.

They get their color from `.selected-item` (`color: #0b3a53` in file-thumbnail.css) via `fill="currentColor"`. One more override next to the text colors here, e.g. `[data-theme="dark"] .selected-item { color: var(--charcoal) !important; }`, makes them visible again.

> Image

src/index.css
```diff
@@ -61,10 +61,159 @@ button,
...
+
+/* Map theming */
+[data-theme="dark"] .world-map {
+  filter: invert(0.85) hue-rotate(180deg) opacity(0.6);
+}
+
+[data-theme="dark"] .webui-header { background-color: var(--snow-muted) !important; }
+
+/* Tachyons borders */
+[data-theme="dark"] .b--black-20 { border-color: var(--border-color) !important; }
```
lidel, line 181: From testing in dark mode: on the Diagnostics screen, the line under the Logs / Retrieval Check / DHT Provide tabs stays white (see screenshot).

It is a hardcoded inline `borderBottom: '1px solid #e1e5eb'` in diagnostics-content.tsx, so it needs a small change in that file: reuse the same border color inputs get in dark mode (`var(--border-color)`, as in the `b--black-20` override here) instead of the fixed light value.


> Image

lidel (changes requested):
Thanks for picking this up @kushmunjal, and for reading the history around #2322 before starting. CSS variables with `[data-theme]` overrides is the piece the previous attempt was missing, and shared surfaces get themed the right way here. 👍

Because the last attempt merged and then had to be reverted (#2339), I'm extra cautious about quality here: we don't want to ship a half-baked solution again. As noted in https://github.com/ipfs/ipfs-webui/issues/1702#issuecomment-2686363521, dark mode has to include the Explore screen. The overrides in `index.css` are global, so they also restyle what `ipld-explorer-components` renders there: the page comes out partially themed, which is the exact failure mode that caused the revert. Please open a companion PR in https://github.com/ipfs/ipld-explorer-components/ (or add an explicit opt-out here until it lands).

I left inline comments with specifics. The two most user-visible: light mode lost the header background (`.webui-header` has no light rule), and the Settings section title renders a raw `theme` i18n key.

Smaller asks:

- The "Open questions" section of the PR description still contains template placeholders; please replace them with your notes.
- Some light-mode colors changed as a side effect (Box, modal action bar, progress bar, file import header, selected rows). If intentional, say so in the description; if not, the CSS variables can keep the original values.
- A unit test for the theme context (localStorage persistence and the `prefers-color-scheme` listener) would help keep this from regressing.
- Include screenshots of all screens in the PR description rather than in issue comments like https://github.com/ipfs/ipfs-webui/issues/1702#issuecomment-4952610540, so they stay next to the change they document.

src/components/theme-selector/ThemeSelector.js
kushmunjal: Added all keys to settings.json, options use t('themeAuto') etc

src/contexts/theme-context.js
kushmunjal: Exposed resolvedTheme in context setResolvedTheme() called inside applyTheme()

src/contexts/theme-context.js
kushmunjal: Added readStoredTheme()/writeStoredTheme() with try/catch + ALLOWED_THEMES allowlist

src/files/dropdown/Dropdown.js
kushmunjal: background='var(--element-bg)' passed to <Menu>, CSS rule kept as fallback

src/files/file/File.js
kushmunjal: Removed inline border, b--light-gray class now  handles it, dark override in index.css

src/files/sort-dropdown/SortDropdown.js
kushmunjal: background="var(--element-bg)" added to <DropdownMenu> at line 98

src/index.css
kushmunjal: Added palette comment; --gray-muted is wired to the disabled button override

src/index.css
kushmunjal: Brightened --charcoal to #c8c3bc, added .link.blue { color: #69c4cd }

src/index.css
kushmunjal: Added .webui-header { background-color: var(-- action-bar-bg, #F0F6FA) }

src/index.css
kushmunjal: Added [data-theme="dark"] .selected-item { color: var(--charcoal) !important }

kushmunjal (commented):
Hey @lidel! Thanks again for the detailed review — really helped me understand the codebase better.

  Here's a quick summary of the approach taken in this update:

  • CSS variables + [data-theme] overrides remain the core strategy — no component - level theme logic, everything driven from index.css so shared surfaces stay consistent
  • ThemeContext now exposes both theme (the stored preference) and resolvedTheme (the actual light/dark value) so chart components re- render immediately when the OS scheme changes in auto mode
  • All localStorage access is wrapped in try/catch with an allowlist ['auto', 'light', 'dark'] - gracefully falls back to auto in Tor Browser / private mode
  • Anti-FOUC script added in <head> of index.html to set data-theme before the bundle loads
  • Inline hardcoded borders in File.js and diagnostics-content.tsx replaced with CSS variable-aware alternatives so dark overrides can reach them
  • WCAG AA contrast addressed for charcoal text and .link.blue on the Status page
  • Explore page — added scoped overrides under .explore-page-container as a temporary measure until a companion PR lands in ipld-explorer-components

kushmunjal:
hi @lidel any updates of changes please update
thanks

lidel:
@kushmunjal sorry to leave you hanging. Shipyard is no longer reviewing new features; until Sep 30 we only handle security fixes, and after that our maintenance ends entirely: https://ipshipyard.com/blog/2026-the-end-of-ipfs-at-shipyard/

A future maintainer will have to pick this up. If you'd like to know who that is, that makes two of us :) The rest of the ecosystem is waiting for the same answer over at https://discuss.ipfs.tech/t/what-happens-to-ipfs-maintenance-and-public-infrastructure-after-september-30/20324

Lets keep this open so future maintainers have it ready for another review.

